PR #95, opened 2016-09-04
build roods

mikiones:
build roads:

simple algorithm to determinate best location for paths based on history of movements.
each location has count which is incremented when creep stamp on it (or mine)
after 100 tick the counts below mean are pruned and rest are reduced by mean value
best 2 are chosen to be new location for road.


ViezeVingertjes:
`t.length > 0 && t.n > 10` Seems incorrect, the number of steps is not on that level in the map.

mikiones:
yep i removed it just now. seam not working in not highly populated rooms. where 1 creep is mining

PR #100, opened 2016-09-05
Road planner fixes

mikiones:
Privateers can build roads, log fixes


mikiones:
there was bug in mean calculation making it fall to undefined.

PR #102, opened 2016-09-05
Add reserve action for claimers

torin555:
Allow claimers to reserve rooms without claiming them using a grey flag. Lower priority than the claiming action.


cyberblast:
Looking good. I will maybe only change the flag color, moving it into the "claim" color range..

torin555:
Alright sounds good, my choice of grey was rather arbitrary lol

PR #105, opened 2016-09-06
Prioritise building by adding priority value to every construction site object

mikiones:
build buildings with higher priority first.


mikiones:
will build extensions and containers before roads and another buildings. need adding rest of types and adjust priorities for them

cyberblast:
I merged a bit shorter solution for the sorting.. 
Sorting will be only once per loop (in room.js), not in building action when looking for a target (everytime again) => performance

Issue #17, opened 2016-07-18
Auto Construction


cyberblast:
**Roads**
- get path between spawn/source/controller & put sites
- save established connections to mem
- check periodically (1000 ticks? slowing down?)
- set all fields range 1 & 2 around sources & controller to roads
- set road:  x+y%4 == 0 || x-y%4==0 => lighter checkers

cyberblast:
What about auto setting extensions. 
- For each source.
- For each field with distance x (minimum range to source + next if belt fully build) around source. 
- When x + y % 2 == 0 (checkers) set Extension site. Else road. 
- After first scan set all inner fields for roads (checkers)

cyberblast:
__Spawn__
* Try to calculate mid point between Sources & Controller
* if it's a wall check concentric
* target should also have no adjacent wall (including diagonal)

cyberblast:
__Towers__
* Check concentric around spawn for a free field

Dasku:
Internal issue # 23

PR #117, opened 2016-09-07
build extensions around spawn.

mikiones:
auto build extensions around spawn


mikiones:
#17 kind of

mikiones:
upgraders should not work until max of extensions are builded.
unless there is downgrade

torin555:
Do you have this set up in a way it can be disabled?

mikiones:
not atm.

torin555:
Might be good to do that before the merge, this has the potential to break in bad ways.

cyberblast:
I'm also unsure about it. I think we should do some evaluation with it and maybe further development. 
I will set up a new branch for it.

mikiones:
rn is very simple build in predefined pattern if place is not taken. the RoomPosition extension is probably worth pulling into dev branch now. rest i still toying with

cyberblast:
Can you create a new pull request for the new branch please

mikiones:
sure give me sec

Issue #94, opened 2016-09-04
build container next to sources so miners can deposit to them


mikiones:
haulers, workers pick it up

mikiones:
not sure what should be a trigger for action, just missing container (or construction site for one) once we have miner 1 block away?

cyberblast:
miners get a dedicated source assigned only once and it will be stored to memory. Thats a good moment to check if there is a container at the assigned source present...

Dasku:
This currently works already doesn't it? I have a container by both sources and mineral. 

Can this be closed?

cyberblast:
no, its meant to be build automatically

PR #125, opened 2016-09-09
Find best position for mining container.

mikiones:
More goodes in RoomPosition extension:
1. find open positions in range(range <= ) and at range (range == )
2. find subset of positions next to given position.


mikiones:
part of #94

PR #132, opened 2016-09-09
Better path finder across multiple rooms:

mikiones:
* prefer own rooms,
* hight ways
* marked with exploit flag

Fixes issue with running into guarded rooms and stacking on rooms with protection.

### DIFF
--- a/creep.Action.js
+++ b/creep.Action.js
@@ -98,6 +98,29 @@ var Action = function(actionName){
         }
     };
     this.getPath = function(creep, target, ignoreCreeps) {
+
+        //defrent rooms? choose best routce main, marked or highways
+        if (creep.pos.roomName != target.roomName) {
+
+            var route = Game.map.findRoute(creep.room, target.roomName, {
+                routeCallback(roomName) {
+                    let parsed = /^[WE]([0-9]+)[NS]([0-9]+)$/.exec(roomName);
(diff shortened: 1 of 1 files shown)


cyberblast:
Great Idea! But I'm unsure what it will return if `new RoomPosition(25,25,route[0].room)` is solid rock... 
Maybe better to set the exit to the next room as target? 
May need some tests eventually...

mikiones:
the 2525 make sure it not stay on room boundaries. and once it is in target room it will change it to new target anyway.

mikiones:
potentially insteed setting point we can look for closes exit from path and search path to it.. problem is it will do search 2'ce 1'st to find pos to exit point second to find a path to exit point.

cyberblast:
> the 2525 make sure it not stay on room boundaries

Why should it stay there? 

> and once it is in target room it will change it to new target anyway.  

and immediatly move. There should not be a single loop without movement, also when setting current rooms exit as target (so far in theory... )

> potentially insteed setting point we can look for closes exit from path and search path to it.. problem is it will do search 2'ce 1'st to find pos to exit point second to find a path to exit point. 

I don't think this is required.

mikiones:
i'm having it in my live version and did not seen any issue with it. where without it i had allot loosed privateers trying route through protected room on way back to base.

cyberblast:
Ok, lets try it :)

PR #135, opened 2016-09-10
miners build own containers and repire them

mikiones:


cyberblast:
Yes that may be a workaround. But this can't be the solution. 
The miner is configured to harvest 1 source in exactly the time it takes until it restores. This is the optimal way. If the miner does other jobs as well, it can't do it's job efficiently. 
This is a worker & tower task...

mikiones:
the mining and repairing is in same tick in same tick so it still harvest all energy from sources

cyberblast:
I dont think so, as harvesting and repairing is within the same [action pipeline](http://support.screeps.com/hc/en-us/articles/203137792-Simultaneous-execution-of-creep-actions).

mikiones:
oh i see

PR #138, opened 2016-09-10
better privateer target choosing

mikiones:


cyberblast:
The path refactoring is fine. 
I'm unsure if the privateer calculation is required. Thats what the exploitMod was intended to do: ensure a balanced distribution of privateers among exploitation rooms.

mikiones:
well all privateers of to first flag, then to second... that is what i seen with dev code

mikiones:
same the old code is using line distance between rooms where even if room is adjusting there can be much longer path or no path at all

mikiones:
the Cloking must be / source in room as thy can be extinguish in deferent times.

mikiones:
i put newest version in deferent pr

PR #144, opened 2016-09-11
fix building walls in not owned room

### LINKED ISSUE #139: Privateers are retiring and building enemy's wall
annoying and funny was milking one of neighbours when he decides to build walls around himself.
guess what. the privateers decided to help him and build all walls and repairs them.

mikiones:
fixes #139


cyberblast:
Not sufficient to only check for walls... 
if ( Game.rooms[target.pos.roomName].controller &&  !Game.rooms[target.pos.roomName].controller.my)   return null;
may be better. But I will have a look on it. The problem may also occur in other actions as well..

PR #145, opened 2016-09-11
consider containers as target too

### LINKED ISSUE #141: robbing action is not taking in account containers ( neutral structure)

mikiones:
fixes #141


cyberblast:
Better to do only 1 search, with an improved filter...

mikiones:
those are 2 deferent search types. 
the containers are not enemy structures.

cyberblast:
its both "structures" (FIND_STRUCTURES) in a room not owned by us ;)

Issue #26, opened 2016-07-22
Add "dismantling" Action


cyberblast:
creep.dismantle  
by flag or for "workers" invading  
also for privateers to remove walls blocking way or to get energy from any unknown tructure

cyberblast:
Could be for privateers going into invasion to dismantle walls etc

mikiones:
any idea how choose target to dismantle ? in my working code i have that privateers have 
   Creep.action.robbing,
   Creep.action.picking,
   Creep.action.dismentle,
   Creep.action.harvesting,

PR #146, opened 2016-09-11
dismantle action

mikiones:
#26


cyberblast:
There should be a target validation. Or the creeps will eventually eat your base ;)

mikiones:
yes i do have it inside privateer if in target room do things..

cyberblast:
Yes, but in 4 weeks nobody will remember that. Could you please also implement an isValidTarget or isAddableTarget check? To see if its within our room. 
Dismantle is interesting for different things and will also get its own flag color for use within owned rooms...

cyberblast:
Implemented it requirig explicit flag instead of "in my room" check (to enable dismantling in owned rooms as well)

PR #147, opened 2016-09-11
Room ownership

mikiones:
some action validation depends on room ownership where this action happened.


cyberblast:
I believe  target.room.controller will also be null, when you don't have sight... 
Please also have an eye on typo.

cyberblast:
included in 50f5440b6c3bdc55f983b13aea7e15fbcf33b47e

Issue #124, opened 2016-09-08
Mix with dropmining


cyberblast:
Currently it's an all or nothing decision. Do dropmining or not. But there are situations when it's good and some when it's not. So this requires an enhancement. 

Dropmining may be used (auto, per miner), when: 
* mining container is full
* no mining container present (IN, at the source)

but only when
* at least 1 hauler present (else carry yourself)

problem to solve
* picking is only possible, when there is no enemy, because they drop too (at least first corpses) and that would lure haulers to go there and they may die

possible solution: 
* when invaders present, only pickup energy in near of source => in picking.newTarget

There is no need for an additional action (dropping). Dropping may be included in harvesting (conditional).

torin555:
+1 for this

Issue #109, opened 2016-09-06
Controller right next to Source issue


SirLovi:
The code has problems to distinguish which container is for the upgrader.
(There is another container under the harvester)

![capture](https://cloud.githubusercontent.com/assets/20193391/18273785/5c5f7d3a-743f-11e6-9545-5b56cd3b84da.PNG)

cyberblast:
yes, thats right. I have the same problem in one of my rooms. it will work if you place one of them out of the range of the source while in range of the controller (__current workaround__). 

Problem is: currently containers are IN or OUT. 
IN is for the miner and the hauler will get its energy
OUT is for the upgrader and the hauler will fill it. 

The container you need would be both, wich is not supported (currently). 
Will surely need a fix.

cyberblast:
Proposed solution outline: 

* The container gets filled by miner 
* The container gets used by the upgrader
* Main problem is the hauler
  * The hauler _could_ ignore the container. problem solved. 
  * But, as "upgrading" is the final desired action, as much energy as possible should go there. What about the other energy...
  * sending "as much energy as possible" into that container will also be a bad idea, because then the container is full and the miner can't mine :-1: 
  * The hauler should send some energy when it's quite empty (below x%) :+1:  
  * And take energy when its quite full
* To reach that we need to define a better classification of the containers
  * pure miner container (like current IN) - will be emptied by haulers
  * versatile OUT (like current OUT) - will be filled by haulers
  * managed containers
    * a container that should never be empty and never be full
    * filled AND emptied by haulers
  * an additional property may be if its the container for the upgrader (in addition to its class)
* container classification will be evaluated once and then stored into memory
* room will be extended with those properties
  * containers (all container)
  * upgraderContainer (0 or 1 container)
  * chargeablesIn  
    Containers (and Links), haulers shall take energy from, including managed (when full)
  * chargeablesOut  
    Containers (and Links), haulers shall put energy into, including managed (when empty)

cyberblast:
Does anybody want to implement this one?

mikiones:
i working on auto extensions and containers  atm.

cyberblast:
Gonna implement it in branch ContainerClassification 
https://github.com/cyberblast/screeps.behaviour-action-pattern/tree/ContainerClassification 
and merge into dev when complete...

PR #148, opened 2016-09-12
Container classification

cyberblast:
@ScreepsGamers/members 
This is a larger one, thats why I am using a feature branch. I think it could be merged now. My rooms are already running on it. 
Please have a look at it. Maybe somebody could also do some additional tests. 

Implements 
* #109 source & controller nearby 
* #124 conditional dropmining


torin555:
Yeah I think this should be merged

PR #156, opened 2016-09-13
Adjusted worker minEnergyAvailable

SirLovi:
Helps bigger RCL rooms recover after complete wipe when most of the extensions are empty.


cyberblast:
Yes but that is a corner case. 
In general the minEnergyAvailable helps to ensure that bigger creeps are created instead many smaller. 
But you are right, it's not perfect. There are even some more problems with it. 

I thought about it before and think the best solution will be to completely remove that function for relative comparison . Instead we should use minAbsEnergyAvailable as a function and calculate the min absolute Energy. That calculation could also take into account if all creeps have been killed, or such.

PR #162, opened 2016-09-14
Defcon - SPAWN_DEFENSE_ON_ATTACK

BringerOD:
This is up to date and ready to merge.

I changed this to a Boolean setting that when set,  attempts to store enough energy for a fight, then will spawn troops,  when attacked.

You do not need a flag.  If you have a flag for defense, this is additive.

Let me know what you think.


cyberblast:
Hey, sounds great! :+1: 
Unfortunately I got to go for now. :( 
I will have a look at it later.

BringerOD:
Special note here.

I stop producing creeps of various types to ensure there is enough energy, once there is, then those troops can get spawned. 

Also,  I added some logic not to spawn a worker unless there is some work to do, and changed the upgrader count to 2. We can remove this if you want, but its working nicely for me.

creep.setup.worker.js
```diff
@@ -7,6 +7,9 @@ setup.minEnergyAvailable = function(spawn){
     return 0.3;
 };
 setup.maxCount = function(spawn){
+    if (spawn.room.situation.invasion) return 0;  // Do not create in the middle of a fight.  We should be building fighters if we have energy left.
+    if (setup.ShouldWeConserveForDefense(spawn)) return 0;
+    if (spawn.room.constructionSites.length == 0)  return 0;
```
mikiones: check if we have tower and if nothing to retire first

creep.setup.upgrader.js
```diff
@@ -7,7 +7,13 @@ setup.minEnergyAvailable = function(spawn){
     return 0.5;
 };
 setup.maxCount = function(spawn){
-    return spawn.room.containerController.length > 0 ? 1 : 0;
+
+    let numberToBuild = 1;
+    if (spawn.room.situation.invasion) return 0;  // Do not create in the middle of a fight
+    if (setup.ShouldWeConserveForDefense(spawn)) return 0;
+    if (spawn.room.constructionSites.length == 0)  numberToBuild +=1;
```
mikiones: would be nice if we check how full the container is if > 90% spawn additional upgraders

BringerOD:
Good catch!  Agreed when storage is > x% spawn more upgraders.

BringerOD:
Is there a way to accept your change?

creep.setup.worker.js
BringerOD: Not sure you want to spawn an entire worker for just a few repairable.  Maybe just urgent repairables.

cyberblast:
Why do you want to have more upgraders? 
More creeps mean less cpu & less space. 
We should stick to the minimalistic approach and instead adjust creep size... (at least until > 50 body parts required)...

cyberblast:
Can we please remove the worker/upgrader adjustment for the moment? To have the pure defcon solution as a first step. I think we all agree on this part (?). And I would like to merge that. 
Of course we may discuss further steps, but I don't want to mix that...

BringerOD:
sure I will remove.

BringerOD:
Done.

cyberblast:
Creep size is not only limited to 50 parts, its also the max available energy all your extensions can hold to invest in a creep at once. 
So .. there indeed is the requirement to make more upgraders possible. 
But at first the upgrader should grow. If then the storage level ist still increasing, an additional upgrader will be allowed.

cyberblast:
I'm unsure regarding the worker. 
The worker also supports with feeding, fueling etc. 

On the other hand there is nothing ONLY the worker can do, except building. And there are idle times for the worker. 

I think we should add an issue for it to implement at a later time. I dont think its that expensive keeping the worker for the moment. What do you think?

Issue #101, opened 2016-09-05
Utilize Links


cyberblast:

Animatrixy:
a much needed feature +1 for this

PR #174, opened 2016-09-17
very simple link implementation closest to storage is out far on  is in,

mikiones:
links very simple. a working start


mikiones:
#101

mikiones:
added miner links, thy need to be in range of 2 from source to be consider as valid target.
thy are still valid store target.

mikiones:
no point to keep around

Issue #161, opened 2016-09-14
Miner (& upgrader) positioning 


cyberblast:
There is still a problem with the managed container ( #109 ). 
Both, the upgrader & the miner want to sit on it. This can be a problem for the miner, as he needs to be in range of the source and in range of the container. (The upgrader has a long range (3) on the contoller, making it easier for it). 

1. There should be a calculation of optimal the place for the miner (maybe also the upgrader). Maybe thats what miki had in mind ( #127 ).  
Currently, it is also a problem, when the container is not directly adjacent to the source (also with unmanaged containers). They will not get used!  
So this requires an update. 

2. Still the problem with managed containers. Both creeps (miner/upgrader) could calculate the same place to go. It also could still be on top of the container.  
So managed containers will maybe require a special target place calculation.

mikiones:
there is already code to calculate optimum container position closest to spawn and having most coverage for miners just cherry pick it

cyberblast:
yes, but its all mixed up with a lot of other changes, making it hard to pick (merge) things. 
We should try to focus on isolated topics - solve it, show it, test it, merge it.

PR #177, opened 2016-09-19
assign spot around container farrest from spawn first

mikiones:
fixes Upgraders part #161


cyberblast:
This may be the best position in some cases, but surely not in all cases. There may also be spawns at various locations. I don't think this makes any sense at all.

PR #179, opened 2016-09-21
Links

cyberblast:


cyberblast:
cooldown is sufficient for the miner-links. there should at least be 1 additional container: at the controller.

cyberblast:
Seems to be working real good. after removing the miner containers it's saving additional (repairing) energy.

PR #189, opened 2016-09-24
don't fill up extractors's container with energy

mikiones:


cyberblast:
nope, thats not the problem/fix.

PR #191, opened 2016-09-25
fixing case when in room extractor is not yet build

mikiones:


cyberblast:
But what about 
```
if( _.isUndefined(this.memory.minerals)) {
    this.saveMinerals();
}
```
this is exactly meant to ensure this.memory.minerals is !== undefined

mikiones:
weird  i got error,  cant .forEach on null.

cyberblast:
is this ongoing? with the latest set? can't reproduce it...

PR #193, opened 2016-09-25
Extra hauler for every 2k backlog in mining containers

mikiones:
fixes problem in my base when mining spots are to far from storage

### DIFF
--- a/creep.setup.hauler.js
+++ b/creep.setup.hauler.js
@@ -14,6 +14,10 @@ setup.maxCount = function(spawn){
         if( spawn.room.links.length > 2) count = 1;
         else count = 2;
     }
+    
+    let csum = spawn.room.containerIn.map(e => _.sum(e.store)).reduce((a, b) => a+b, 0);
+    count += Math.floor(csum / 2000);
+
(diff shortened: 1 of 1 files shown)


cyberblast:
miners will fallback to dropmining when container is full. 
containers are very expensive, I'd really advice to switch to links as fast as possible. 
And I really want to reduce creep numbers as much as possible. 
But I'll think about it a second time.

mikiones:
yes and  i have 2 containers full of energy and miners dropping on floor.

cyberblast:
how about bigger hauler? would that be sufficient?

mikiones:
dunno there is so much one hauler can do. i mean need to visit all sources (energy miners, minerals,links). its not case that he is to small. it that he travel too slow. but he cant move faster , ergo more haulers are needed.

cyberblast:
yes I understand, but if its larger, it will not need to come back that soon and has more time to visit all the other targets. In the end its only a matter of how much can be moved around, and thats the sum of all carry parts of all haulers. no matter how many they are.

PR #214, opened 2016-10-13
Dont allow worker/privateer upgrading with storage

cyberblast:
For quite a while, worker weren't allowed to upgrade the controller as soon as there is a storage (due to a bug). 
In the meanwhile the upgrader is much more variable in its size, and I like that. 
Now (as the bug is fixed) workers (and privateers!) will support with upgrading (as soon as the storage level reaches MAX_STORAGE_ENERGY). That will block them to do other jobs. 
I believe the upgrader is able to manage the energy level of the storage on its own (will become larger/smaller), and should do it alone. 

What do you think?

### DIFF
--- a/creep.action.upgrading.js
+++ b/creep.action.upgrading.js
@@ -2,9 +2,6 @@ var action = new Creep.Action('upgrading');
 action.targetRange = 3;
 action.isAddableAction = function(creep){ 
     return !creep.room.storage;
-    if( creep.room.storage ) 
-        return creep.room.storage.store.energy > MAX_STORAGE_ENERGY;
-    return true; 
 };
 action.isAddableTarget = function(target){ return true; }; 
 action.isValidAction = function(creep){
--- a/creep.action.upgrading.js
+++ b/creep.action.upgrading.js
@@ -1,6 +1,7 @@
 var action = new Creep.Action('upgrading');
 action.targetRange = 3;
 action.isAddableAction = function(creep){ 
+    return !creep.room.storage;
     if( creep.room.storage ) 
         return creep.room.storage.store.energy > MAX_STORAGE_ENERGY;
     return true; 
--- a/creep.setup.hauler.js
+++ b/creep.setup.hauler.js
@@ -8,14 +8,16 @@ setup.maxMulti = function(room) {
         max += 2; 
     let contSum = _.sum(room.containerIn.map(e => _.sum(e.store)));
     max += Math.floor(contSum / 1000);
+    max += Creep.setup.upgrader.maxMulti(room);
     return max;
 };
 setup.maxCount = function(room){
+    let count = 0;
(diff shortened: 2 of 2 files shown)


ValkyrX:
Is there anyway to have more upgraders ? I think especially in early game as it takes soooo long to get to rcl 3 ? I think in later game say rcl 6 or 7 that no workers should spawn unless there is something to build

SirLovi:
It depends... 
For example some of my rooms are staying at the MAX_STORAGE_ENERGY limit as intended, but some of my other rooms get so much energy supply that the storage is almost full of energy and haulers can't keep up.. And several huge upgraders just idle next to the empty container..

I like the idea that privateers and workers would contribute to upgrading or at least charging the upgrader container and not just stuffing the storage if it's already at it's limit...

In the few recent days I've been thinking how to improve the supply effectiveness..
I've tried changing the priorities.. Making more workers or/and haulers..
I'm still testing it all out though..

cyberblast:
This change will only have effect at RCL 4+ (having a storage). 
Before you have a storage, workers & pioneers will (now) support with upgrading. 
How do you want to power more or bigger upgraders?  
The problem in lower levels is, that there is so much to do, everything seems expensive. So I think the key in the early levels is austerity. Don't build any roads or containers @ rcl 1. Build only few roads @ rcl 2 (main routes). At RCL3 build the tower, walls, only few ramparts, and 3 containers (controller, sources). 
Always use ~ 4 pioneers until RCL 3 (they will commonly be spent from other rooms and be bigger than what you can spawn in that room). You should be able to get quite fast to that RCL. From then on it's slowing down a bit, as you have to build defenses and try get rid of pioneers (e.g. reduce to ~2 until RCL 4).

SirLovi:
For example in this room..
There are huge upgraders because the storage is filled up..
I have reconsidered this and now think that the worker wastes too much time walking to the controller. So it shouldn't be allowed to upgrade. It should instead feed the link to controller whenever possible..

I've also noticed that the "full storage problem" is only happening in rooms that have storage quite far from the controller's container and link connection does not seem to be enough..
But other rooms that have controllers really close to storage are nicely keeping the 200 000E limit.

capture

cyberblast:
Ok, I think the problem is a lacking energy distribution. your upgraders require a lot of energy. commonly the haulers should simply put it into the link. That also works better when having 4 links. 
But your haulers are filling the container , I guess :/ 

Proposal: 
As soon as there is an additional upgrader allowed, it will also get an additional hauler.

SirLovi:
That's a nice idea... The fact that there is so much energy must also mean that energy is net being distributed properly, so while increasing the size and number of upgraders also increase the number of haulers...
That's a pretty easy fix (y)

cyberblast:
Is everybody fine with 6345704 ?

PR #232, opened 2016-10-24
Implementing mineralMiner

SirLovi:
Minerals renew after they are completely mined out.
So it should be a priority to mine them out as fast as possible to increase mineral acquisition in the long term.


cyberblast:
Hmm. possible. But a lot of (duplicate) code. Maybe more effective to simply increase the right "common miner"? 
I've got to think about it ...

cyberblast:
Also required to kill all existing "common" miners at minerals once this code is deployed. As it will spawn an expensive mineralMiner wich will not go to work until the old one is gone...

PR #260, opened 2016-11-03
Create a small claimer at lvl 3 to allow for claiming empty rooms.

forkmantis:
Since it only takes 1 CLAIM component to claim an unclaimed controller, create a small, affordable claimer setup to allow for claiming emptry rooms at lvl 3.


cyberblast:
Yes, but you'll need at least 2 parts for reserving... just to keep in mind...

Issue #265, opened 2016-11-04
Storage thresholds tiered by RCL


cyberblast:
Follow-up of #248

PR #268, opened 2016-11-04
Convert MIN/MAX STORAGE ENERGY to be tiered #265

forkmantis:


forkmantis:
My thinking was that since that formula was only present in the creep setup for RCL 8, that the only way you would be using that setup was if the room itself were an 8.  In that case, you'd be checking if the energy in storage is > than MAX_STORAGE_ENERGY[8], which would be 300000.  Am I thinking right on that, or have I missed something?

cyberblast:
Yes, you are absolutely right. sorry. 
Things are changing too fast currently. My fault. 
Great work!

PR #299, opened 2016-12-15
Make Invaders More Agressive

TafThorne:
It seemed to me that invading creeps would only attack creeps that
attacked them (what I understand to be the hostile creeps).  When
performing an invasion I felt it would be better to have then attack
anything that is not a white-list player's.  This will include creeps
that cannot attack (such as miners, claimers and upgraders) which helps
keep contested rooms free.

### DIFF
--- a/creep.action.invading.js
+++ b/creep.action.invading.js
@@ -47,38 +47,44 @@ action.newTarget = function(creep){
 
     if( !flag.room.controller || !flag.room.controller.my ) {        
         //attack healer
-        var target = creep.pos.findClosestByRange(creep.room.hostiles, {
+        var target = creep.pos.findClosestByRange(creep.room.enemies, {
             function(hostile){ return _.some(hostile.body, {'type': HEAL}); } 
         });
-        if( target ) 
+        if( target ) {
             return target;
+        }
         //attack attacker
-        target = creep.pos.findClosestByRange(creep.room.hostiles, {
+        target = creep.pos.findClosestByRange(creep.room.enemies, {
             function(hostile){ return _.some(hostile.body, function(part){return part.type == ATTACK || part.type == RANGED_ATTACK}); } 
         });
-        if( target ) 
+        if( target ) {
             return target;
-  
+        }
+        
         // attack tower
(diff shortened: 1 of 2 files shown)


TafThorne:
I might be mistaken but when I was on holiday for the past two weeks I noticed that my invaders would not attack claiming and mining units coming into a room I was exploiting.  They would attack other creeps once provoked.  By that tiem I had lost the room.  

I looked over the code an wondered if it was being a bit too gentle when invading.  This pull request should make any non-player owned or non-whitelist creep a target during an invasion.

pandabear41:
Targeting should work perfectly fine without any changes.
What color flag are you using to kill players in an unowned/reserved room?

Your code will break creep.action.invading.js @ line 135, 139, 143
since rangers don't check for your own creeps, they will attack any of your other creeps.

TafThorne:
> Targeting should work perfectly fine without any changes.

It did not seem to, hence my attempts to check what FIND_HOSTILE_CREEPS did.  Based on what I could see going on I assumed it ignored creeps with no attack or healing parts.  That is why I tried substituting it for FIND_CREEPS.  

> Your code will break creep.action.invading.js @ line 135, 139, 143
since rangers don't check for your own creeps, they will attack any of your other creeps. 

I am in my own white list.  This will be like any alliance member (although I am not an alliance member at present).  It did not seem necessary to check for self separately from the whitelist.  That could be done, or a note added to the whitelist warning that invasions could be self destructive if your own name is not there.

pandabear41:
FIND_HOSTILE_CREEPS is any creep alive where creep.my is false (https://github.com/screeps/engine/blob/0ddf1684a519dc83345b0dd341b8c64ddb3f0145/src/game/game.js). Attack or heal parts don't matter. 
The attack routines in this repo prioritize Heal and Attack parts, but don't ignore other creeps.

TafThorne:
Everyone seems sure that the behaviour I thought I saw is not how things actually work.  I am willing to accept that. We can close this pull request.  Next time I try to defend a harvesting room or take one over I will keep a close eye on what guards or invaders do.

PR #309, opened 2016-12-28
add controller signing

ESchuderer:


cyberblast:
I think he's aiming on a more automated approach, Where every room gets signed automatically. Which is fine.

creep.js
```diff
@@ -334,6 +334,12 @@ var mod = {
             }
         };
 
+        Creep.prototype.controllerSign = function() {
+            if(CONTROLLER_SIGN && (this.target.sign.username != this.owner.username || this.target.sign.text != CONTROLLER_SIGN_MESSAGE)) {
```
cyberblast: Maybe preferable to use a `&&` here? 
If YOU have set an individual message, you will not want to overwrite that...

parameter.js
```diff
@@ -76,6 +76,8 @@ var mod = {
     NOTIFICATE_INVADER: false, // Also log common 'Invader' hostiles
     COMBAT_CREEPS_RESPECT_RAMPARTS: false, // causes own creeps not to leave through ramparts
     COST_MATRIX_VALIDITY: 1000,
+    CONTROLLER_SIGN: true,
```
cyberblast: CONTROLLER_SIGN should maybe be `false` as default? unsure about that.

webremote:
No, I was more aiming on setting the `controllerSign` functionality _per action_ (like there's only two creeps right now capable of actually signing the controller) instead of checking who's in the area

cyberblast:
yes, but (via this PR solution) every room will get its sign automatically once you reserved it. Without having to iterate over all creeps etc. 

I see the point that this is no solution for setting an individual message to any room. But to do that you will also need to actually *send* a creep to the controller...

ESchuderer:
It was intended by me to only have creeps sign controllers if they are there anyways, so neither much CPU noir creeps which could do something more useful are wasted.

creep.js
ESchuderer: I've used this one so you're able to update messages, if we do not want to override existing messages then we can ommit this.target.sign.text of course. Not sure about that.

PR #363, opened 2016-12-30
update statistics.js to fix buggy email

webremote:
fixes incorrect html for controller with reservation

### DIFF
--- a/statistics.js
+++ b/statistics.js
@@ -69,9 +69,9 @@ var mod = {
                     }
                     else if( !room.controller.my && room.controller.reservation ){
                         // controller
-                        message = '<ul><li><b>Room ' + room.name + '</b><br/><u>Controller</u><ul><li>Reservation: ' +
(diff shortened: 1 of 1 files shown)


cyberblast:
hmm, why? this is the exactly same pattern as for the owned controllers. And it should look the same, no?

webremote:
image

the ones i removed were in fact doubles causing a similar #310

cyberblast:
hmm strange, looks completely different for me. 
Also the resulting HTML should be valid already, no? 

image

cyberblast:
> the ones i removed were in fact doubles causing a similar #310 

No, there you only wrapped the header notification into a div.

webremote:
yes, in this one i fixed the double UL
image

not sure why yours is showing differently, perhaps overriden or your viewer fixes the html or something?

webremote:
hm, now it sends it just right, without the hotfix applied. it's weird.. :/

cyberblast:
hmm, ok. 
I'm going to close the PR, ok? We still have the branch, if we need it again...

PR #378, opened 2017-01-02
Correct references to 'creep' in fleeMove on creep.js

whardeman:
Holdover references from creep.behaviour.privateer.js refactoring, ref commit aa7c14943eec112f2e383390690702f2049ab56b


whardeman:
As is, when a creep attempts to flee, the following exception is thrown:

`ReferenceError: creep is not defined`

PR #379, opened 2017-01-02
Allow containers to be used as drop offs

whardeman:
- containers built near a room edge (distance of 2 is arbitrary, could be 3-4) are used by privateers for drop off points, but the full containers will just sit there since the haulers/workers never target them
- adds a new property "edge" to containers, if true, allows them to be targeted for 'uncharging' action
- primarily useful in rooms with RCL < 5 or with too many exits to be covered by available links


cyberblast:
hmm, we could also let all containers be `in` (hauler will empty), except when controller = true. Would make it much easier. 
```
let byType = c => c.controller == false;
this._in = _.filter(this.all, byType);
``` 
but privateers should maybe only fill those where `c.source === false && c.mineral === false`

whardeman:
Sure that makes sense. I haven't explored deeply. How do privateers determine whether they drop in the container? Is it currently just if the container is 'out' ?

whardeman:
After digging into the code, it looks like this is how privateers select a target for _charging_:
```
    let deposit = [];
    if( creep.carry.energy == carrySum ) deposit = creep.room.structures.links.privateers;
    if( creep.room.storage ) deposit.push(creep.room.storage);
```
So, I think you're suggesting something like the below pseudocode:
`deposit.push( "containers that are not also mineral/source containers" )`

whardeman:
This has been tested and works great. One problem: with `out` having all non-source/non-mineral containers, haulers are also trying to fill privateer edge containers. ☹️ 

So, I'm continuing to set the `edge` property in saveContainers and filtering them from `out`.

TafThorne:
Do you care about it being an "edge container" when the remote energy comes in?  Should the things bringing in remote energy not just get to a home room, pick the closest drop off point to them that has space and dump their load?  I don't mind if the energy goes into a container, storage, link or even spawns and extensions.  What we want is delivery to the room completed and the remote carrier back out collecting.  The in-room haulers can then balance the energy and move it about if needed.

TafThorne:
> haulers were also trying to fill privateer edge containers

Will this not sort itself out once the "edge" ones end up at +70% full?  Containers hold 2000 energy and most transports will have less than that will they not?  At that point the hauler behaviour will move energy from edge ones to central ones as with anything else.  

Would it be better to work out "source" containers and "storage" containers instead of "edge" ones?  For the most part, once we have some links constructed you do not want to use containers for storage at all or do you...

whardeman:
All good questions. I'll do my best to explain my thinking.

I don't care about it being an edge container at all. In fact, I don't care where a privateer drops the energy, as long the following conditions are met:
- The creep uses the closest structure
- The structure isn't a source or mineral container (don't want those filling up completely)

To your point, the whole reason for doing this is efficiency, both the efficiency of keeping Privateer work parts out in the field (also solved if we change to remote miners/haulers) and the efficiency of the in-room hauler not moving to the container (which is typically far from the room center) until necessary. 

So, we need to make sure that containers placed in this way are filled _only by privateers_ and emptied by any appropriate in-room creep (hauler, worker, etc). Otherwise, efficiency degrades significantly because of 1) haulers taking trips out to the container _to fill it_ and 2) privateers more frequently encountering full containers where they could have otherwise dropped their loads.

whardeman:
Here's a screengrab of the thing in action. A hauler has just picked up from the container because it was over the threshold required for it to be a valid target, and a privateer has just entered the room to drop 300 energy.

image

If the privateer had entered through the North exits, where there is no nearby container, the creep would walk the energy to the Storage to be dropped off.

@TafThorne Hopefully I'm making sense here. I'm not sure if you're objecting or just curious about what this is supposed to accomplish, so I'm trying to give plenty of detail. As I noted in the first comment, this is primarily useful in rooms below RCL5 or rooms with lots of exits and exploitable rooms nearby. Obviously links are far more efficient, but you don't always have the luxury of a link.

TafThorne:
> I'm not sure if you're objecting or just curious about what this is supposed to accomplish

I find myself unsure about that sometimes :-P 

It is mostly the curiousity but I was questioning whether their was a need for the "built near a room edge (distance of 2 is arbitrary, could be 3-4)" part of identifying things.  When a privateer gets into the room it just needs to find the closest structure to dump energy into and then leave.  In room haulers can move much more stuff about so even if it dumps stuff into a mining container or link that happened to be the closest thing it should still get emptied soon.  

If you are not prescriptive about where in the room the container/link/storage is it can be anywhere and be the closest.  Once the energy is stored in the room it is up to the hauler and link code to do sensible things with it.

whardeman:
Ok, I understand. As for being built near a room edge, the privateer doesn't care. It is looking for the closest structure to dump into and that's pretty much it. As for ignoring mineral/source containers, that was already in the code before I got to it...I'll let @cyberblast state the reasons for that rather than put words in anyone's mouth.

The only reason to identify "edge" containers apart from anything else is to manage hauler behavior. We _really_ don't want them putting energy into those containers. They are expressly there for remote mining efficiency and haulers should only withdraw from them. Having the 'edge' property is the simplest way I could find to exclude them from `room.structures.container.out`. If someone has a better way, I'm open to it.

I should also say: No one is required to use containers like this. If you don't build a container near an exit, your privateers won't ever bother with it. The code I've added in the pull request is _required_ to make privateers even notice them.

cyberblast:
I'd propose to remove all containers from out, except the controller container and the "managed" if empty. 
This way we don't need an "edge" rule. 
And I'd advice not to use miner containers, as miner will start dropmining when their container is full. 
Filling extensions by privateers may be possible, but will take more time for the privateers. Unsure about that. I'd personally tend to keeping it simple: privateers put it in the storage (or a container), haulers manager its distribution from there.

whardeman:
Agreed on all points.

PR #389, opened 2017-01-03
missing return in task.defense.nextAction

cyberblast:


Dasku:
This fixes the travel issue!

PR #391, opened 2017-01-03
reduce workers and allow upgrading with full storage

cyberblast:
* reduce worker numbers to 0 in some cases
* allow worker upgrading when storage filled


creep.setup.worker.js
```diff
@@ -22,7 +36,9 @@ setup.RCL = {
         fixedBody: [],
         multiBody: [CARRY, WORK, MOVE],
         minAbsEnergyAvailable: 200,
-        minEnergyAvailable: 0.3,
+        minEnergyAvailable: room => ( !room.population || 
+            !room.population.typeCount['hauler'] || room.population.typeCount['hauler'] == 0 ||
+            !room.population.typeCount['miner'] || !room.population.typeCount['miner']== 0 ) ? 0 : 0.3,
```
TafThorne: Could this copy-paste code be abstracted out to library or function?  `vettedMinEnergy( 0.3 )` would look much tidier and make maintenance of what else needed to be `OR`ed  together much simpler.  

I occurs 5 times in this set of changes.

cyberblast: feel free

TafThorne: I do not have a very friendly development environment setup at present.  It would be hard to do and check for me and I should not blind type code.

TafThorne: Are `room.population.typeCount['hauler'] == 0` and `!room.population.typeCount['miner']== 0` correct?  You do want to check that there are 0 haulers and not 0 miners.  I am only asking becasue I would normally write ` != 0` for the not equal test and it is an easy typo to miss a ! from the start of  block.  

I think the code is correct.  Just worth highlighting while we are already looking at it.

cyberblast: no, there has been an error indeed in that expression. Hope this is better now...

TafThorne: I am glad my eyeballing leads to something productive as instead of just making me sound like I like bell acking a lot. 

New code looks correct based on the names of the functions given.

PR #394, opened 2017-01-04
don't try to reserve a room with no controller (another source keeper room wrinkle)

karlthepagan:
Flag `invade.exploit` triggers behavior for both privateer and claimer. The claimer case needs guards in the code to exclude rooms without a controller.


karlthepagan:
oops, this isn't quite right

PR #404, opened 2017-01-05
Increase Upgraders Used In RCL3 Rooms To 3

TafThorne:
At Room Level 3 there cannot be a storage unit in the room (assuming it has not been downgraded).  Currently only 1 upgrader is utilised which makes progress form Room Level 3 to Room Level 4 slow.  As part of the efforts to improve the low level game behavior we can increase the target number of upgraders to help get the room up to RCL4, with storage, as fast as possible.  

Currently the code defaults to 1 upgrader.  Room Controller Level 3 allows 10 extensions of 50 capacity and 1 spawn of 300 capacity, making the maximum spend on a single creep 800.  The current upgraders fixed parts cost 300 and each multi set costs 350, meaning the largest upgrader an RCL3 room can make will have 1 multi giving 5 work parts. That is 5 energy into an upgrader per tick.  

If we try to fully utilise one source of 3000 energy over 300 ticks, we need to transfer 10 energy per tick into the upgrader.  That will take 2 upgraders.  Most people have two sources in the room so we could supply 4 upgraders if all the energy went into them.  Let us make sure to leave something spare for building and so forth by making the target 3 upgraders.  

This should make it take only 33% of the time to reach room control level 4 in an ideal world.


TafThorne:
> but your workers will support with upgrading until there is a storage.

Good question.  My idle watching suggests not but let me do the same check to see what the workers add.  

Looking at the current worker setup for RCL3
```
    3: {
        fixedBody: [],
        multiBody: [CARRY, WORK, MOVE],
        minAbsEnergyAvailable: 200,
        minEnergyAvailable: 0.3,
        maxMulti: 8,
        maxCount: room => ( room.situation.invasion ) ? 1 : 4,
        maxWeight: 9600
```

So we get up to 4 workers with up to 8 multi part sets.  Each multi part costs 200 so the reality is we can only build a creep with at most 4 sets of parts.  That is 4 work parts or 4 energy per tick per worker upgrading.  So workers are almost as efficient at the transfer as upgraders.  

So if workers can take every from a container by the spawn and help upgrade we would need 1 upgrader and 4 workers to move 21 energy per tick.  

**TLDR** : If your workers have nothing better to do, 4 workers and 1 upgrader would be enough if workers used Containers as energy sources.

cyberblast:
commonly I'd suggest to also add some pioneers or increase the worker count.

TafThorne:
I do not think that workers use containers to refill their energy.  This can mean they spend a lot of their time moving about, rather than upgrading.  An Upgrader can sit in place and let the haulers bring it energy.   

Am I mistaken in thinking that workers do not take energy from containers, as upgraders do?  

In my current level level room this made worse by the awkward path between the energy sources and then controller.  
image

You can see the workers in the above image, they have smaller yellow parts and are currently no where near the controller.  Even if they do take energy from containers, any fueling, feeding, or repairing jobs that arrive will draw them off.  

There are also two pioneers in there and some privateers too it seems.  

I have been running this code on my game since about 13:00 (UK time).  At 16:00 I should be able to compare the previous 3 hour period to the current one in the Room Overview and see what the control points behaviour looks like.

cyberblast:
> Am I mistaken in thinking that workers do not take energy from containers, as upgraders do? 

They are only taking from source containers

> any fueling, feeding, or repairing jobs that arrive will draw them off.

That's the PRO argument for workers. Because in those cases you don't want the energy to go into the controller but in your jobs.

TafThorne:
> That's the PRO argument for workers. Because in those cases you don't want the energy to go into the controller but in your jobs.

I agree we need workers and that their jobs need doing.  Combined with pioneers the jobs do get done reasonably quickly.  

My point is that without enough upgraders to move energy into sources it gets left on the ground.  More than once I have found the source containers full and energy being drop mined.  For the cost of a few more upgraders we can guarantee that haulers stay busy and at least 75% of the energy that could come from 2 sources could make it into a controller.  

Leaving it to workers means waiting for them to move about.  If your source and controller are very close this works quite well (at least until one source is depleted).  If your source and controller are not so close together the travel time can be a big cause of delay in making it to RCL4.   

Once storage can be built, it should be.  After that point all spare energy in the room is gathered up.

cyberblast:
hmm we could set allowed upgraders to 2 if no storage present and add an additional upgrader when there is energy on the ground next to the source.

TafThorne:
> hmm we could set allowed upgraders to 2 if no storage present and add an additional upgrader when there is energy on the ground next to the source (maybe only if > x).

I was going for a computationally cheap fix and programmatically simple for the small amount of time a room should be at level 3.  We already have a simple check for level 8 and once the room reaches level 4 it is into the territory of existing code.    Currently my 4 rooms are at 50/60 CPU with nothing out of the ordinary going on.  

> I have been running this code on my game since about 13:00 (UK time). At 16:00 I should be able to compare the previous 3 hour period to the current one in the Room Overview and see what the control points behaviour looks like.

As promised it is now 16:06 and I can give the figures for the last few 3 hours slots.  I had been running as a level 3 room with finished extensions since first thing this morning (about - 9 hours).

Time      CP Energy 
-3 Hrs    38.4k
-6 Hrs    14.6k
-9 Hrs    12.9k

It is a bit of a crude measure but it suggests 3 upgraders worked OK.  

It takes 135k to get from RCL3 to RCL4 so at the rate of around 38k for 3 hours it would take about 12 hours to reach RCL4.  Not long to worry about it not being the ideal delivery rate of energy.  


I do not mind whether the more complex check for dropped energy goes in or the simpler RCL3 = 3;

TafThorne:
p.s. I like seeing this hive of industry:
image

TafThorne:
> Just had that case when I reached rcl 4. my upgraders run dry.

So my original code was perfect for your example.  At level 3 extra upgraders.  At level 4 revert to previous behaviour and hope a Storage unit is up and running soon.  There will be at worst the lifetime of a creep of overlap while the two extra upgraders time out.

cyberblast:
but the upgraders will run dry on RCL 3 the same. 
Also, maybe not everybody will immediately build the storage at RCL 4.

TafThorne:
> but the upgraders will run dry on RCL 3 the same.

I never witnessed that but I did only run a test on a couple of rooms.  If you are confident that the code change you proposed will get most rooms upgraded from 3 to 4 much more swiftly then that is fine.  It just seems a much more complicated code change.  You also require drop mining to have started (and therefore energy to have begun expiring) before adding new upgraders.  

> Also, maybe not everybody will immediately build the storage at RCL 4.

That is true. Continuing to dump spare energy into the controllers is always a good idea.  My proposed change did not cover that case.

cyberblast:
> You also require drop mining to have started 

Yes. This way it's like a self adjusting mechanism. If your energy consumption is sufficient, why spawn the 3. upgrader.

TafThorne:
I was doing the minimum change for what I saw a need for.  One test that changed the code behaviour in a narrow window.  If you are happy to add a more complicated set of metrics then what you have proposed will do a better job for more situations.   

My level 4 room is dumping about 32k every 3 hours into the controller.  My level 6 rooms are down at 26k.  Hopefully when the remote mining improvements come in the added efficiency will bump that up.  I have not build anything in those rooms for a day or two so the output is all maintenance and creeps.

cyberblast:
hmm, I guess thats because your storage is not full? 
Upgrader size and count is throttled by storage energy level. You can decrease those thresholds to enable more upgrading...

TafThorne:
> hmm, I guess thats because your storage is not full?
They are all over the maximum value.  I think it is just the rate at which the rooms have settled.  All those pioneers cost energy to produce.  I can also see about 6k per 3 hours going on building maintenance.

cyberblast:
over max is not enough ;) 
you get a bigger upgrader for every 20k surplus above max. 
and an additional one when reaching max*2

or something like that

TafThorne:
If Max is not Max then what is it :-P 

I guess it will make a larger one, drain the store, make a smaller one, repeat.  More energy into the room would beef that up.  More efficient remote mining is the only way to do that on a room that already has two dedicated miners.

cyberblast:
its the amount of energy you want to have in your storage. when you get above it we gonna burn it using upgraders.

Issue #399, opened 2017-01-04
Defenders only for own rooms


cyberblast:
spawn defenders (Task.defend) only for our own owned, reserved and exploiting rooms. Not "any" room, where a foreign creep has been spotted. 

Should not change anything regarding yellow/yellow flags (guards).

karlthepagan:
IMO exploiting and reserved rooms should possibly have a parameter for the user preferences.

cyberblast:
hmm. or a defense black list. of rooms to ignore...

PR #408, opened 2017-01-05
don't auto-defend rooms without a visible controller

karlthepagan:
this may introduce a few side-effects in existing combat scenarios


cyberblast:
hmm did you see #399? 
but may be a valid temporary solution...

karlthepagan:
IMO exploiting and reserved rooms should possibly have a parameter for the user preferences.

PR #410, opened 2017-01-06
Feature loyal privateer

karlthepagan:
these are the framework changes needed to customize how privateers operate when they are full of stolen goods

### DIFF
--- a/creep.behaviour.privateer.js
+++ b/creep.behaviour.privateer.js
@@ -16,7 +16,7 @@ module.exports = {
             logError('Creep without action/activity!\nCreep: ' + creep.name + '\ndata: ' + JSON.stringify(creep.data));
         }
     },
-    nextWorkAction: function(creep, carrySum){
+    nextWorkAction: function(creep){
         Creep.behaviour.worker.nextAction(creep);
     },
     nextAction: function(creep){
@@ -40,7 +40,7 @@ module.exports = {
                 }
                 //if( Creep.action.storing.assign(creep) ) return;
                 if( Creep.action.charging.assign(creep) ) return;
-                this.nextWorkAction(creep, carrySum);
+                this.nextWorkAction(creep);
                 return;
             }
             // empty
@@ -50,7 +50,7 @@ module.exports = {
             else {
                 // no new flag
                 // behave as worker
-                this.nextWorkAction(creep, 0);
+                this.nextWorkAction(creep);
                 return;
             }
         }
--- a/creep.behaviour.privateer.js
+++ b/creep.behaviour.privateer.js
@@ -16,6 +16,9 @@ module.exports = {
             logError('Creep without action/activity!\nCreep: ' + creep.name + '\ndata: ' + JSON.stringify(creep.data));
         }
     },
+    nextWorkAction: function(creep, carrySum){
+        Creep.behaviour.worker.nextAction(creep);
+    },
     nextAction: function(creep){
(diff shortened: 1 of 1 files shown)


cyberblast:
hmm, the room additions already came in with #408 . 
Why do you want to encapsulate the worker call?

karlthepagan:
whoops, looks like I didn't rebase #408 properly... I think i've straightened my current commits out

here's the internal feature for the worker call: https://github.com/cyberblast/screeps.ocs.internal/pull/156

cyberblast:
I don't think this makes any sense at all. 
nextWorkAction is called ONLY when at home room... 

What do you want to achieve?

karlthepagan:
@cyberblast I've pushed a clean reabse of this feature.

nextWorkAction is also called when the energy is full outside the home room.

The intent of my customization is to prevent privateers from repairing roads / containers in border territories and instead come home with their energy stores. See https://github.com/cyberblast/screeps.ocs.internal/pull/156

cyberblast:
I'd propose to completely remove "repairing" instead. we've got "repair on driveby" implemented recently. You may try if that is sufficient...

karlthepagan:
No problem. I only really issued this PR in order to justify #408. As long as the ability to customize how the privateer goes to work exists then everyone can customize it's behavior as they need.

karlthepagan:
https://github.com/cyberblast/screeps.ocs.internal/pull/156 isn't really consequential. I just want a mechanism to extend the work behavior specifically. It's easier than rewriting the whole privateer class.

creep.behaviour.privateer.js
```diff
@@ -16,6 +16,9 @@ module.exports = {
             logError('Creep without action/activity!\nCreep: ' + creep.name + '\ndata: ' + JSON.stringify(creep.data));
         }
     },
+    nextWorkAction: function(creep, carrySum){
```
cyberblast: carrySum is included in creep (as cached property creep.sum)

karlthepagan: Thanks, didn't see the caching.

cyberblast:
but 

> nextWorkAction is also called when the energy is full outside the home room. 

is not correct...

karlthepagan:
Oops, sorry I misunderstood the home room concept. Of course that's the case, but privateers operate outside of allied rooms. (robbing energy from structures, exploiting unguarded sources).

Both calling cases here are significant to me: https://github.com/karlthepagan/screeps.behaviour-action-pattern/blob/8a41d1ed52b9cc1ccd733774c164b869bbfbe4d7/creep.behaviour.privateer.js#L43

I was using privateers to great advantage in this room https://screeps.com/a/#!/room/W73S55

Unfortunately after they grabbed the energy they proceeded to repair my neighbor's roads.

cyberblast:
No. homeRoom is always the room of its birth. The room where it will bring the energy to. its home. 
The reason why it is repairing those structures is 
https://github.com/karlthepagan/screeps.behaviour-action-pattern/blob/02f956650a706c7b1b57dedd567a4088195bb633/creep.behaviour.privateer.js#L104 

That's what I was referring to when proposed to completely remove that repairing action ;)

cyberblast:
I'll create a PR...

karlthepagan:
Sorry I was probably misled by some debugging statements I put in there. This makes more sense.

cyberblast:
no worries. I'm glad to help getting things sorted

PR #411, opened 2017-01-06
prototype inheritance for mixins

karlthepagan:
I need this for when I'm mixing-in an extend:

```
const mod = {
    extend: function(){
        this.prototype.extend(); // basically super.extend();
        // extension code
    }
}
```


karlthepagan:
needs testing

PR #413, opened 2017-01-06
provide class-like inheritance to static mixins

karlthepagan:
example `viral.room.js`

```
module.exports = {
    extend: function(){
        Object.getPrototypeOf(this).extend.unbound.apply(this);

        Object.defineProperties(Room.prototype, {
            // additional Room.prototype properties
        }
    }
};
```

The prototype chain continues all the way until the value provided by load (in this case Room). That way all static members of Room are also available via `this` inside the module (so that API objects and OCS objects behave similarly).

### DIFF
--- a/global.js
+++ b/global.js
@@ -1,9 +1,9 @@
 var mod = {
     init: function(params){
         // Load extension functions
-        Creep.extend = load("creep").extend;
-        Room.extend = load("room").extend;
-        Spawn.extend = load("spawn").extend;
+        Creep.extend = load("creep", Creep).extend;
+        Room.extend = load("room", Room).extend;
+        Spawn.extend = load("spawn", Spawn).extend;
         // make params available globally
         _.assign(global, params);
         // Add more stuff to global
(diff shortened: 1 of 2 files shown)


karlthepagan:
updated example,

`Object.getPrototypeOf(this).extend();` does not offer full method polymorphism in superclasses.

karlthepagan:
Still working out how to get full method polymorphism.

karlthepagan:
New commit provides a property on the function exposing its unbound form.

cyberblast:
but _.create doesn't seem work for overriding simple values. Like in parameter.js.

cyberblast:
hmm we could add the original method as well. So that we can call it from the override.

karlthepagan:
I'll have a closer look at viral.parameter use-cases

karlthepagan:
I gave it a shot, but I want `this` inside Room.extend(), it was the first reason I started messing around with #415.

I would like to understand how _.create and _.assign produce different outcomes. It is not clear from looking at the overrides provided by screeps.ocs.internal.package

karlthepagan:
Superseded by #415

PR #415, opened 2017-01-06
call bind for viral function overrides

cyberblast:
allows to use `this` inside viral function overrides


cyberblast:
Calling base function: 
example using global.js override of `custom()`
in `global.js`
```
    custom: function(){
        console.log('base');
    },
```
in viral.global.js: 
```
    custom: function(){
        if( this.baseOf.viral.custom ) this.baseOf.viral.custom();
        console.log('custom');
    }
module.exports = mod;
```
in internalViral.global.js: 
```
    custom: function(){
        if( this.baseOf.internalViral.custom ) this.baseOf.internalViral.custom();
        console.log('internalViral');
    }
```
result: 
image

karlthepagan:
OK I think I get the solution. Will integrate into #413 application code.

cyberblast:
> Please store a reference to the unbound function for me ;) 

uhm. did you read the sample? It's there and its called. even through 2 overrides

karlthepagan:
> uhm. did you read the sample?

I see it, I just couldn't delete comments on mobile.

karlthepagan:
In order to get the same inheritance polymorphism as #413 I still need to customize `this` when I call into the parent function. Because `this` does not have prototype inheritance that means that methods declared in the viral will need to be explicitly overloaded and explicitly call the parent function for my case to be completely consistent.

`this.baseOf.internalViral.extend.apply(this);`

However, I can live with that for now.

karlthepagan:
If you're ok with giving the warning "this feature does not support virtual/prototype functions" then please do merge.

We can iterate later to find a good compromise between _.create and _.assign.

karlthepagan:
Err.. just tested. The lack of prototypes doesn't hurt the way I thought it would. It's complicated ;)

karlthepagan:
`this` is not available inside `Room.extend();`

```
module.exports = {
    customField: 0,
    custom: function() {
        console.log(this.customField);
    },
    extend: function(){
        debugger;
        const mod = this;

        this.customField++; // after this line Room.customField is NaN
        this.custom(); // [10:04:12 AM]TypeError: this.custom is not a function
    }
}
```

cyberblast:
hmm, no I don't get that error. Instead I get: 
image

viral.global.js: 
```
var mod = {
    customField: 0,
    custom: function() {
        console.log(this.customField);
    },
    extend: function(){        
        if( this.baseOf && this.baseOf.viral.extend ) this.baseOf.viral.extend();
        debugger;
        const mod = this;

        this.customField++; // after this line Room.customField is NaN
        this.custom(); // [10:04:12 AM]TypeError: this.custom is not a function
    }
}
module.exports = mod;
```

karlthepagan:
Ah, look me a minute to figure out why. I was referring to adding customField on the base class and did not make it clear in my example.

Here's a more complete example:

## IN global.js
```
var mod = {
    customField: 0,
    custom: function() {
        console.log(this.customField);
    },
    extend: function() {
        debugger;

        this.customField++; // NaN
        this.custom();
/* [4:30:59 PM]TypeError: this.custom is not a function
    at Object.extend:11:14
    at Object.extend:8:89
    at Object.extend:8:73
    at Object.module.exports.loop:124:28
    at __mainLoop:1:15387
    at eval:2:4
    at Object.c.runCode:6:14086
*/
    },
    init: function(params){
        debugger;
        this.customField++; // 0 -> 1
        this.custom(); // OK!
        // rest of class
    }
}
module.exports = mod;
```

## viral.global.js
```
module.exports = {
    extend: function () {
        debugger;

        this.customField++; // 1 -> 2
        this.custom(); // OK!

        if( this.baseOf && this.baseOf.viral.extend ) this.baseOf.viral.extend();
    }
};
```

Also you're probably testing with internalViral.global.js which is a problem for testing global because it's extend doesn't call the super method.

## IN internalViral.global.js
```
var mod = {
    extend: function(){
        debugger;

        this.customField++; // 2 -> 3
        this.custom(); // OK!

        if( this.baseOf && this.baseOf.internalViral.extend ) this.baseOf.internalViral.extend();

        // rest of class
    }
}
module.exports = mod;
```

It probably seems pedantic, but this is a problem because I am trying to refer to `this` from inside Room.extend so that I can mixin different costMatrix strategies.

## As-Is: pretty good without using prototypes!

All functions bind to `mod`, and `mod` has its function references overwritten... so it's fairly robust with this exception.

The only wrinkle is that if you are calling super-methods you have to fake the bind to get `this` operating correctly.

## Solution?

For all modules which assign extend to an API method:
```
module.exports = _.bindAll(mod);
```

Without that the above problem manifests itself in my custom costMatrix branch when viral is removed:
image

karlthepagan:
Related feature: https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/tree/feature-creep-custom-getpath

Internal feature coming soon.

cyberblast:
accessing members in super did work great as posted in slack.
customField in global.js, accessing it in viral.global.js override using `this` works like a charm...

karlthepagan:
@cyberblast whoops, looks like my commit removed my authority to merge.

PR #424, opened 2017-01-06
defense only for owned,reserved,exploiting,claiming&reserving rooms

cyberblast:
\+ a blacklist to exclude rooms from defense

#399


karlthepagan:
Comments inline. IMO there are 3 cases where we want to reactively defend a room:
* player controls it (room.my)
* player is trying to claim it (FLAG_COLOR.claim)
* player has plans to dismantle the room (FLAG_COLOR.invade)

Claiming, exploiting, and robbing are situations with a delicate opportunity cost and auto-defending in these cases is likely to invalidate the benefit of at least some of these scenarios.

cyberblast:
> player controls it (room.my) 

agreed

> player is trying to claim it (FLAG_COLOR.claim)

yes. we may add that...

> player has plans to dismantle the room (FLAG_COLOR.invade)

uhm yes. but no. This is not a job for a defender, but for an offender. 

why wouldn't you want to defend your remote harvesting room against invaders? Leaving it for 1500 ticks is quite expensive. Your roads (or even containers) will decay during that time...

karlthepagan:
> why wouldn't you want to defend your remote harvesting room against invaders?

If it's a player I'm stealing from.

cyberblast:
then its robbing, not exploiting ;)
we may remove robbing...
and in corner cases you can still add that room to the blacklist...

Issue #422, opened 2017-01-06
Allow hauler to spawn sooner


cyberblast:
Currently it's only allowed to spawn a hauler, when there is at least one miner. We could maybe remove that rule... or maybe better change it to "if haulers+workers >= 2" or something like that...

As defined in #412

Issue #421, opened 2017-01-06
No mineralMiner without hauler


cyberblast:
Only spawn mineralMiners when there is at least one hauler. and also reduce its priority. 
As defined in #412

Krazyfuq:
Hi cyberblast,

Seems no mineralminers are spawning now with the newest dev

Krazyfuq:
my bad, minerals are empty sorry

PR #427, opened 2017-01-06
Adjust hauler spawning priorities

cyberblast:
#421 & #422


creep.setup.hauler.js
```diff
@@ -11,8 +11,12 @@ setup.maxMulti = function(room) {
     return Math.min(max, 16);
 };
 setup.maxCount = function(room){
+    if( !room.population ) return 0;
```
karlthepagan: We should consider eliminating haulers in the case of a single source room. (because some players are in a rough neighborhood like me)

cyberblast: hmm, valid point. and not only for haulers. I will create a new issue...

PR #430, opened 2017-01-06
a player is their own ally

karlthepagan:
silly that I forgot this :)


karlthepagan:
Naming is hard. I guess I'd be looking for the most cohesion in these terms and how they're used & how they relate to the game state. So `myIcon` would mean that my icon is shown on the main map.

PR #434, opened 2017-01-07
configurable pathing strategy

karlthepagan:
this allows the pathStrategy to be replaced (and its configurable!)

requires #415


karlthepagan:
@cyberblast I can reopen this to merge into dev afte #415 is done

creep.js
```diff
@@ -279,21 +279,8 @@ var mod = {
...
-
-            let path = this.room.findPath(this.pos, targetPos, {
-                serialize: true,
-                ignoreCreeps: ignoreCreeps
-            });
-            if( path && path.length > 4 )
-                return path.substr(4);
-            else return null;
+        Creep.prototype.getPath = function() {
+            return mod.pathStrategy.handle.apply(this, arguments);
```
cyberblast: why do you use `mod` instead of `this`?

cyberblast: why extracting that function and implementing the same full code? 
Simply override getPath instead of pathStrategy ..!?

karlthepagan: The function assigned to Creep.prototype may not be bound.

Mod however must be inheritable. I was playing around with using different references and I'm not a JS expert but I think that making a `const mod = this;` reference above is the most safe and will work nicely if we move to `"use strict"`

karlthepagan: Wait a moment, I might be overcomplicating it.

Do you mean that I should re-declare Creep.prototype.getPath in my viral.room.js extend()?

karlthepagan: To answer the first question succinctly: the `this` context inside Creep.getPrototype must be the creep, so `getPath` cannot be a bound function. The strategy/handler pattern guards the function from being bound, but inheritance is used in Room (to call costmatrix super functions).

cyberblast: > the this context inside Creep.getPrototype must be the creep 

ah, right, pathStrategy is not a member of creep but only of mod. 

> Do you mean that I should re-declare Creep.prototype.getPath in my viral.room.js extend()? 

Why not? keep it simple. 
This way you could also re-declare it for specific creeps only (e.g. by re-declaring that function for a certain creep instances, maybe from within a behaviour). 
Or you re-declare it in the creep's prototype to change it for all creeps.

karlthepagan: I keep trying to favor inheritance, but I don't see a need for it in this case.

karlthepagan:
Correct close, there's no changes with the new getPath in viral.

Issue #398, opened 2017-01-04
Use memory based task assignment instead of flags


ESchuderer:
```javascript
Room.prototype.addTask = function(task, sourceRoom, targetRoom) {
    Memory.tasks = Memory.tasks || {};
    Memory.tasks[targetRoom] = Memory.tasks[targetRoom] || {};
    Memory.tasks[targetRoom][task] = {
        sourceRoom: sourceRoom
    };
}

Room.prototype.removeTask = function(task, targetRoom) {
    if(Memory.tasks[targetRoom][task]) {
        delete Memory.tasks[targetRoom][task];
    }
}

function Task(task, sourceRoom, targetRoom) {
    this.task = task;
    this.sourceRoom = sourceRoom;
    this.targetRoom = targetRoom;
}

Task.prototype.mining = function() {
    console.log("Task " + this.task + " executed for targetRoom " + this.targetRoom + " from sourceRoom " + this.sourceRoom);
}

if(Memory.tasks) {
    var taskObject;
    var taskMemory;
    
    for(var room in Memory.tasks) {
        if(!_.size(Memory.tasks[room])) {
            delete Memory.tasks[room];
            console.log("Removed empty task object for room " + room);
        }
        
        for(var task in Memory.tasks[room]) {
            taskMemory = Memory.tasks[room][task];
            
            if(taskMemory.sourceRoom) {
                taskObject = new Task(task, taskMemory.sourceRoom, room);
                taskObject[task]();
            } else {
                delete Memory.tasks[room][task];
                console.log("Removed invalid task " + task + " for targetRoom " + room);
            }
        }
    }
}
```

Quick idea how to implement something like this.

cyberblast:
uhm yes. Question is if/when we need it. I mean, the guard task is only an example. Of course there will be a lot of other tasks completely without flags. Just like the defense task. 

The question is if we want to enforce every task to use this (your example).

ESchuderer:
Of course this is looking into the future, a possible solution would be to use flags for adding and removing rooms, still utilizing the findBestRoomFor methods.

If the flag is placed in the room, add the corresponding task to Memory and remove the flag, if the task already exists and a flag is placed - remove it. Still could keep flags for dismantle sites and so on as is.

karlthepagan:
Related to this I have a serious proposal which will affect many users, but if we go down the route of using flags for commands this will drive people nuts unless addressed:

[...]

PR #438, opened 2017-01-07
FLAG_COLOR.claim is now green/* (white reserved for flag command API) #398

karlthepagan:
FLAG_COLOR.claim is now GREEN instead of WHITE (which is the dialog default).

This makes way for using flags as temporary commands (see #398).


cyberblast:
hmm you did a good work here, but I'm unsure if we need the migrate. Its a lot of code. to only migrate a few flags. Sorry. 
I'd suggest to keep it clean and simple instead and migrate that one white flag which someone may have manually...

karlthepagan:
Migration is about 25% of this patch, and  error report is another 25% - I'll trim that 50 LoC.

karlthepagan:
cleaned up and rebased on to dev

global.js
```diff
@@ -54,7 +54,31 @@ var mod = {
...
+                        color: COLOR_GREEN,
+                        secondaryColor: COLOR_RED,
+                        filter: {'color': COLOR_GREEN, 'secondaryColor': COLOR_RED }
+                    },
+                    reserve: { // reserve this room
+                        color: COLOR_GREEN,
+                        secondaryColor: COLOR_GREY,
+                        filter: {'color': COLOR_GREEN, 'secondaryColor': COLOR_GREY },
+                    },
+                    mining: { // from branch:feature/remoteMining
```
cyberblast: oops ;)
 mining: { // from branch:feature/remoteMining

karlthepagan: intentional - but I had 2 helpful choices here
* send a PR into feature/remoteMining
* leave it in with the note that it was reserved for the feature

karlthepagan: resolved

whardeman:
Could the migrate code be rolled up into an ad hoc module to be used when updating?

karlthepagan:
I'll push a migrations branch with this viral: https://github.com/karlthepagan/screeps.ocs.personal/blob/master/viral.flagDir.js

PR #441, opened 2017-01-08
Prevent divide-by-zero in Creep.Setup.js

whardeman:
If Creep.Setup.compileBody is called with `multiBody=[]` CPU limit will be exceeded until bucket empty because of divide-by-zero condition.


Dasku:
This might well be what the problem was with my task.claim as the task tried to spawn a 0 Multi body part.

Currently away but will be home later today and will test it out!

PR #444, opened 2017-01-08
changing infect to auto call base function

cyberblast:


karlthepagan:
This interferes with my motivation for requesting #415

cyberblast:
hmm. sure. if you want to replace a function instead to extend it, it would require an explicit reassignment now (this PR), which is more work than simply defining the function in your override file. But this way its also quite obvious whats going on, due to this explicit assignment. 

hmm. I think its also a matter of what we need most. 
If most of the time we want to completely replace a function, it should require an explicit call for the base function (from within the override). 
If most of the time we want to extend functions, it should require an explicit call to replace functions (explicitly). 

But I'm unsure yet about what we need ^^

karlthepagan:
The "principle of least surprise" favors an explicit call to the extended method IMO.

For extend it will be very common tho.

cyberblast:
Maybe you are right, we are not extending arrays or objects as well... 
Extending functions may not be the expected behavior.

PR #451, opened 2017-01-10
Creep.onError and Creep.actionError event

karlthepagan:
This change consolidates most action errors into `Creep.prototype.onError(tryAction, tryTarget, errorCode)`

Additionally the event Creep.actionError is raised and event listeners should signal that the error was recovered and should not be logged by calling `delete creep.resolvingError`

Example implementation: https://github.com/karlthepagan/screeps.behaviour-action-pattern/commit/60aff5612a6cf88c23327023838552625d69d53d

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -28,6 +28,9 @@ var mod = {
             reallocating:load("creep.action.reallocating"),
             recycling:load("creep.action.recycling")
         };
+        for (const action in Creep.action) {
+            if (Creep.action[action].register) Creep.action[action].register(this);
+        }
         Creep.behaviour = {
             claimer: load("creep.behaviour.claimer"),
             hauler: load("creep.behaviour.hauler"),
@@ -41,6 +44,9 @@ var mod = {
             upgrader: load("creep.behaviour.upgrader"),
             worker: load("creep.behaviour.worker")
         };
+        for (const behaviour in Creep.behaviour) {
+            if (Creep.behaviour[behaviour].register) Creep.behaviour[behaviour].register(this);
+        }
         Creep.setup = {
             claimer: load("creep.setup.claimer"),
             hauler: load("creep.setup.hauler"),
@@ -54,6 +60,9 @@ var mod = {
             upgrader: load("creep.setup.upgrader"),
             worker: load("creep.setup.worker")
         };
+        for (const setup in Creep.setup) {
+            if (Creep.setup[setup].register) Creep.setup[setup].register(this);
+        }
         Creep.loop = function(){
             var run = creep => creep.run();
             _.forEach(Game.creeps, run);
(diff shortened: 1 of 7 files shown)


creep.js
```diff
@@ -422,6 +431,21 @@ var mod = {
                 }
             }
         });
+
+        Creep.prototype.onError = function(tryAction, tryTarget, workResult) {
+            debugger;
+            if (this.resolvingError) return;
+
+            this.resolvingError = tryAction;
+            Creep.actionError.trigger({creep: this, tryAction, tryTarget, workResult});
```
karlthepagan: I would rename this onError, but that means that two things with the same name might be confusing (the "static" event and the prototype function).

creep.js
```diff
@@ -28,6 +28,9 @@ var mod = {
             reallocating:load("creep.action.reallocating"),
             recycling:load("creep.action.recycling")
         };
+        for (const action in Creep.action) {
+            if (Creep.action[action].register) Creep.action[action].register(this);
```
karlthepagan: the static register methods are not strictly part of this feature, but I think they are needed for the example I implemented. They keep the handlers cohesive (located near the code they relate to).

cyberblast: which example? I can't find a register function for actions...

creep.js
cyberblast: hmm, maybe better to generalize even more? like 
`Creep.prototype.onError = function(errorCode, debugData)`, where debugData may be something like 
```
{
actionName: 
targetId: 
somethingElse:
}
```

creep.js
karlthepagan: https://github.com/karlthepagan/screeps.behaviour-action-pattern/commit/60aff5612a6cf88c23327023838552625d69d53d <-- here's the impl

creep.js
karlthepagan: Ok iterating on that I think I would make the default implementation of onError just take an object that gets forwarded to the event callbacks.

```
Creep.prototype.onError({errorCode, action, target});

Creep.prototype.onError = function(errorData) {
    // ....
    Creep.errorEvent.trigger(errorData);
}
```

This way all the user data is propagated out to the subscribers. I might even attach utility methods to that data object using prototype...

creep.js
```diff
@@ -422,6 +431,21 @@ var mod = {
                 }
             }
         });
+
+        Creep.prototype.onError = function(tryAction, tryTarget, workResult) {
+            debugger;
+            if (this.resolvingError) return;
```
karlthepagan: I'm going to replace this API with a .preventDefault() function.

creep.js
```diff
@@ -422,6 +431,20 @@ var mod = {
                 }
             }
         });
+
+        Creep.prototype.onError = function(tryAction, tryTarget, workResult) {
+            if (this.resolvingError) return;
+
+            this.resolvingError = tryAction;
+            Creep.actionError.trigger({creep: this, tryAction, tryTarget, workResult});
```
cyberblast: hmm, maybe we can still generalize it even more by simply naming it Creep.error.  
Or even better `global.error.trigger({caller: this, type: 'creepWorkFailure', error:{tryAction, tryTarget, workResult}})`

karlthepagan: Will do. I want to try to preserve the `preventDefault()` feature.

I'll keep iterating on this idea and come up with some more robust examples. Not ready yet.

PR #456, opened 2017-01-11
syntax checks for virals

karlthepagan:
especially nice when you're editing code directly in game ;)


main.js
```diff
@@ -15,6 +15,10 @@ module.exports.loop = function () {
             mod = require(path);
         }
         catch (e) {
+            if (!(e.message && e.message.startsWith('Unknown module'))) {
```
cyberblast: Please add 
if(DEBUG && [...]

karlthepagan: I'm going to change this so that DEBUG is assumed to be true if parameter.js has not been loaded.

karlthepagan: Resolved and tested on prod.

PR #458, opened 2017-01-11
fixing FlagDir find for local flags using a callback

### LINKED ISSUE #449: defence task issue
I just started a new room, at level 2 it started spawning ranger-def creeps which left the room and into a centre room and died to a source keeper. This carried on until the colony was dead. This is similar to what happened with the enemy room which I think we fixed. 

After 300 ticks, the first spawn was another ranger-def :)

I've took out defence task until it spawns a worker again. But I don't think defenders should ever spawn for centre rooms or SK's. 

p.s. is there a command to clear spawn queues?

cyberblast:
should fix #449


flagDir.js
```diff
@@ -18,7 +18,7 @@ var mod = {
         let filter;
         if (typeof flagColor === 'function' ) {
             filter = flagEntry => ( flagColor(flagEntry) && flagEntry.cloaking == 0 &&
-                (!local || !pos || !pos.roomName || flagEntry.roomName != pos.roomName));
+                (!local || !pos || !pos.roomName || flagEntry.roomName == pos.roomName));
```
karlthepagan: The one-liner you want is: `!(!local || (pos && pos.roomName && flagEntry.roomName === pos.roomName))`

I'm in favor of this being on multiple lines:
```
filter = function(flagEntry) {
   if ( flagColor(flagEntry) && flagEntry.cloaking === 0 ) {
      if ( !local ) return true;
      if ( pos && pos.roomName && flagEntry.roomName == pos.roomName ) return true;
   }
   return false;
}
```

I'm assuming that `pos && pos.roomName` is only for error handling.

PR #462, opened 2017-01-12
Smallest reserver possible

Dasku:


pandabear41:
Why should we only spawn the smallest one possible? 
If it spawns some with more claims, it will just boost the reservation in the room and then after it ticks remaining is greater than 2500 it won't spawn any more until it lowers.

Not related to your PR, but in regards to task reservation. 
Why does the base claimer have a static body of two claims? That causes it to not spawn in lower RCLs.

whardeman:
@pandabear41 I think it's just because two CLAIM parts are required to actually increase the reservation of a controller.

Dasku:
@pandabear41 we did have the max claimer possible to start with for the reason you mentioned. We are just testing the 2 part claimers as smaller creeps mean your spawn is "in use" for less time. This could end up changing again later if we can show it works better with larger creeps.

and as @whardeman said, you need 2 parts minimum to reserve or the time does not increase, 1 part claimers are only good for claiming task.

pandabear41:
Two CLAIM are required to increase, but what if I am RCL3 and I want to boost my energy a bit by having some remote mining with a full 3000 energy source. It might not increase, but it will certainly just be reserved. 
We could put in a special case if the room can't spawn a two CLAIM reserver, it could just spawn two creeps with a single CLAIM.

Dasku:
Definitely an interesting idea, maybe we can create an issue for a future improvement?

karlthepagan:
+1 that sounds like a remote mining strategy

PR #464, opened 2017-01-12
Feed spawn before extensions

Dasku:
To allow faster spawning of next creep


cyberblast:
I don't think this will work. 
The filter function will always only return true or false. No matter how you are designing the expression. It will not cause a prioritization of spawns. 
It also doesn't matter where that energy is. Energy within extensions can be used by any spawn as well as energy within any spawn can be used by any spawn. That means even a spawn with 0 energy is able to spawn. As long as there is enough energy (sum) within all extensions + all spawns.

Dasku:
Ahh ok thanks

Issue #206, opened 2016-10-10
Improve remote mining


SirLovi:
Having dedicated miners right next to spawns with containers in exploited rooms and then having remote-haulers to pick up the energy could be more energy efficient.

![capture](https://cloud.githubusercontent.com/assets/20193391/19236472/efbf286e-8ef7-11e6-8488-02739481e353.PNG)

cyberblast:
Sure, why not... 

* reserving is working already (white/gray flag). 
* miners + haulers for exploitation rooms 
  * correlates with the "Operator"  idea (merge workers with privateers & pioneers) 
  * miners are probably too slow, may need some additional MOVE parts eventually 
  * may be worth a try... 

I'd propose to send 1 miner and 1 hauler + 1 privateer (as hauler and mechanic) per exploit flag. 
Current best practise is to put 1 flag per room, would be 1 per source then. 

On the other hand... remote harvesting rules are about to be changed drastically soon (no more boost by reserving). Maybe it will be better to wait until then? I'm unsure how to face it...

erik-smit:
I took an ugly stab at this at https://github.com/erik-smit/screeps.behaviour-action-pattern/commit/0f46214936f3fff86c869e53cdc2e6a1768e4312

I turned the privateer in a dedicated miner and added a 'haulateer' for the hauling.

Issues I ran into:

  * privateer cloaks flag when source is empty, preventing flagDir from finding it for the haulateer
  * privateer also does the building and maintaining in the exploited rooms. With current code it still does, until the container is built. Who will do it when there is a dedicated miner?

cyberblast:
Maybe we could do it like we did with owned rooms. 
In the beginning there are only worker. Then we add miner and hauler, keeping one worker. 

The alternative could be to add one work part to the hauler and repair on drive-by. 
But construction would be quite slow with that. But we could use pioneers for that. or simply a worker. Those types could be merged anyway (worker, privateer, pioneer).

Maybe we can also use the current hauler setup, adding a new action to get the energy from remote rooms. Similar for miner. 

I'd propose to wait until we have the spawnQueue. I'm currently working on that. Will make it possible to also set a destiny for the new creep. This can be evaluated one the creep finished spawning (e.g. for the miner to assign it to a room).

Dasku:
[...]

Issue #445, opened 2017-01-09
Remote mining branch ready for testing


whardeman:
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/tree/feature/remoteMining

With a WHITE/BROWN flag, miners and haulers will spawn and they will go to the target room and perform work. A couple of points:
- This is still pretty unstable and needs some more work.
- Definitely need to optimize creep sizes and memory mgmt.
- These creeps don't repair-on-driveby, so roads/containers decay. That need to get sorted.
- Also, the mining flag doesn't spawn a claimer, so you'll need to place a WHITE/GRAY flag with each of your WHITE/BROWN ones.
- Last thing: If you place a flag, not much will happen until you have visibility in the room. I'd recommend starting with privateers to get roads built out and to start reserving the rooms. Then place mining flags.

Notes about the code:
- Remote miner behaviour is just miner wrapped with an initial travel action.
- Remote hauler is just a stripped-down privateer.
- Both could be cleaned up I'm sure.

Any and all contributions are welcome. I'm not attached to the way I've done things. This is just best I could do to move this feature forward.

whardeman:
Another thought about roads in exploit/remote mining rooms: Skip 'em.

I've setup haulers to have enough movement that they are at zero fatigue even when full of energy. If we just add a couple move parts to miners, they can still reach their targets quickly at a very low cost. That's a lot less effort than building/maintaining roads.

karlthepagan:
I'll have to track this down. It looks like bestSpawnRoomFor is returning a room which I don't have vision to.

```
[6:48:12 AM]TypeError: Cannot read property 'structures' of undefined
    at Object.checkForRequiredCreeps (task.mining:108:74)
    at Object.handleFlagFound (task.mining:37:25)
    at Flag.found.on.flag (task.mining:4:44)
    at handlers.slice.forEach.h (global:28:57)
    at Array.forEach (native)
    at LiteEvent.trigger (global:28:44)
    at register (flagDir:81:24)
    at baseForOwn (/opt/engine/node_modules/lodash/index.js:2046:14)
```

whardeman:
@karlthepagan That's possible, I suppose. There's an undefined check for the mining room, but I hadn't considered that there might be an invisible room returned for spawnRoom.

Has bestSpawnRoom been changed recently such that it might return a non-owned room?

karlthepagan:
[...]

PR #466, opened 2017-01-12
Feature/remote mining

whardeman:
#445


karlthepagan:
I think that any outstanding issues are best resolved by new PRs.

whardeman:
@karlthepagan Are you just generally commenting on the readiness of this branch?

karlthepagan:
Possible issues to address later?

* remote workers should only spawn if there are construction sites, and should fallback to hauler
* replace the hauler base body with  [work,carry,move] if any roads in target room are visible. Tune setups to make this work.
* hauler should flee from source when idle (evade idle?)
* refactor spawn queue memory to account for users manually clearing the spawn queue
* other spawn refactor problems (priority inversion)

whardeman:
Yes, I have a list of those as well. Although it's mainly the economic question of roads vs haulers with more MOVE parts.

Dasku:
@whardeman I'm pretty sure someone will have done the math on this and if MOVE was more efficient than maintaining roads, surely everyone would do it and roads would become pointless and no one would use them?

karlthepagan:
Let's leave that up to the user. Also my comment was partly about readiness and partly about getting more people iterating on this feature so we can evaluate the harder economic decisions.

Dasku:
@whardeman I asked in help and got this as one response:

"@asku I don't know if there's exact math on how _energy_  efficient it is, but even if it were exactly equal it'd be worth it to do roads to save on spawn time

[5:34]  
if your haulers can have half the move parts, your spawns can produce that many more haulers and mine that many more rooms"

Also in my mind, all creeps share the cost of maintaining roads, they all use them, and as pointed out, it means you need half as many MOVE parts on each creep.

Dasku:
image

cyberblast:
> * remote workers should only spawn if there are construction sites, and should fallback to hauler

or recycle when done

> * hauler should flee from source when idle (evade idle?) 

In a perfect world there shouldn't be something like an idle hauler. If there is an idle one we may have too many of them. 
For remote haulers I think it would be the simplest approach to simply ping pong back and forth. Grab what you can and get it home. Running a 0 tour is sad but a definite sign of too many haulers.

whardeman:
I'll open a new issue to discuss improving remote mining. Let's get this one merged and closed down.

Thanks everyone for the great input.

karlthepagan:
Remote mining claimer ratio should default to zero. The claimer task is frequently fighting with this task's claimers. (spammy in the logs)

TafThorne:
This mostly resolves #206 which was closed two days ago.  The good news is that remote mining is improved with the caieates noted above.  

> remote workers should only spawn if there are construction sites, and should fallback to hauler

If we do that what ends up maintaining roads?  Maybe there is another ticket somewhere which will tell me this.

Methel:
Im getting a lot of  "Mining Flag tried to queue a zero parts body claimer. Aborted" errors, not quite sure how to debug that. 
It's spawning remote miners, haulers and workers but not claimers. Not related to new rooms unable to spawn big enough claimers.

Dasku:
Hi Methel,

What RCL is the room in question?

Methel:
rcl 7

Dasku:
May be worth coming on slack to discuss if you can, Can see if others are having the issue. Looks odd to me, there is no way that room should not be able to queue that creep.

whardeman:
@Methel We should have another issue specifically related to fixing this. It's a bug probably due to either maxCount or multiple tasks trying to spawn claimers.

Dasku:
I will submit a PR for including mining flags for reserving task later today

cyberblast:
> I will submit a PR for including mining flags for reserving task later today

? But it's included already... 
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/blob/dev/task.reserve.js#L20

Dasku:
mining flags is commented out, current its only exploit and reserve flags

Dasku:
May help if we merge this branch and fix in dev?

cyberblast:
Ah, I see, there is a new Flag color for remote mining.
But I don't think we should keep both mining approaches...

TafThorne:
Having the experimental mining flag temporary would let people test the
code on a few rooms rather than all or nothing. Id agree that one flag
would be better in the long term once the code is reliable.

PR #467, opened 2017-01-12
picking is better than idle

karlthepagan:
this should be extremely rare


karlthepagan:
Yup, this is if the worker (and hauler) has > 50% energy but still room to store.

pandabear41:
This is good!
It only apply to RCL1 though.

PR #470, opened 2017-01-12
fix maxCount based on flags

karlthepagan:
this fixes the issue where all flag-based spawns come in waves

this was non-trivial... so it also adds Creep.Setup.debugType. With DEBUG enabled, set Creep.Setup.debugType = Creep.setup.ranger.type in order to see the valid setup decisions printed.

### DIFF
--- a/creep.Setup.js
+++ b/creep.Setup.js
@@ -61,20 +61,26 @@ var Setup = function(typeName){
         return memory;
     };
     this.isValidSetup = function(room){
-        if( room.controller.level < this.minControllerLevel )
+        if( room.controller.level < this.minControllerLevel ) {
             return false;
+            if (DEBUG && this.type === Creep.Setup.debugType) console.log(room.name, Creep.Setup.debugType, 'low RCL');
+        }
 
         let rcl = this.RCL[room.controller.level];
         let minAbsEnergyAvailable = this.SelfOrCall(rcl.minAbsEnergyAvailable, room);
         let minEnergyAvailable = this.SelfOrCall(rcl.minEnergyAvailable, room);
         if( room.remainingEnergyAvailable < minAbsEnergyAvailable ||
-            room.relativeRemainingEnergyAvailable < minEnergyAvailable )
+            room.relativeRemainingEnergyAvailable < minEnergyAvailable ) {
+            if (DEBUG && this.type === Creep.Setup.debugType) console.log(room.name, Creep.Setup.debugType, 'not enough energy');
             return false;
+        }
 
         let maxCount = this.SelfOrCall(rcl.maxCount, room);
         let maxWeight = this.SelfOrCall(rcl.maxWeight, room);
-        if( maxCount == 0 || maxWeight == 0 )
+        if( maxCount == 0 || maxWeight == 0 ) {
+            if (DEBUG && this.type === Creep.Setup.debugType) console.log(room.name, Creep.Setup.debugType, 'zero count || weight');
             return false;
+        }
         if( maxCount == null )
             maxCount = Infinity;
         if( maxWeight == null )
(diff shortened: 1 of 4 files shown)


karlthepagan:
Taking a look at some issues with task-based spawning I'm having now.

karlthepagan:
My issue was a bad feature rebase.

creep.Setup.js
```diff
@@ -217,4 +225,32 @@ Setup.compileBody = function (room, fixedBody, multiBody, sort = false) {
...
+            }
+            if( !measureByHome ) {
+                max++;
+                return;
+            }
+            flag = Game.flags[flagEntry.name];
+            if( !flag.targetOf || flag.targetOf.length == 0 ) {
+                max++;
+            } else if( _.some(flag.targetOf, 'homeRoom', room.name) ) {
+                max++;
```
cyberblast: Why are you allowing an additional creep if there is another creep already assigned to that flag, which has been spawned within the same room?

karlthepagan: because this is compared to the population which is spawned in the room.

For example - I have 2 defender flags. When there are no defenders spawned there. The spawn then builds 1 defender, the previous algorithm immediately reduced the maxCount to 1. Then the maxCount was equal to the count of defenders spawned and the 2nd flag would not have an assignment.

cyberblast: But shouldn't it then always simply return the max total allowed amount, which is simply the sum of all flags (of that color)? And completely ignore flag.targetOf. 

There are 3 flags, it's allowed to have 3 creeps. 
flag.targetOf would then only be relevant for targetting (which flag to approach).

karlthepagan: The problem is, as soon as the first of those 3 creeps spawns your `maxCount` **decreases** while your `currentCount` simultaneously **increases**. This appears to work if you have 3 spawns, then 3 creeps come out and as soon as they hit the street their targets get `targetOf` set and that affects spawning logic. IMO you do want this feature to prevent two rooms from fighting over a flag, and so that explains why I don't simply max++ for every flag.

Test it! ;) It's impossible to stagger-spawn 2 creeps with the `dev` branch. They always come in waves.

cyberblast: But isn't it wrong to increase **and** decrease simultaneously? 
Wouldn't it be correct to freeze `max` at flag count and only increase `current` count once a creep starts spawning? 

And for "measure by home", in some cases there have been complex rules which calculate max by taking into account other owned rooms and otherflags nearby (privateerMaxWeight, claimerMaxWeight). 
Other cases, where we didn't have those complex rules and measureByHome set to true it simply didn't really work (guards). 

Thats one of the reasons why we started to implement the task/event based system. To ensure every single flag/trigger gets its creep/attention. Without dirty estimation formulas to get load balanced. We had greedy spawns which tried to spawn and always had too look from their perspective, which is simply wrong. Instead we are now implementing the other way round. We have the event in focus. Now we need to get a better method to find the "best" room to spawn "something", which is comparably easy. Maybe also add some more flexibility for handling the queues, agreed. But I wouldn't try to keep those old mechanisms alive. We need to get rid of them. 

So in a mid range, we should try to get rid of those "maxCount per creep type" at all. We have a requesting task event, "ordering" its creeps. Each task may have its completely different logic and code, and there may be some of them measuring the total (or local) creep type population. But in most cases I don't think we really need it anymore (luckily).

cyberblast: But regarding this PR, I think it's fine. ;)

karlthepagan: > isn't it wrong to increase and decrease simultaneously?

yes, but that's what happens `without` this patch ;)

The ideal fix is that we only increase the `local currentCount` - while we decrease the `local maxCount` for every other room which is looking at this flag. I think i'm really close but I need to update spawning creeps so that they preemptively set their target flags.

karlthepagan: > Thats one of the reasons why we started to implement the task/event based system. To ensure every single flag/trigger gets its creep/attention. Without dirty estimation formulas to get load balanced.

I really like the setup / estimation approach because it should scale better. I'd love to continue to debug these so that you can keep the task systems simpler.

cyberblast: > yes, but that's what happens without this patch ;) 

yes I know. With all it's downsides ;) (as described)

PR #472, opened 2017-01-12
Ensures claimers are prioritised

Dasku:
gives more priority if ticks are less than 500 or unreserved.


Dasku:
An issue with this is that if a creep is already queue'd in low  its validated so it won't switch to medium queue unless the tasks are refreshed somehow on a semi regular basis.

Dasku:
Could only validate the low queue in andleSpawningStarted to ensure it could always push to medium queue

PR #477, opened 2017-01-14
refactoring mining task

cyberblast:
A lot of clean up and refactoring.

Also added:
* only spawn a worker, when there are construction sites
* clean up memory when flag removed
* 1 work part for remoteHauler
* 1 carry part for remoteMiner


Dasku:
I get 
ReferenceError: carrySum is not defined
    at Object.nextAction (creep.behaviour.remoteHauler:26:43)
    at Object.run (creep.behaviour.remoteHauler:7:18)
    at Creep.run (creep:152:43)
    at run (creep:58:38)
    at Function.Creep.loop (creep:59:15)

whardeman:
@cyberblast I like the work going into this refactor, but it looks like creep bodies are being hard coded into the task method for determining when to spawn. If it helps, I'm happy to assist in converting those to "infectable" properties.

karlthepagan:
That would be very helpful!

cyberblast:
oh yes, thats nice. 
there is still something wrong with the hauler :/

cyberblast:
thats better

cyberblast:
seems to be running quite stable now...

task.mining.js
```diff
@@ -2,182 +2,161 @@ var mod = {
...
+            delete memory.queued.worker;
+        }
 
-    nextAction: creep => {
-        // if not in the target room, travel there
-        if( creep.room.name != creep.data.destiny.room ){
-            Creep.action.travelling.assign(creep, Game.flags[creep.data.destiny.flagName]);
+        return memory;
+    }, 
+    spawn: (flag, type, fixedBody, multiBody, register) => {
```
whardeman: Is there planned functionality to use the register argument or is this magically used somehow that is not obvious to me?

cyberblast: oh, no its a bug. Should be removed.

cyberblast:
yes, once assigned the recycling action will get executed. 
I think we can file that as future tuning potential.

creep.behaviour.remoteHauler.js
```diff
@@ -17,79 +17,67 @@ module.exports = {
...
-
-                // carrier full
-                }else {
-                    this.goHome(creep);
-                    return;
-                }
+        // at target room
+        else if( creep.data.destiny.room == creep.pos.roomName ){
+            // get some energy
+            if( creep.sum > (creep.carryCapacity*0.8) || !(this.assign(creep, Creep.action.picking) || this.assign(creep, Creep.action.uncharging)) ){
```
whardeman: I don't think the conditions in this are working out, but I don't have an easy way to test it. If I'm not mistaken, the creep should only go home if one of the conditions is true, but that would mean the creep goes home when the `uncharging` action is successfully assigned.

whardeman: After further review, is `this.assign(creep, Creep.action.uncharging)` actually resulting in false (and thus preventing the creep leaving) because this.action expects a target argument, which is not being supplied?

cyberblast: no, the creep should go home if both are false. 
thats why its !( picking || charging ). 
But I agree, it's a bit confusing. I will rearrange that. 

That target parameter is optional and commonly not used (within "classic" actions). It will execute action.newTarget if it's not defined.

PR #482, opened 2017-01-14
Pioneer task 

Dasku:
Replaces current pioneer setup. Works the same.

Noticed 1 issue though, rooms will try and spawn their own pioneers due to the bestSpawnRoomFor once up and running.


task.pioneer.js
```diff
@@ -0,0 +1,167 @@
...
+        if (flag) {
+            // get task memory
+            let memory = Task.pioneer.memory(flag);
+            // clean/validate task memory running creeps
+            let running = []
+            let validateRunning = o => {
+                let creep = Game.creeps[o];
+                // invalidate old creeps for predicted spawning
+                // TODO: better distance calculation
+                if( creep && creep.name != name && creep.data !== undefined && creep.data.spawningTime !== undefined && creep.ticksToLive > (creep.data.spawningTime + (routeRange(creep.data.homeRoom, flag.pos.roomName)*25) ) ) {
```
cyberblast: hmm, there is a creep.data.predictedRenewal calculated in line 105. 
But yes that's also missing in task.guard. 
Will fix both... 

Also, its 25 steps from room center to the border, but 50 from room center to room center. Thus estimating routeRange * 50 gives a rough approximation.

PR #484, opened 2017-01-14
Allow privateers to work in mining rooms

Dasku:
Allows privateers to use containers instead of harvesting when sent to a room being remote mined.


cyberblast:
but that is already included in robbing...

Dasku:
hmm sorry.

whardeman:
@cyberblast Would that not require the robbing flag to be set as well?

cyberblast:
oh, yes you are right. it's currently validating on that in isValidAction of the robbing action. 
We could maybe remove that validation...

creep.action.robbing.js
```diff
@@ -2,7 +2,7 @@ var action = new Creep.Action('robbing');
 action.maxPerTarget = 2;
 action.maxPerAction = 10;
 action.isValidAction = function(creep){
-    return ( creep.sum < creep.carryCapacity && (FlagDir.find(FLAG_COLOR.invade.robbing, creep.pos, true) != null) );
+    return ( creep.sum < creep.carryCapacity );
```
cyberblast: or we could do `return ( creep.sum < creep.carryCapacity && !creep.room.my);` to allow robbing in any non-owned room, preventing privateers to steal from our homes...

Issue #480, opened 2017-01-14
trace debugging


karlthepagan:
Demo forming up here:
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/tree/debuggingWithTraces

problem: you want to add debugging but it's a lot of effort to track down all the places you need to see

solution: use objects like where clauses:

Logging statement in code (creep.js:158)
```
if( TRACE ) trace('creep',{creepName:this.name,creep:'behaviour'}, behaviour && behaviour.name);
```

Querying this in the game:
```
Memory.debugTrace.creep = 'behaviour'; // prints what behaviour the creep selects
Memory.debugTrace.behaviour = true; // prints ALL internal behaviour logic for all creeps
Memory.debugTrace.creepName = 'miner-850-1'; // restricts match to just this miner
```

So basically the conditions which are present in the log statement must match what's set in `Memory.debugTrace` or they must be absent.

Still working this out in order to gain better insight into spawning.

karlthepagan:
rebased onto the refactor

karlthepagan:
I think #490 is ready.

cyberblast:
@karlthepagan can you please add a wiki page describing how to configure trace filters in Memory.debugTrace?

karlthepagan:
Very basic docs started here: https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/wiki/Log-filtering-using-Memory.debugTrace

karlthepagan:
This is implemented.

PR #490, opened 2017-01-15
trace debugging method

karlthepagan:
Implementation for #480

example traces: https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/tree/debuggingWithTraces (will need rebase, so I'll push when this is ready)


global.js
```diff
@@ -1,3 +1,18 @@
+const reduceMemoryWhere = function(result, value, key) {
+    const setting = Memory.debugTrace[key];
+    if (setting === undefined) {
+        return result;
+    } else if (result) { // default result
+        return setting === value;
```
karlthepagan: all parameters present must match and there must be at least one

karlthepagan:
Example for CLI:

```
Memory.debugTrace.creep = true; // displays all actions in the "creep" category (action & behaviour selection)
Memory.debugTrace.no = {creepName:'hauler-1200-1'}; // ignores this creep
```

There's a lot of logic that I avoided implementing in the "where clause" logic. Consider this an exercise to the user to add callbacks or regex's.

global.js
```diff
@@ -168,6 +183,13 @@ var mod = {
                 let text = dye(CRAYON.system, roomName);
                 console.log( dye(CRAYON.system, `<a href="/a/#!/room/${roomName}">${text}</a> &gt; `) + message );
             },
+            trace: function(category, entityWhere, ...message) {
+                if(! DEBUG) return;
```
cyberblast: hmm I guess you mean `if(!TRACE)` ? or maybe `!(DEBUG&&TRACE)` ?

karlthepagan: In the pattern that I'm using, I guard TRACE outside the method (preventing object creation), but I guess it wouldn't hurt to do it inside.

karlthepagan: @cyberblast after seeing some more error conditions I think that trace should work as-is. We can use the same function for exceptional case logging, and the practice writing these categories will hopefully translate into more trace logging written.

karlthepagan: However I see that overloading the word trace is going to cause friction so I'll add 2 logging methods, one for error conditions and one for traces.

karlthepagan:
Adding trace log conventions will allow users to temporarily suppress spammy error conditions.

karlthepagan:
rebased onto the refactor

PR #491, opened 2017-01-15
unifying task creep definitions

cyberblast:


Dasku:
Using this now, seems to work fine, will update robbing task with new definitions too

PR #501, opened 2017-01-15
set tasks minRCLs

cyberblast:
unsure about the claim/reserve limits. how was that again? 😀


Dasku:
Claim is RCl 3, reserve is RCL 4

PR #516, opened 2017-01-16
define count for handleSpawningCompleted

Dasku:
stop the undefined error


task.reserve.js
```diff
@@ -85,6 +86,8 @@ module.exports = {
     // when a creep completed spawning
     handleSpawningCompleted: creep => {
         // ensure it is a creep which has been requested by this task (else return)
+        let count = memory.queued.length + memory.spawning.length + memory.running.length;
```
cyberblast: what do you need it for? I think we can remove it... also in line 110

Dasku: Ahh ok, makes sense

Issue #522, opened 2017-01-16
Checking room controller before spawning second Reserver


damintsew:
Fix from Reserving Improvement Thread 'https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/515'

When queuing a second reserver, check the controller to ensure there is space for 2 reserver's. If not, delete the queued reserver in low queue when spawning from Medium.

cyberblast:
Not implemented in that way. 
see comments in #523 & implementation in #555 instead.

PR #523, opened 2017-01-16
Checking room controller before spawning second Reserver #522

damintsew:
Checking room controller before spawning second Reserver #522

### DIFF
--- a/room.js
+++ b/room.js
@@ -340,6 +340,15 @@ var mod = {
                     return this.memory.sourceAccessibleFields;
                 }
             },
+            'controllerAccessibleFields': {
+                configurable: true,
+                get: function() {
+                    if( _.isUndefined(this.memory.controllerAccessibleFields)) {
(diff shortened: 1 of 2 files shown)


cyberblast:
hmm, yes. but then you won't queue a medium claimer 😉 
The reason for the medium queue is, that the low queued doesn't seem get spawned (due to other spawn jobs). Thats why we want to elevate it...

damintsew:
So in this situation will be spawned two Claimers ?
Or it will be evicted from low queue ?

cyberblast:
yes, currently it will keep the first claimer in that low queue, which may eventually spawn as well. 

So the striven solution would be: if the controller is low on reservation and there is a creep for that task in the low queue, then remove that creep from low and put it into medium. 

This way we don't have a problem with 1 spot controllers AND we have that priority elevated.

cyberblast:
closing this one. Implemented in PR #555

PR #527, opened 2017-01-16
swap to pioneer flag once claimed

Dasku:


creep.action.attackController.js
```diff
@@ -23,8 +23,20 @@ action.newTarget = function(creep){
     if( !creep.flag.room || creep.flag.pos.roomName != creep.pos.roomName){
         return creep.flag;
     }
-
-    return creep.flag.room.controller;
+    if( creep.flag.room.controller.my ) { 
+        // already claimed, change flag
```
cyberblast: maybe we could add a second secondary color to specify if we want to claim or reserve the liberated room...

Dasku: That's a good idea. I'll update this when I get home from work

Dasku: If we reserve it, do we change the flag to mine? or reserve? (reserve feels like a waste)

cyberblast: could also be a tactical reserve

Dasku: Can do both sure

PR #530, opened 2017-01-17
Cap multi-body to avoid going over 50 part limit

OverlordQ:
```
[10:11:37 PM]W2N5 > Offspring failed: ERR_INVALID_ARGS
 - body: {"carry":33,"move":17,"work":1}
 - name: hauler-2600-1
 - behaviour: hauler
 - destiny: undefined
```

Testing on local server, ran into case where it was trying to spawn 16 multibodies, then adding in the fixedBody put it at 51 total parts which is over the 50 limit.


cyberblast:
Obsolete since #570

PR #531, opened 2017-01-17
withdraw energy isn't available until the next tick, so double threshold

karlthepagan:
upgraders are missing a tick every time they withdraw

to fix this double the withdraw threshold


cyberblast:
Thats why we don't withdraw when it's at 0, but has only 1 last shot

karlthepagan:
I've tested it some time this morning: https://screeps.com/a/#!/history/E49N82?t=16808554

You can see me swap back and forth, sometimes the finishing tick never happens.

cyberblast:
uhm, yes - the creep switching back and forth is the worker right? 
But I didn't see an **upgrader** missing a shot... !?

Upgraders would never move back and forth. even without any energy...

karlthepagan:
no the upgraders missed shots, let's review this live right now

cyberblast:
your container is empty...

karlthepagan:
well, I'll make a gif if you can't test it

cyberblast:
Yes please. I watched your replay a few times now and couldn't see a missing shot :/

karlthepagan:
Failed recently because my upgradder's work count is a divisor of 50.

karlthepagan:
I'm mistaken. Thanks for the closer look.

cyberblast:
no worries

PR #534, opened 2017-01-17
Feature/task diplomacy

karlthepagan:
some feedback on an internal module as well as discussion about NPC specific combat lead me to create this

Task.diplomacy has several utility methods to make reaction attitudes quickly. As well, it adds a memory caching system which tasks can use (in addition to memory). The cache is a good example of how to merge memory and in-code settings.


cyberblast:
uhm. I'm unsure if this is a "task"... 
What does it do? What does it react on? "In the world of the creeps" 

Looks more like a helper class?

karlthepagan:
What is a Task anyways? I'm not sure yet.

This needs to be a helper class with memory. The only other construct like that is a fat component like Population, flagDir, etc.

karlthepagan:
rebased onto the refactor :)

karlthepagan:
rebased onto upstream/dev

cyberblast:
whats the overall idea and plan behind it? 
You want to observe hostile acts and reduce player reputation score. 
How would they raise it?

karlthepagan:
The user is able to set the reputation of players after observing their actions and deciding how well they trust them.

The task itself would dynamically adjust that (e.g. velocity of hostile acts might put another player quickly negative). This task then affects the room population (ally & hostiles list) and could probably send game notifications when other players cross certain boundaries.

It isn't a task that directly manages creeps it will only observe them. So in that way it is somewhat different.

This also introduces a task dependency into room.js so if you really want it to be a top-level module for that reason I can do it.

cyberblast:
No thats fine. 
But maybe we can rename it. I mean diplomacy should be something more formal, political? I guess this is somehow more like a task.security ... !?

karlthepagan:
I'll rename it Task.reputation

PR #537, opened 2017-01-17
robber fix 

Dasku:
task functions hotfix


creep.behaviour.privateer.js
```diff
@@ -6,7 +6,10 @@ module.exports = {
         if( creep.action == null  || creep.action.name == 'idle' ) {
             if( creep.data.destiny && creep.data.destiny.task && Task[creep.data.destiny.task] && Task[creep.data.destiny.task].nextAction )
             Task[creep.data.destiny.task].nextAction(creep);
-
+            
+            if( creep.data.destiny && creep.data.destiny.task && Task[creep.data.destiny.task] && Task[creep.data.destiny.task].exploitNextRoom )
+            Task[creep.data.destiny.task].exploitNextRoom(creep);
+
```
cyberblast: hmm, no. nextAction will call exploitNextRoom if it is required. I don't think it should be called here explicitly. It will override the assigned action from nextAction.

PR #540, opened 2017-01-17
Dont reserve SK rooms

Dasku:
isSKRoom identifies the centre 9 rooms


cyberblast:
👍 
But I think Karl is right, we should keep things named correct. 

How about those room functions: 
* isCenterRoom // only the 5/5 center room
* isSKRoom // the 8 rooms around the center room
* isHighwayRoom // a room with a 0 coordinate
* isControllerRoom // "common" rooms with a controller. all rooms which are not one of the above.

Dasku:
Ok I've added CenterNine as a block, SK as the 8 around center, center is still there and highways around the edges.

cyberblast:
hmm, this will not work

`return ((parsed[1] % 10 >= 4) && (parsed[1] % 10 <= 6) && (parsed[1] % 10 !== 5)  ) && ((parsed[2] % 10 >= 4) && (parsed[2] % 10 <= 6) && (parsed[2] % 10 !== 5));` 

you are not allowing rooms with x != 5 && y != 5 (cross). so effectively you are only allowing 4 rooms. 
Going to change it...

PR #555, opened 2017-01-18
move queued reserver to medium queue upon low reservation

cyberblast:
Solving #522 & #523

### DIFF
--- a/task.reserve.js
+++ b/task.reserve.js
@@ -39,13 +39,28 @@ mod.checkForRequiredCreeps = (flag) => {
     // count creeps assigned to task
 
     let count = memory.queued.length + memory.spawning.length + memory.running.length;
-    // Allow a second claimer in medium queue if reservation low
+    
+    // check reservation level
     let lowReservation = ( !flag.room ||
             (flag.room.controller && !flag.room.controller.reservation) ||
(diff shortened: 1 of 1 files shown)


whardeman:
We should also check `room.controller.accessibleFields` > 1 if we're going to spawn a second reserver before the first one is close to expiring.

task.reserve.js
```diff
@@ -39,13 +39,28 @@ mod.checkForRequiredCreeps = (flag) => {
...
 
     let count = memory.queued.length + memory.spawning.length + memory.running.length;
-    // Allow a second claimer in medium queue if reservation low
+    
+    // check reservation level
     let lowReservation = ( !flag.room ||
             (flag.room.controller && !flag.room.controller.reservation) ||
-            (flag.room.controller && flag.room.controller.reservation && flag.room.controller.reservation.ticksToEnd < 250)) && count == 1; // always let it queue one in low as well
+            (flag.room.controller && flag.room.controller.reservation && flag.room.controller.reservation.ticksToEnd < 250)); 
+
```
whardeman: The declaration above is already pretty complex, but I want this to return false if `room.controller.accessibleFields` < 2. Thoughts? At a minimum, it deserves commenting.

whardeman:
I'll make a second PR and you guys can refactor and clean it up later. Just need to stop wasting so much energy on useless reservers.

Issue #529, opened 2017-01-17
assigning actions usually does not clear path or target selection


karlthepagan:
for example
`Creep.action.harvesting.assign(Game.creeps['worker-200-2'], Game.getObjectById('guid of a mine'))`

this should always put the worker on that mine, selecting an open square to mine in

the creep should immediately discard its path and move to the new target

cyberblast:
Yes. it's commonly done within the behaviour, after a new action has been assigned. But I agree, this should be moved into action.assign.

cyberblast:
implemented with #556

PR #556, opened 2017-01-18
clear creep path upon each action assignment

cyberblast:
solving #529

### DIFF
--- a/population.js
+++ b/population.js
@@ -71,13 +71,17 @@ mod.registerAction = function(creep, action, target, entry) {
     else this.actionWeight[action.name] += entry.weight;
 
     let targetId = target.id || target.name;
+    let oldTargetId;
     if( entry.targetId ) {
         // unregister target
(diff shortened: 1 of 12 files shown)


population.js
```diff
@@ -87,6 +91,11 @@ mod.registerAction = function(creep, action, target, entry) {
             target.targetOf = [entry];
         else target.targetOf.push(entry);
     }
+    // clear saved path
+    if( targetId != oldTargetId ) {
```
karlthepagan: also compare actionId here?

cyberblast: why?

karlthepagan: I'm not familiar enough to know if multiple actions could have the same target. I'll take a closer look.

For example, if a privateer going to uncharge storage is manually assigned to dismantle.

cyberblast: callign assign will always assign a target as well (defined as argument or it will call newTarget). Same target same path... ;)

cyberblast: callign assign will always assign a target as well (defined as argument or it will call newTarget). Same target same path... ;)

PR #559, opened 2017-01-19
Don't queue a second reserver if controller.accessibleFields < 2

whardeman:
Resolves my comment in #555


cyberblast:
We are not spawning a second claimer. Not after merging #555

whardeman:
@cyberblast Ah, that's good. I didn't quite understand #555, but now I see you're right. Closing.

PR #567, opened 2017-01-19
Move FlagDir operations before Population as Population checks depend…

OverlordQ:
… on flags being present to validate actions


cyberblast:
hmm. where? 
I couldn't find a reference to FlagDir. 
Only to the common Game.flag object, populating the flag.targetOf array. 
It shouldn't matter...

OverlordQ:
Since population is checked before flags, in `population.validateAssignment` it calls `action.validateActionTarget` which checks to see if the current action is valid. For the `invade` action, the `isValidAction` check requires a invade flag. Since FlagDir hasn't been run yet, `hasInvasionFlag` will return false, so an invader with flap between invade/idle.

cyberblast:
Yes, you are right. 

So I'd then call it even as last (for the analyze). 

But I'd like to keep it at first for the execute. This will cause reactions on population changes first, prior to calling flag tasks. This way these tasks will have an up to date status about their creeps in their memory (if hooked on creep died etc...)

PR #569, opened 2017-01-20
multiple changes

karlthepagan:
* Action.chat()
* miner travel
* upgrader controller sign and upgrade on approach
* worker travel tweaks
* Creep.compileBody supports maxMulti
* Room.findSpawnRoom fixes and tracing
* Task.spawn supports fixed spawn room
* task.mining uses Room.findSpawnRoom
* remote miner can spawn as 4 work, but will try for 5 work with enough move to arrive efficiently via roads


karlthepagan:
Apologies for not breaking these up into individual branches... but I'm headed out and wanted to give you a shot at a single merge.

task.mining.js
```diff
@@ -153,8 +182,9 @@ mod.memory = key => {
 };
 mod.creep = {
     miner: {
-        fixedBody: [MOVE, MOVE, MOVE, WORK, WORK, WORK, WORK, WORK, CARRY],
-        multiBody: [],
+        fixedBody: [MOVE, MOVE, WORK, WORK, WORK, WORK, CARRY],
+        multiBody: [MOVE, WORK, WORK],
```
whardeman: I would switch this to 5 WORK in fixed and 1 WORK in multi. Still only requires a minimum of 650.

task.mining.js
```diff
@@ -153,8 +182,9 @@ mod.memory = key => {
 };
 mod.creep = {
     miner: {
-        fixedBody: [MOVE, MOVE, MOVE, WORK, WORK, WORK, WORK, WORK, CARRY],
-        multiBody: [],
+        fixedBody: [MOVE, MOVE, WORK, WORK, WORK, WORK, CARRY],
```
torin555: I'm not sure a remote miner with 4 WORK parts should ever be spawned, it'd be wasting energy. it's better to just wait for the extra 100 energy than to lose 1/5th of the potential energy the creep would mine.

task.mining.js
karlthepagan: Without reserving 4 work makes some sense and will allow remote mining at GCL 2. Please discuss

torin555: I don't think it's worth that. Because there's a high chance a 4 work creep would get spawned if you just happen to be slow refilling extensions or something, and then it wastes 1/5th of the energy it would have mined.

task.mining.js
```diff
@@ -55,7 +56,16 @@ mod.handleSpawningStarted = params => {
...
 mod.checkForRequiredCreeps = (flag) => {
-    let spawnRoom = Room.bestSpawnRoomFor(flag.pos.roomName);
+    const spawnRoomParams = {
+        targetRoom: flag.pos.roomName,
+        minRCL: mod.minControllerLevel,
+        allowTargetRoom: true,
+        minEnergyCapacity: 550,
+        rangeRclRatio: 2,
+        rangeQueueRatio: 51,
+        callBack: room => room.controller.level < 4 || room.energyAvailable > 1000
```
karlthepagan: Note this restriction on spawn room. Actually that should be 800 min energy

karlthepagan: Argh, well this doesn't get validated when the queue is popped. I'll remove the 4 work change

task.mining.js
whardeman: Everything else looks downright awesome, by the way.

whardeman:
@karlthepagan I will reintegrate first thing in the morning and test.

task.mining.js
```diff
@@ -55,7 +56,16 @@ mod.handleSpawningStarted = params => {
 };
 // check if a new creep has to be spawned
 mod.checkForRequiredCreeps = (flag) => {
-    let spawnRoom = Room.bestSpawnRoomFor(flag.pos.roomName);
+    const spawnRoomParams = {
+        targetRoom: flag.pos.roomName,
+        minRCL: mod.minControllerLevel,
+        allowTargetRoom: true,
+        minEnergyCapacity: 550,
```
karlthepagan: also changing minEnergyCapacity to 650

karlthepagan:
I see cyber advancing in a similar direction. I'll be rewriting a large part of this today.

Issue #517, opened 2017-01-16
Pioneer Improvement thread 


Dasku:
Ideas below, will add to this list:

1) Pioneers create the spawn on their own flag, regardless of if they are spawn pioneers or additional pioneers (from green/red flag). We could simply only build on green/white flag or:
       * Increase number of pioneers sent from spawn flag from 4-5/6? (so green/red flag no longer needed for new rooms) (I think 4 is enough but it seems most people send more with green/red)
       * Don't delete pioneer spawn flag when spawn completes, instead leave it until RCL4 and mining starts (reduce to 3 if/when mining requirements reduce to RCL3)
              * Reduce number of pioneers after spawn finished to 2 until RCL3-4 is reached (this is not always a good move if its a hard trip to the new room via SK rooms so deleting the flag after spawn completes is still possible)

2) Additional Pioneers (Green/Red) could then be redeveloped a bit and be used as a dead room revival mechanism #493 triggered from a task maybe?

Not sure what else pioneers *need* to do so this could be a quick change.

cyberblast:
I think the easiest and most flexible solution would be to ONLY use 1 flag for the pioneers (spawning 1 each). 
An additional manually set flag marks the position for the first spawn. 

Flag colors could stay the same...

torin555:
+1 for having green/red be used to spawn creeps for room revival

torin555:
One issue we need to address for room revival with pioneers is to never spawn the pioneer in the room with the flag. There's never a reason for that to be done.

torin555:
@cyberblast I think the auto flag placement is bugged, it's placing them in rooms that aren't dying.

cyberblast:
hmm reproducible?

torin555:
No clue why it's happening, but it's happened in two of my rooms

cyberblast:
strange. 

I forgot flushing the cached property. But I don't think it can be the reason. Anything special happened to both rooms when they got the flag?

torin555:
I'm not sure, wasn't watching the rooms at the time. I'll keep an eye out for it to happen more.. If it doesn't happen again for the next couple days I think we can just say it was a freak occurrence.

torin555:
Nope still happening, it's happened to two more of my rooms. Still no idea why, it seems totally random.

karlthepagan:
I have no idea how but I'm seeing this now. I'll triage when I can.

cyberblast:
[...]

PR #573, opened 2017-01-20
change pioneer flag system

cyberblast:
* Only FLAG_COLOR.claim.pioneer will spawn pioneers (1 each). 
* Use FLAG_COLOR.claim.spawn to mark the position for the first spawn (will not spawn additional pioneers). 

#517


OverlordQ:
I'm confused on the purpose of this because it breaks a lot of expected behavior

PR #575, opened 2017-01-20
Bug fix for bestSpawnRoomFor

pandabear41:
When claiming a new room the privateer and other tasks will select the newly claimed room to spawn creeps even though a spawn hasn't been built yet.


pandabear41:
Oh I was using old code. This should work since we no longer use bestSpawnRoomFor but findSpawnRoom.

PR #580, opened 2017-01-21
Recycler creep

whardeman:
If you REALLY want to make a creep recycle itself, this is an extremely easy way.

`Game.creeps[<creepName>].data.creepType="recycler";`


karlthepagan:
Recycler v2 ideas https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/581

whardeman:
Good suggestion 👍

cyberblast:
I don't think we need this. 
You can already do 
`Creep.action.recycling.assign(Game.creeps['<creepName>'])`

karlthepagan:
Assigning recycling fails to complete for several behaviours: upgrader and a few others.

@whardeman also needs to recycle a creep which was mistakenly spawned, so let's make sure that works.

Also see rev 2 ideas. This would be the template behaviour for hauling energy from room to room.

cyberblast:
> Assigning recycling fails to complete for several behaviours: upgrader and a few others.

why?

karlthepagan:
upgraders (and miners?) have a run method which does not check for a currently assigned action.

cyberblast:
Maybe we should fix that instead of developing a new behaviour.

whardeman:
@cyberblast I don't care if we keep it this way, but frankly, in the moment we are struggling through quite a lot of changes, mainly due to the new task systems. I have tried assigning the recycle action from console many, many times because it _is_ obvious. The problem is, it simply doesn't work many times, either due to a race condition, or the creep's own action priority list overrides it. 

If you want to fix it so that assigning the recycle action works 100% of the time, that would be great.

cyberblast:
I didn't know there are issues with assigning recycling, and I'm using it as well.

PR #583, opened 2017-01-21
hotfix cannot read property of undefined

webremote:
```TypeError: Cannot read property 'isValidSetup' of undefined
    at probe (spawn:18:25)
    at arraySome (/opt/engine/node_modules/lodash/index.js:1493:13)
    at Function.some (/opt/engine/node_modules/lodash/index.js:7051:14)
    at Spawn.execute (spawn:24:22)
    at run (spawn:109:33)
    at Function.<anonymous> (/opt/engine/node_modules/lodash/index.js:3346:13)```

this is a hotfix that makes sure this error won't happen

### DIFF
--- a/spawn.js
+++ b/spawn.js
@@ -15,8 +15,12 @@ mod.extend = function(){
         // old spawning system 
         let that = this;
         let probe = setup => {
-            return setup.isValidSetup(room) && that.createCreepBySetup(setup);
-        }
+            try {
+                return setup.isValidSetup(room) && that.createCreepBySetup(setup);
+            } catch(e) {
+                return false;
+            }
+        };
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cyberblast** [comment]
but why does it happen? What has been undefined?

File: spawn.js [hunk 1]
```diff
@@ -15,8 +15,12 @@ mod.extend = function(){
         // old spawning system 
         let that = this;
         let probe = setup => {
-            return setup.isValidSetup(room) && that.createCreepBySetup(setup);
-        }
+            try {
+                return setup.isValidSetup(room) && that.createCreepBySetup(setup);
+            } catch(e) {
+                return false;
+            }
```
karlthepagan: undefined checks will always be faster than a try/catch block

if you identify what the undef is please add a falseyness check

PR #598, opened 2017-01-22
Huge bugfix patch for task.robbing

pandabear41:
- situation invasion doesn't matter since enemy might be still alive when robbing.
- destiny.targetName is not a room name. 
- can't take energy from sources when room isn't owned by yourself.
- can't uncharge or withdraw since those action are for owned rooms.
- traveling when full, can't check isValidAction, isAddableAcion, assign when already assigned travel.
- in robbing action disabled dismantling ramparts since creeps only have one work part (would take forever to dismantle).
- Exploit next room call missing.
- Do not want to use exploit mod since it counts energy in sources, we don't care about harvesting sources.


pandabear41:
Isn't it implied since ramparts do not have a store, energy, or mineralAmount property?

cyberblast:
no. 
I didn't mean to exclude ramparts but structures ON ramparts. 
Because you can't rob those. 
Previously, if it got such a structure assigned it dismantled it until it could get to the structures energy. 
Now we need to exclude structures ON ramparts.

pandabear41:
I understand. ;)

pandabear41:
Done. 
Also found an issue which could be fixed in the future but the robber sits in the room forever since it will repair roads, containers, and other nearby structures.

PR #602, opened 2017-01-22
cpu log trace, always show error logs that use entityWhere

karlthepagan:
Memory.debugTrace.main='cpu' to see basic cpu stats


main.js
```diff
@@ -222,7 +227,7 @@ module.exports.loop = function () {
         };
     }
     if (Memory.debugTrace === undefined) {
-        Memory.debugTrace = {};
+        Memory.debugTrace = {error:true, no:{}};
```
karlthepagan: not having `error: true` in the default is a bug

PR #603, opened 2017-01-22
strategy API

karlthepagan:
Demo implementation: https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/commit/58ca146a026900894eb55732e59060a1a8cf2c5d

Documentation in comments here.

### DIFF
--- a/creep.Action.js
+++ b/creep.Action.js
@@ -97,5 +97,13 @@ let Action = function(actionName){
     // assignment postprocessing
     // needs implementation in derived action
     this.onAssignment = (creep, target) => {};
-}
+    // empty default strategy
+    this.defaultStrategy = {
+        name: `default-${actionName}`
+    };
+    // strategy accessor
+    this.selectStrategies = function() {
+        return [this.defaultStrategy];
+    };
+};
 module.exports = Action;
--- a/creep.js
+++ b/creep.js
@@ -1,3 +1,5 @@
+const strategy = load("strategy");
+
 let mod = {};
 module.exports = mod;
 mod.extend = function(){
@@ -34,7 +36,7 @@ mod.extend = function(){
         if(Array.isArray(partTypes))
             return (this.body.some((part) => ( partTypes.includes(part.type) && part.hits > 0 )));
         else return (this.body.some((part) => ( part.type == partTypes && part.hits > 0 )));
-    } 
+    }
     Creep.prototype.run = function(behaviour){
         if( !this.spawning ){
             if(!behaviour && this.data && this.data.creepType) {
@@ -95,6 +97,8 @@ mod.extend = function(){
                 if( SAY_ASSIGNMENT ) this.say(String.fromCharCode(10133), SAY_PUBLIC);
             }
         }
+
+        strategy.freeStrategy(this);
     };
     Creep.prototype.leaveBorder = function() {
         // if on border move away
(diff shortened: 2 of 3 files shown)


karlthepagan:
### Documentation

#### Fundamental structures
* agent - the component whose operations are parameterized with strategies, e.g. creep
* client - the components who provide parameterizations, e.g. action, behaviour, task
* staticCustomStrategy - a custom strategy which may not refer to the runtime state
* customStrategy - a custom strategy which will not be cached and may refer to runtime state

karlthepagan:
#### Instrumenting an Agent
`strategy.decorateAgent([agent|prototype], [...client definitions])`

This method injects methods onto the agent (or agent prototype). Each client definition describes how clients can be identified and retrieved:
* definition.default - retrieves the client id for the agent in its current state
* definition.selector - retrieves a client for the given id

The clients form a chain of scopes and the ids of the clients selected in each scope are combined to define the `strategy key` for strategy lookup. For example: `[uncharging, remoteHauler, remoteMining]` defines the key for the strategy with that given action, behaviour, and task.

#### Client Scopes

Scopes in the client definitions and API accessor methods are in order or priority (low to high). The high priority client scopes take precedence over lower priority client scopes.

The client ids may be explicitly specified on API access, but in the current use-case only the lowest priority scope is specified.

karlthepagan:
#### Retrieving Strategies
`let handlerFunction = agent.getStrategyHandler([...scope ids],handlerName, ...handler arguments);`

This retrieves a strategy-specific function. The context for the comparison or decision being carried out by this strategy is specified by handler arguments.

The returnVal of this method will be a function to evaluate one or more arguments provided. For example the following is the standard target scoring client pattern:
```
const targetPool = // ...
const targetScore = agent.getStrategyHandler([lowScope.name], 'targetScore', agent, context);
if (!targetScore) return;
const targets = _.chain(targetPool).map(targetScore).filter('score').sortBy('score').reverse().value();
return targets[0] && targets[0].target || null;
```

#### Implementing Strategies

Each client must provide a strategy selector, which will be used to separate interface and implementation of this API. Additionally this could lead to further strategy refinements in the future. The `selectStrategies` method returns an array of strategies which will be combined (low priority first) and cached.
```
client.selectStrategies = function(lowScopeId) {
    return [mod.strategies.defaultStrategy, mod.strategies[lowScopeId]];
};
```

It is recommended that implementers provide higher-order functions in order to build scoring and logical functions in the context of the game state. For example if the state important to a score function is a pair of underlying rooms, but the entities being scored are structures inside the second room the client code for this strategy handler would look like:
```
client.defaultStrategy.targetScore = function (room1, room2) {
  // setup calculations comparing 2 rooms
  return function(targetStructure) {
    // perform calculations evaluating the target structure, return that structure's score
  }
}
```

karlthepagan:
completed https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/wiki/Strategy-Extensions

karlthepagan:
Here's another example of elegantly solving a problem with strategies:

https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/commit/8fa31a8e07849e9d744775c1a112397cc12249ca

Remote workers were border thrashing for me after I finely tuned my hauler ratio. There was never energy left over for workers. The logical loop was solved by always allowing workers to uncharge no matter the amount available.

PR #604, opened 2017-01-22
make sure TOUGH parts are always at the beginning

ESchuderer:


ESchuderer:
creep

creep.Setup.js
```diff
@@ -163,6 +163,12 @@ let Setup = function(typeName){
                 parts.splice(index, 1);
                 parts.push(HEAL);
             }
+
+            if( parts.includes(TOUGH) ) {
```
karlthepagan: This will move a single `TOUGH` part. Isn't this better handled by allowing part sorting? (using the parts comparator)

ESchuderer: That's right, I should not try fixing things when tired. Will look into it later this day.

cyberblast:
`parts.sort(Creep.partsComparator);` (https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/604/files#diff-58c6d49d304798758faaa73cc7ae7dc7R158 ) will already ensure TOUGH parts are first. 
Maybe something else fucked up !? 
some bad overrides? tried delete Memory.modules?

ESchuderer:
Very strange indeed, I've only seen one creep have this behaviour. No overrides except parameters and custom tower code. Will see if it happens again.

PR #608, opened 2017-01-24
movement traces, default message for trace

karlthepagan:
traces I added looking at the movement cycle bug

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -99,15 +99,20 @@ mod.extend = function(){
     Creep.prototype.leaveBorder = function() {
         // if on border move away
         // for emergency case, Path not found
+        let dir = 0;
         if( this.pos.y == 0 ){
-            this.move(BOTTOM);
+            dir = BOTTOM;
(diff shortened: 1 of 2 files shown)


creep.js
```diff
@@ -120,8 +125,19 @@ mod.extend = function(){
     // enoughRange: range at wich the creep may calm down and interact with the target (may be higher than intentionRange)
     // range: current distance (optional)
     Creep.prototype.drive = function( targetPos, intentionRange, enoughRange, range ) {
-        if( !targetPos || this.fatigue > 0 || range <= intentionRange ) return;
+        if( !targetPos ) {
+            if( DEBUG && TRACE ) trace('Creep', {creepName:this.name, drive: 'no target', Creep:'drive'}, 'no target');
+            return;
+        }
         if( !range ) range = this.pos.getRangeTo(targetPos);
+        if( range <= intentionRange ) {
```
karlthepagan: comparing undefined range to intention range would always fail, probably not a common case

PR #613, opened 2017-01-27
Add Clear Spawn Queue Examples To commands.js

TafThorne:
When the code gets stuck it is helpful to clear the command queues.  These examples will clear each level of queue in a room. 

This patch was mentioned in issue #525.


TafThorne:
I would be quite happy for @webremote 's suggestion to go in as well.  What is the worst that would happen if it was run?  I expect things would recover their queues over time.  If it is not there someone might not notice a room with a stuck queue.

cyberblast:
Clearing the spawnQueues manually will have side effects. As some tasks may still have memory entries reflecting that a creep has been queued, waiting for a  Creep.spawningStarted event which will never occur.

karlthepagan:
If we use the Room.prototype wrappers then tasks will have a chance to reconcile these changes.

Issue #614, opened 2017-01-27
Must Always Be Able To Build A Worker With 200 Energy


TafThorne:
When a room collapses you can only safely assume that there is 0 energy in the extension and 0 energy in the spawn.  The game will give you a full spawn of energy after a while and from this life can resume.  If our code cannot build a worker with 200 energy at any RCL, there is a chance the room will _never_ automatically recover from a collapse.  It also makes it harder to spawn a replacement worker as you must used a command such as `Game.spawns['<spawnName>'].createCreepBySetup(Creep.setup.worker);` rather than `Game.rooms['<roomName>'].spawnQueueLow.push({parts:[MOVE,WORK,CARRY],name:'max',setup:'worker'});`. 

There is little chance of room collapse I hear you say?  Well it happened often enough in development when code cannot be tested to the Nth degree.  It is also quite possible for an invasion to severely dent a room but not destroy it out right.  If you have 0 haulers or workers left in a room it would be better to ensure that at least one worker could spawn than none.  

There is a manual work around.  Sure but I am not sat on a computer, watching the game 24x7.  If my room flakes out at 02:00 I would rather see it running when I wake up and check at 09:00.  That would also reduce the exposure I have to opportunistic neighbours spotting the struggling room and invading.  It is also a little confusing when you call `Game.spawns['S11W19'].createCreepBySetup(Creep.setup.worker);` and do not get one spawned.  

The current code base requires more than `minAbsEnergyAvailable: 200` for rooms with a Room Controller Level greater than 4.

cyberblast:
But we are already spawning a worker automatically when there is no hauler, worker or pioneer around using the queue. No minAbsEnergyAvailable required for that. 
The higher minAbsEnergyAvailable ensures worker have a larger min size when everything runs fine.

karlthepagan:
Forkmantis is looking into integration test services. We'll make an announcement when it's viable.

TafThorne:
I am not sure if this is related or if I should raise a seperate issue.   Since updating my code to the latest on dev yesterday I have had at least two rooms stall with a too but to build worker in the High priority queue.  

![image](https://cloud.githubusercontent.com/assets/586550/22395661/9046fe8e-e53c-11e6-8ad4-07ab17929bfe.png)

[...]

PR #615, opened 2017-01-27
Require 200 Energy To Spawn A Worker

TafThorne:
When a room collapses you can only safely assume that there is 0 energy in the extension and 0 energy in the spawn. The game will give you a full spawn of energy after a while and from this life can resume.  We should always be able to spawn a worker to drive room recovery with that 200 energy.  

This is a simple fix for issue #614 Must Always Be Able To Build A Worker With 200 Energy.  

It could be improved by making the minimum energy level required to spawn a worker dependent on the population of the room.  If there are other workers or haulers in the room, we could use the higher level.  This would ensure larger workers are always made at higher levels unless a population collapse has happened.  The advantage of this change is that it is simple, it works and it solves the issue this second.

### DIFF
--- a/creep.setup.worker.js
+++ b/creep.setup.worker.js
@@ -59,7 +59,7 @@ setup.RCL = {
     5: {
         fixedBody: [],
         multiBody: [CARRY, WORK, MOVE],
-        minAbsEnergyAvailable: 400,
+        minAbsEnergyAvailable: 200,
         minEnergyAvailable: room => setup.hasMinerOrHauler(room) ? 0.3 : 0,
         maxMulti: 8,
         maxCount: room => setup.maxWorker(room),
@@ -68,7 +68,7 @@ setup.RCL = {
     6: {
         fixedBody: [],
         multiBody: [CARRY, WORK, MOVE],
-        minAbsEnergyAvailable: 600,
+        minAbsEnergyAvailable: 200,
         minEnergyAvailable: room => setup.hasMinerOrHauler(room) ? 0.3 : 0,
         maxMulti: 8,
         maxCount: room => setup.maxWorker(room),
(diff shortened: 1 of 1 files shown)


cyberblast:
But we are already spawning a worker automatically when there is no hauler, worker or pioneer around using the queue. No minAbsEnergyAvailable required for that.
The higher minAbsEnergyAvailable ensures worker have a larger min size when everything runs fine.

TafThorne:
Well why was one _never_ spawning for me until I set the number to 200 then?  I had several hours of room replay to prove it never happened too.

chobobobo:
@TafThorne I think this was resolved by https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/666 if so, can this be closed?

TafThorne:
#666 is talking about 300 energy rather than 200 energy.  Does that not mean it is possible for creeps that are too large to spawn will still be left in the queue, blocking the 200 size worker?

chobobobo:
@TafThorne as I understood it, the issue was that you could get into situations where a room could not spawn a new worker.  Since a room will naturally regenerate up to 300 energy this should resolve that case by removing all creeps that are above that threshold.  If a room has "collapsed" meaning no haulers/workers/pioneers then task.pioneer.handleRoomDied should spawn a 200 energy worker to get things going again.

TafThorne:
I was sure spawns only held 200 energy... why was that?  No matter, 300 seems correct so I can see that removing all creeps with a value over 300 from the queues should sort things out.

chobobobo:
@TafThorne assuming this means it is resolved for you.

TafThorne:
I have not seen a collapse with the more recent code base.  Removed branch as requested.

PR #617, opened 2017-01-27
Feature/remote hauler capacity

karlthepagan:
This is a very basic capacity-based hauler spawning:

documentation linked from the tuning page:
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/wiki/Settings-&-Tuning#3-task-settings

### Quick Start
**viral.parameters.js**
This effectively unlocks the capacity feature.
```
    REMOTE_HAULER_MULTIPLIER: 4,
    REMOTE_HAULER_REHOME: true,
```

**commands**
This changes the carry capacity by the number of carry parts built to fill the route. Stored in task memory. -20 removes the full capacity for 1 source 1 room away, -40 to remove 1 source 2 rooms away. Positive numbers add more carry parts. Negative numbers remove carry parts. Repeat the command to continue changing the balance number.

This waits for the next spawning cycle. A higher hauler multiplier may be needed to achieve the number of needed parts.
```
Task.mining.carry('E48N81', -10)
```

### DIFF
--- a/task.mining.js
+++ b/task.mining.js
@@ -1,6 +1,7 @@
 let mod = {};
 module.exports = mod;
 mod.minControllerLevel = 4;
+mod.name = 'mining';
 mod.register = () => {
     // when a new flag has been found (occurs every tick, for each flag)
     Flag.found.on( flag => Task.mining.handleFlagFound(flag) );
@@ -8,11 +9,12 @@ mod.register = () => {
     Flag.FlagRemoved.on( flagName => Task.mining.handleFlagRemoved(flagName) );
     // a creep starts spawning
     Creep.spawningStarted.on( params => Task.mining.handleSpawningStarted(params) );
+    Creep.spawningCompleted.on( creep => Task.mining.handleSpawningCompleted(creep) );
 };
 mod.checkFlag = (flag) => {
     if( flag.color == FLAG_COLOR.claim.mining.color && flag.secondaryColor == FLAG_COLOR.claim.mining.secondaryColor ) {
         flag.memory.roomName = flag.pos.roomName;
-        flag.memory.task = 'mining';
+        flag.memory.task = mod.name;
         return true;
     }
     return false;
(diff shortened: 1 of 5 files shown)


task.mining.js
```diff
@@ -198,3 +229,66 @@ mod.creep = {
...
+};
+function haulerWeightToCarry(weight) {
+    if( !weight || weight < 0) return 0;
+    const multiWeight = _.max([0, weight - 500]);
+    return 5 + 2 * _.floor(multiWeight / 150);
+}
+function haulerCarryToWeight(carry) {
+    if( !carry || carry < 0) return 0;
+    const multiCarry = _.max([0, carry - 5]);
+    return 500 + 150 * _.ceil(multiCarry * 0.5);
```
karlthepagan: These numbers are based on the body definitions. I'll try to make them derived soon.

karlthepagan:
I need to strategize how to remove the WORK part from capacity haulers. Taking it off all but the first is a good step, but we will need better movement strategies so that roads don't die.

karlthepagan:
WARNING: you **will** hit your CPU limit if you ramp this feature up too quickly

karlthepagan:
I'm going to backport carry capacity into in-room hauler calculations. It's really useful.

karlthepagan:
Looking for more testers of multiple GCL and RCL!

See the commands in the description to get started.

chobobobo:
I'll have a look at this and give some feedback, the goal of this is to dynamically spawn haulers based on distance/need?

karlthepagan:
Yes, but not really dynamically. It is just based on how far and how many sources. It should be very simple and in my experience these numbers should handle the worst case (not counting invasions).

The miner replacement is somewhat dynamic. Their predicted renewal is based on literal arrival time.

chobobobo:
> commands
> This debits the carry capacity by the number of carry parts built to fill the route. Stored in task memory. -20 is the full capacity for 1 source 1 room away, -40 for 1 source 2 rooms away, etc.
> 
> Task.mining.carry('E48N81', -10)

I tried this, but all my remoteHaulers weren't replaced, I instead used positive numbers and they seem to have reappeared.  Typo?

karlthepagan:
Sorry, I updated my comment to make it more explicit what that command does. Haulers will be respawned as they are needed but not recycled in order to fit the new specification. There is a minimum size for each spawn, use the parameter `REMOTE_HAULER_MIN_WEIGHT: 500` to reduce the minimum from 800 to 500.

karlthepagan:
I think this is ready to come into dev.

PR #621, opened 2017-01-28
Set Reserver Priority To Low

TafThorne:
Reservers are cheaper to build than many pioneers or remote miners.  This can lead to remote rooms that are being exploited or mined getting filled with reservers who do not generate revenue, rather that creeps or do.  Get some cash first, then worry about the chance of getting more cash.  

Reservers are cheaper than some haulers and workers too.  If they are on the medium priority queue then the reserver might get generated first.  This could be very bad as a main room without any workers or haulers can quickly grind to a halt.  

This change makes the generation or reservers a low priority spawn, as already happens for claimers.  

This may go some way to addressing #614 Must Always Be Able To Build A Worker With 200 Energy; where it seems that sometimes a new worker is never generated.  It might help reduce the chances of a level 6 room collapse too as #525 documented for two rooms.  The exact causes of that are unknown though.

### DIFF
--- a/task.reserve.js
+++ b/task.reserve.js
@@ -6,7 +6,8 @@ mod.creep = {
         fixedBody: [CLAIM, CLAIM, MOVE, MOVE],
         multiBody: [],
         name: "reserver", 
-        behaviour: "claimer"
+        behaviour: "claimer",
+        queue: "Low"
     },
 };
 // hook into events


TafThorne:
I am going to try running with this code change for a while and see if it helps prevent: 

- Level 6 room collapse
- No replacement workers getting spawned

Dasku:
I think this may be a good change for say RCL1-3 (maybe 4 too), but once you have mining working and rooms set, you really need the medium reserver queue bump to make sure it stays reserved or you lose 1/.2 your available energy from that room.

TafThorne:
Looks like this did not fix the issue.  I have had another stuck room develop today 
image

Seems that a large worker destined to be a pioneer might be stuck in the pipeline.

TafThorne:
Have twice as much energy in the remote rooms is no help when you have 0
miners,haulers or privateers though and that is what is happening. This is
at level 5, 6 and 7.

TafThorne:
Also a room should _never_ build a claimer when it does not have at least 1 worker, 1 hauler and 1 miner in it.  I would also be happy with either 2 workers or 2 haulers and 2 miners.  Basically anything that means a room has a guaranteed energy supply when 1 creep dies and no replacment is forthcoming.  

Over in #614 I just documented a replay where the only creep in a room is a small worker.  Instead of spawning any miners, haulers or workers to rebuild the room the next thing out of the spawn was a claimer for a remote exploit.  Then the sole worker in the room expired and no further creeps were ever generated.

Dasku:
Seems now everything is queue based, this is becoming an issue for a few people,

pandabear41:
Default queue is low. 
This isn't the real problem. There is an issue with the auto room collapse.

TafThorne:
@pandabear41 It is the problem with exploiting remote rooms.  You get many more claimers/reservers than you do privateers.  That costs you a lot of energy for no return on investment.  If the privateers were getting spawned first you would get energy in which could then be spent on reservers to make it so you got more energy in.

TafThorne:
I agree that there is a seperate issue caused by some combination of factors that makes some rooms very prone to collapse.  There is also an issue where collapsed rooms will not auto recover either.  Ensuring claimers/reservers get spawned after the privateers (or miners and haulers) for a remote room may help both those issues or it may not.

karlthepagan:
I have started working on another way to rebalance this: only build reservers for remote mines when their carrying capacity is over 50%.

Another part of this change is that you will use multiple flags to override that decision: green/brown to mine and default reserving; green/grey to always try to reserve.

karlthepagan:
Is this resolved by #668?

TafThorne:
Honestly Karl it is very hard for me to even know whether I have pulled the patch into my running code, let alone test the fis at present.  You ask if an issue fixes something mentioned in this pull request.  I can see that issue had two commits that were merged 6 days ago but when you look at the changes themselves they say they only existed two days ago.   I probably pulled code in the last 6 days but not in the past 2.  So what do I have:
```
~/screeps/ocs-package/ocs.public$ git describe --tags
stable-2017-1-13-114-gdd15d65
```
So when did commit gdd15d65 happen?  Oh Git Hub does not look those up does it only the long version.  Fine dd15d65e6272494ca22b688fe1f3f81bea9bf3de happened on Jan 30th so I did not pull in the past two weeks.  

I will pull now and see what results I get over the course of the day.

TafThorne:
Oh crud.  It is the magic package things causing the lag.  Is it safe for me to manually pull the sub-modules instead?  Well we can see what happens.  `git pull` performed on both ocs.public and ocs.internal and then a `grunt deploy`.  

I can see there are a few reservers sat in exploited rooms without an mining right this second so we can see if things improve.  Hang on a moment.  Despite there being a bunch of new branch notes when I did the pull it seems I am still on the same revision as before.  

Well now I am lost.  Eyeballing the top level directory did not reveal anything obvious about a revision number being fixed.  Looking at the package repository history does not reveal some magic file change to move on the revisions either.  

I will be back after I learn more about sub-modules, grunt or GitHub.  Whitched pile of magic it that is keeping me on old code.

karlthepagan:
Let us know how reservers are behaving in the current dev branch @TafThorne .

TafThorne:
I have been running with a88e856b7cfdce308019114f6f0fb499b84ebd17 and I was working around the issues addressed by #809 I think (but cannot be certain) that the claimers were still getting spawned before any privateers (or is it pioneers I forget right now).  That may have changed in the past coupe of weeks since that changeset though.

karlthepagan:
Ok, I understand the use-case. I think that it is reasonable to restrict reservers to the low-priority queue if the privateers are the ones triggering this claim request.

karlthepagan:
If #1103 satisfies this need please close.

karlthepagan:
Implemented in #1103. Please reopen if this is not sufficient.

PR #625, opened 2017-01-28
Add Memory Stats for Grafana Tracking

Spedwards:


parameter.js
```diff
@@ -6,6 +6,7 @@ let mod = {
     SAY_PUBLIC: true, // creeps talk public
     DEBUG: true, // gimme some more details, use false not undefined to unset
     TRACE: false, // use Memory.debugTrace for low-level information
+    GRAFANA: true, // track for Grafana data
```
karlthepagan: :+1:

Perhaps default to false for a stable upgrade path.

main.js
```diff
@@ -130,7 +130,8 @@ global.install = () => {
         FlagDir: load("flagDir"),
         Task: load("task"),
         Tower: load("tower"),
-        Events: load('events')
+        Events: load('events'),
+        Grafana: load('grafana'),
```
karlthepagan: `GRAFANA && load(...`

karlthepagan:
Some of these are redundant with analyze, but that's a premature optimization.

Spedwards:
Added an interval to Grafana tracking. It was pointed out that as your GCL gets higher, this could be expensive to run.

pandabear41:
Do you really need an interval? Grafana should monitor on every tick and uses less than 0.1 CPU. 
Also for those players Grafana challenged, you should put a setup guide or a link to one in our wiki. I know that we will get some issues opened up for people confused about it or not able to get it working.

Spedwards:
@pandabear41 Someone pointed out that people with a lot of rooms will have more CPU used. And it also depends on what you're gathering.

Screepspl.us only fetches data once every 15 seconds, so having it run every tick is pointless. It should be no less than 3 ticks.

pandabear41:
Yeah true. 
I don't use screepspl, but my own Grafana setup so I like the many data points.
Plus I could always set it to 1 anyways.

Issue #428, opened 2017-01-06
Allow Miners and Haulers to Spawn at RCL 2


torin555:
For this recent respawn I tried changing the spawn requirements of miners and haulers to RCL 2 instead of 3 and it's working much better. It allows for a constant supply of energy for the workers instead of making them mine it themselves. 

It's also brought to light the fact that using miners and haulers at RCL 2, you can spawn way more workers and get to RCL 4 much faster, since all that energy would just go to waste otherwise.

cyberblast:
Sounds good :) 👍

whardeman:
I love this. I've actually forced this through myself at times when I felt relying on tiny workers was going far to slowly. I'm still trying to expand as rapidly as possible (about to hit GCL 5), so I want the most fastest and most efficient RCL increase I can get as I move into more crowded territory.

Is there any active work being done for this change? What are the factors to be considered for making the change fit any room setup? Here's what I came up with...

- For haulers: source proximity (closer implies fewer required haulers) seems to be a key factor in spawning most efficient number. This includes
  - sources to Spawn
  - sources to Controller, and
  - sources to one another 
- For miners, always spawn 1/source, but respawn timing for soon-to-expire miners needs to take into account source proximity to Spawn.

cyberblast:
> respawn timing for soon-to-expire miners needs to take into account source proximity to Spawn.

Yes there has been a recent change to improve that calculation. 
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/blob/dev/creep.behaviour.miner.js#L75

chobobobo:
I'm a bit new to forks and pull requests, but I've forked the repo and this is the first change I will start working on.  Getting miners & haulers to start working is simple, I also plan to allow workers to continue mining a source as long as the miner has < 5 work parts so that we fully utilize the source [#545](https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/545) should provide us with more dynamic worker populations.

karlthepagan:
I have a work in progress idea for this that you're welcome to take over @chobobobo !

https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/tree/feature/lowMiner

chobobobo:
@karlthepagan ok, sounds good, so I should switch to that branch in my own fork, finish it and then submit a pull request right?

karlthepagan:
sounds good! you can PR from your own repo until you get promoted into the org.

[...]

PR #629, opened 2017-01-29
Feature/low miner

chobobobo:
Code to enable miners, haulers and upgraders at low RCL to speed up initial founding.


chobobobo:
For some reason, when I rebased on my own fork, it seems to have committed twice?  Not sure what I did wrong there.

This satisfies https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/428

Issue #565, opened 2017-01-19
saveContainer bug


pandabear41:
When a container is near a source and a controller it will be assigned to both. Creeps will get stuck in a loop of filling and emptying the container. Since controller containers need to be filled and source containers need to be emptied. This happens in one of my brother's rooms since the controller and source are close to each other. 

http://screeps.com/s/JDHtTe

cyberblast:
That's strange. Those containers got some special care and it worked already. Did he change something? 
I don't have such a container anymore currently. Maybe something broke...

pandabear41:
It is because the the controller, source, mineral flags on the memory of a container can be enabled all at once. So in my case
container.controller = true and container.source = true

cyberblast:
Yes. It's called "managed" container: https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/blob/dev/room.js#L88 and gets special handling (may never contain less than MANAGED_CONTAINER_TRIGGER and not more than 1-MANAGED_CONTAINER_TRIGGER (relative)).

cyberblast:
does anybody else has a "managed" container (in range of source & controller)?

ESchuderer:
I have the same problem in one of my rooms, but being lazy just decreased the range where controllers look for containers to 2 fields.

karlthepagan:
I have had to avoid this in 2 of my rooms lately.

cyberblast:
this worked like charm for a long time. we should find out what the reason may be. 
Unfortunately I don't have any managed container anymore. Maybe someone else can debug it...

whardeman:
Is this resolved by #633? Can it be closed?

PR #633, opened 2017-01-29
fix managed container charging/uncharging

cyberblast:
#565


cyberblast:
hmm, no, there are some other sings contained :/ going to clean that up

PR #634, opened 2017-01-29
implementing unpave to remove a drawing

cyberblast:


global.js
```diff
@@ -303,8 +303,17 @@ mod.pave = function(roomName){
     let remove = flag => flag.remove();
     flags.forEach(remove);
 };
-// TODO: remove a pavement
-mod.unpave = function(roomname){};
+mod.unpave = function(roomname){
+    if( !Memory.pavementArt || !Memory.pavementArt[roomname] ) return false;
+    let room = Game.rooms[roomname];
+    if( !room ) return false;
+    let unpaved = structure => Memory.pavementArt[roomname].indexOf('x'+structure.pos.x+'y'+structure.pos.y+'x') >= 0;
```
karlthepagan: Here's a regex to prevent missing the final coord: `/'x'+x+'y'+y+'(?!\d)'/`

cyberblast: yes, but calling `pave` will always add a trailing x. 
`Memory.pavementArt[roomName] = val + flags.map(posMap).join('')+'x';`

karlthepagan: kk, simple!

cyberblast: But thank you :) 
we can still change it somewhen..

PR #635, opened 2017-01-29
fix managed container charging/uncharging

cyberblast:


karlthepagan:
Running this code now, looks fine. I'll rebase my strategy work onto it.

PR #641, opened 2017-01-31
remote haulers uncharge first

karlthepagan:
remote haulers withdrawing from containers and then picking produces a more stable steady-state than the 80% fill & pick first method


pandabear41:
I have thought about this for a while and currently use some other way to haul. 
How about instead something that picks first and then uncharges from nearest container if one is within 3 range?

karlthepagan:
I've started to sketch out a new action for "grabbing" loose energy while traveling. That will be a different goal.

Issue #626, opened 2017-01-28
Travelling action should end on rooms bordering the target room.


karlthepagan:
Difficulty: intermediate to advanced.

## The problem

The majority of pathfinding logical loops are caused by a bad approach to the target room.

### Case 1 - room geometry blocked by walls

The creep enters the target room, then finds their path is blocked by constructed walls.

### Case 2 - room geometry is less efficient than a different entrance

The creeps enters the target room, then finds their path is covered by swamps and the creep navigates outside the room in order to utilize roads.

### Case 4 (out of scope for this proposal) - determinatedSpot or determinatedTarget is occupied

When the spots are occupied creeps cecide to return to home room or path out to look for a new target. This should be resolved elsewhere.

## The Fix.

The simple fix may be to switch to moveTo with API path caching for pathfinding, but this will be less efficient in high traffic situation. Perhaps a special case when adjacent to the target room can use moveTo and a special movement mode (currently "auto" and "evade").

karlthepagan:
Based on movement I observed in E43N81 I believe that the random? movement that happens around action.approach may lead to pathing out of the room.

Shortly after this tick: https://screeps.com/a/#!/history/E43N81?t=17129237

karlthepagan:
Here is one observed loop:

```
{"creepName":"reserver-gammaRMine-2","pos":{"x":49,"y":17,"roomName":"E47N81"},"leaveBorder":7,"nopath":"zero len","drive":"nopath","Creep":"drive"}
{"creepName":"reserver-gammaRMine-2","run":"reserving","claimer":"run","Behaviour":"claimer"}

{"creepName":"reserver-gammaRMine-2","pos":{"x":48,"y":17,"roomName":"E47N81"},"Behaviour":"claimer","Creep":"run"}
{"creepName":"reserver-gammaRMine-2","assign":"reserving","target":"gammaRMine","Action":"assign"}
{"creepName":"reserver-gammaRMine-2","nextAction":"reserving","reserve":"nextAction","Task":"reserve"}
{"creepName":"reserver-gammaRMine-2","pos":{"x":48,"y":17,"roomName":"E47N81"},"direction":3,"moveResult":0,"drive":"ok","Creep":"drive"}
{"creepName":"reserver-gammaRMine-2","run":"reserving","claimer":"run","Behaviour":"claimer"}

[...]

PR #647, opened 2017-02-01
Bug/precalc no route

karlthepagan:
I think this helps the NO PATH bugs for route precalc.

if this shows data, here's the test case: https://screeps.com/a/#!/history/E44N80?t=17126608

my creeps are avoiding E44N81

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -215,16 +215,20 @@ mod.extend = function(){
         }
     };
     Creep.prototype.getPath = function( targetPos, ignoreCreeps ) {
+        let range = 1;
         let tempTarget = targetPos;
         if (ROUTE_PRECALCULATION && this.pos.roomName != targetPos.roomName) {
             var route = this.room.findRoute(targetPos.roomName);
-            if ( route.length > 0 )
-                targetPos = new RoomPosition(25,25,route[0].room);
+            if ( route.length > 0 ) {
+                targetPos = new RoomPosition(25, 25, route[0].room);
+                range = 24;
+            }
         }
 
         let path = this.room.findPath(this.pos, targetPos, {
             serialize: true,
-            ignoreCreeps: ignoreCreeps
+            ignoreCreeps: ignoreCreeps,
+            range,
         });
         if( path && path.length > 4 )
             return path.substr(4);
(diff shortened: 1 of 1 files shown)


karlthepagan:
might need some special logic to prevent pathing thru banned rooms when in or adjacent to the destination room

relates to #626

karlthepagan:
This has been working pretty well for the last few hours.

More testers wanted!

YourHuckleberry:
Added to my running code, will advise.

karlthepagan:
Example parameters:
```
    ROUTE_PRECALCULATION: true,
    ROUTE_ROOM_COST: {
        E44N81: 31,
    },
```

both of these are required for it to have function.

use this to actually avoid SK rooms! (in most cases)

karlthepagan:
I've fixed a pretty serious bug here. As was my suspicion in #626 any place in the code where we terminate movement in an adjacent room border it will result in movement cycling.

karlthepagan:
One more tweak, yes there were still edge cases in precalc route.

The point is to move thru the adjacent room along the route... so that should work now.

karlthepagan:
@chobobobo you're writing features with this so... review some :)

chobobobo:
@karlthepagan can you clearly define for me what the bug was, and how this resolves it?  I have learned enough to know that setting range to 24 means it's just finding a path until it's one square inside the room, and to know that seemingly simple changes to pathfinding can cause incredibly frustrating side effects!

karlthepagan:
using route precalculation creeps try to move to 1 square into the most adjacent room along the route. This will result in logical movement cycles in the case that the creep arrives via a blocked exit.

For example https://screeps.com/a/#!/room/W74S51 entrance from W74S50 on the right. My creeps would always path into the right entrance, see the block and turn around.

karlthepagan:
No it doesn't go to the adjacent room. It airways tries to go one room ahead of the adjacent room. :)

karlthepagan:
I'm going to keep this on hold until we can evaluate #751

karlthepagan:
Closing this in favor of traveler. Leaving branch for now.

PR #650, opened 2017-02-01
privateer can travel to room without vision

karlthepagan:
this is extremely useful when settling a new area (portals etc)

### DIFF
--- a/creep.action.travelling.js
+++ b/creep.action.travelling.js
@@ -8,9 +8,13 @@ action.step = function(creep){
     if(CHATTY) creep.say(this.name, SAY_PUBLIC);
     if( creep.target ){
         let pos;
-        if( creep.target.id == creep.id ) pos = new RoomPosition(25, 25, creep.data.travelRoom);
+        let targetRange = this.targetRange;
+        if( creep.target.id == creep.id ) {
+            targetRange = 24;
+            pos = new RoomPosition(25, 25, creep.data.travelRoom);
+        }
         else pos = creep.target.pos;
-        creep.drive( pos, this.reachedRange, this.targetRange, Infinity );
+        creep.drive( pos, this.reachedRange, targetRange, Infinity );
     }
     if( !creep.target || creep.target.pos.roomName == creep.pos.roomName ){
         // unregister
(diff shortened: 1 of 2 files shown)


karlthepagan:
I just realized this was for privateer not pioneer.

Doesn't change the feature ;) this is what I intended.

creep.behaviour.privateer.js
```diff
@@ -148,6 +148,7 @@ mod.exploitNextRoom = function(creep){
     // no new flag
     // go home
     Population.registerCreepFlag(creep, null);
-    Creep.action.travelling.assign(creep, Game.rooms[creep.data.homeRoom].controller);
+    creep.data.travelRoom = creep.data.homeRoom;
```
chobobobo: By manually assigning this, are we externally hacking at the internals of the travelling action?  Ultimately, shouldn't creep.action.travelling.assign be able to take travelRoom as a parameter?  If this is a workaround, please add a comment.

creep.action.travelling.js
```diff
@@ -8,9 +8,13 @@ action.step = function(creep){
     if(CHATTY) creep.say(this.name, SAY_PUBLIC);
     if( creep.target ){
         let pos;
-        if( creep.target.id == creep.id ) pos = new RoomPosition(25, 25, creep.data.travelRoom);
+        let targetRange = this.targetRange;
+        if( creep.target.id == creep.id ) {
+            targetRange = 24;
```
chobobobo: This has the effect of making the creep regenerate its route once it gets inside the target room?

creep.behaviour.privateer.js
karlthepagan: from what I have found this seems to be the standard mechanism to go to a room name

a target needs to be a real game object, and a room without vision is never a real game object

karlthepagan: @chobobobo see examples in `task.defense.js`

creep.action.travelling.js
karlthepagan: yup, also the only change i made was to let room to room travel have a 24 range

whardeman: I was just wondering about setting targetRange slightly lower to ensure the creep moves away from the border before re-routing, to help prevent the NO PATH thrashing, but maybe that's been resolved elsewhere.

karlthepagan: @whardeman resolved in the general case by #661

Issue #593, opened 2017-01-21
Dying creep workflow


cyberblast:
dave noticed in #590:
"I did notice the invader destroyed a reserver - there was log message Corrupt creep without population entry!!; reserver-Flagx-1. This would be expected since it was destroyed."

This shouldn't be the case. I wonder what is going wrong there. Could be the same root cause for some intermittent memory bugs...

chobobobo:
Perhaps related to this, I have been having remoteMiners that spawn seemingly in the same tick as the creep they are replacing, with the same name.
[5:37:09 PM]W7N7 > Good night remoteMiner-W7N6-mine-1!
[5:37:09 PM]W7N7 > Good morning remoteMiner-W7N6-mine-1!
[5:37:20 PM]pioneer-Flag2-2 > Farewell!
[5:37:42 PM]Corrupt creep without population entry!! : remoteMiner-W7N6-mine-1
[5:37:42 PM]W7N7 > Good morning remoteMiner-W7N6-mine-2!
[5:37:44 PM]W7N7 > Off to work remoteMiner-W7N6-mine-1!
[5:37:44 PM]TypeError: Cannot read property 'execute' of undefined
    at probeSpawn (population:252:61)
    at Array.forEach (native)
    at Object.mod.execute (population:253:28)
    at Object.module.exports.loop (main:265:16)
    at __mainLoop:1:52
    at ContextifyScript.Script.runInContext (vm.js:35:29)

karlthepagan:
Great correlation there.

chobobobo:
Today I accidentally spawned about 30 remoteMiners in rapid succession, this resulted in a few of them having this exact error.

chobobobo:
I stumbled across what may be part of this issue and added some code here: https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/tree/fix/creepLifecycle
It still needs code to migrate over gently, rather than spamming out birth/death notifications when you update versions.

PR #655, opened 2017-02-01
Change/improve remote mining

chobobobo:
Various changes to improve remote mining, the only one I think needs consideration is moving remoteMiner to the medium queue.  I've been having issues where remoteMiners are stuck way down the queue with haulers for several other rooms (that have miners) spawned before them, leaving a lot of haulers wandering around helpless. @karlthepagan if you think it looks good, then it's ready to merge in.


creep.behaviour.remoteHauler.js
```diff
@@ -37,14 +37,16 @@ mod.nextAction = function(creep){
             // no deposit :/ 
             // try spawn & extensions
             if( this.assign(creep, Creep.action.feeding) ) return;
-            // TODO: hauler shouldn't work. drop at spawn instead of calling worker behaviour
-            Creep.behaviour.worker.nextAction(creep);
+            // drop off at the nearest spawn
+            let spawn = creep.pos.findClosestByRange(creep.room.structures.spawns);
```
karlthepagan: how about spawn or spawn construction site?

it would be best if this was a specific spot next to the spawn or CS and combine all piles

creep.behaviour.remoteWorker.js
```diff
@@ -25,6 +25,7 @@ mod.nextAction = function(creep){
                 Creep.action.picking,
                 Creep.action.uncharging,
                 Creep.action.withdrawing,
+                Creep.action.harvesting,
```
karlthepagan: does this harvesting block a remoteMiner?

task.mining.js
```diff
@@ -82,7 +82,7 @@ mod.checkForRequiredCreeps = (flag) => {
 
     // TODO: don't iterate/filter all creeps (3 times) each tick. store numbers into memory (see guard tasks)
     const creepsByType = _.chain(Game.creeps)
-        .filter(function(c) {return c.data.destiny && c.data.destiny.room===roomName;})
+        .filter(function(c) {return c.data && c.data.destiny && c.data.destiny.room===roomName;})
```
karlthepagan: 👍

task.mining.js
```diff
@@ -218,7 +218,7 @@ mod.creep = {
         fixedBody: [MOVE, MOVE, MOVE, WORK, WORK, WORK, WORK, WORK, WORK, CARRY],
         multiBody: [],
         behaviour: 'remoteMiner',
-        queue: 'Low'
+        queue: 'Medium' // not much point in hauling or working without a miner, and they're a cheap spawn.
```
karlthepagan: this is a significant change, but I support it

very often reservers get spit out right away, sometimes there is a priority inversion problem and this is an ok stopgap

creep.action.dropping.js
```diff
@@ -0,0 +1,16 @@
+let action = new Creep.Action('dropping');
+module.exports = action;
+action.targetRange = 1;
```
karlthepagan: targetRange = 0, enoughRange = 1 plus changes below

task.mining.js
torin555: I'd like to voice concerns about this change, (feel free to dismiss this review @karlthepagan) But this doesn't just affect the setup of the remote mine. It would also be affecting all future miner spawning which could be a problem if other things are more important.

karlthepagan:
@torin555 the reasons I have for supporting this are:
1) *everything* is in the low queue, miners can get scheduled out by requested reservers, haulers, workers, military and this will lead to colony collapse
2) because of predictedRenewal miners are the most sensitive for spawn timing

karlthepagan:
@chobobobo have a look at my changes

creep.action.dropping.js
karlthepagan: I've implemented this

creep.behaviour.remoteWorker.js
chobobobo: Yes, I think it could potentially in the situation of one accesspoint until the worker finishes harvesting.  Is there a way for one of our creeps to tell another to move?

chobobobo:
@karlthepagan these look great, one additional change I would like to add is for remoteHaulers to collect dropped energy along their route, I don't want them wandering off their path, but they need to clean up after dead haulers.

creep.behaviour.remoteWorker.js
karlthepagan: It seems to work for worker/miner. Perhaps something to do with targetOf[0].determinatedSpot

karlthepagan:
That would make a good new action, but the easy fix is a little more complex than I'd like (and possibly CPU hungry).

In the mean time what will happen is when containers are empty and the hauler is < 80% capacity it will go look for drops in the destination room.

creep.behaviour.remoteWorker.js
karlthepagan: I take it back... this frequently doesn't work. If workers are idle they will block in-room miners forever :(

chobobobo:
@karlthepagan I have switched out homeRoom to now use bestSpawnRoomFor allowing remoteHaulers to reHome to the nearest room, this allows for remote mining to support lower RCL rooms as we no longer need storage to remoteMine.  Let me know what you think about that, as I may have overlooked something.

whardeman:
@chobobobo I'm also working on a branch that allows the user to explicitly define the room to which a remote mining task's energy should go. The current iteration would send _all_ energy mined in a room to the designated room, but there are ways to split 2-source rooms or possibly even round-robin several rooms from a single source room.

It requires manual designation by the player to have any effect, so anyone who doesn't want this can just ignore it. https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/tree/feature/explicit-remoteHauler-rehome

PR coming soon.

chobobobo:
@whardeman that sounds great, I'll take a look.

chobobobo:
@karlthepagan should be working now, I believe this is ready to be merged now.

karlthepagan:
I'm thinking of adding a population percent to the carry method. I've started using percentage of carry present as a heuristic for reservers.

karlthepagan:
@chobobobo we need to look at hauler spawning

first check your queue depth `_.values(Game.spawns).filter(i=>i.room.spawnQueueLow.length).map(i=>[i.name,'=>',i.room.spawnQueueLow.map(j=>j.name)])
`

then try running `_.values(Memory.tasks.mining).map(i=>delete i.haulersChecked)`

If this suddenly spikes the number of haulers in your queues then there's a problem. The haulersChecked memory should not affect the actual produced count, it's just an optimization right?

karlthepagan:
@chobobobo is the purpose of the spawning memory to recover from failed spawning? Creeps which are actually spawning will show up in Game.creeps

chobobobo:
@karlthepagan I will try those out.  haulersChecked is meant to solve the case where we want to spawn a hauler to meet the desired capacity, but the size returned is below the minimum threshold, this results in us being slightly under-capacity at times. The alternative would be to queue an additional hauler at the minimum threshold.

spawning memory was copied from task.guard and is meant to make sure we count the capacity of haulers being spawned in our calculations, if we ignore those then during the time haulers are being spawned additional ones get queued up.  Already had issues with consistency there however, so perhaps a better solution would be to simply scan all spawns (relatively low cost).

further to that, I believe that some refactoring could happen with the queued/spawning/running lists to generalize for tasks as there is a lot of overlap amongst tasks, but that should probably happen separately to this PR.

karlthepagan:
As discussed don't bother invalidating when creeps die, but increment when a spawn is queued.

chobobobo:
I now increment when a spawn actually happens, should resolve the hauler issue.

I also discovered and resolved an issue when remotehaulers rehome to an RCL2 room with no spawn yet.  They now drop at the spawn flag if they cannot find a pile.

I modified the memory.spawning lists for Task.mining to set timeouts and invalidate entries if (and only if) they have become stale.  If the spawn occurs as expected it will either reset the timeout for the next spawn to occur or remove the timer.  I think the reason this has been happening is because of #593 which is resulting in creeps spawned with no memories.

chobobobo:
Looks good after running overnight.

karlthepagan:
I've merged #678 into this.

karlthepagan:
In production I'm still running into capacity problems. This is impacting #668 because haulers are not spawning up to the level of their predicted maxWeight.

```
[8:22:02 PM] E46N82 > Good morning reserver-alphaTMine-1!
_.chain(Game.spawns).values().map(i=>i.room).unique().filter(i=>i.spawnQueueLow.length).map(i=>[`====${i.name}====>`,i.spawnQueueLow.map(j=>j.name)]).value();
====E46N82====>,reserver-alphaTRMine
 _.values(Memory.tasks.mining).map(i=>delete i.haulersChecked)
 true,true,true,true,true,true,true,true,true,true,true,true,true,true
[8:22:24 PM] E43N82 > Good night remoteHauler-deltaBMine-1!
 _.chain(Game.spawns).values().map(i=>i.room).unique().filter(i=>i.spawnQueueLow.length).map(i=>[`====${i.name}====>`,i.spawnQueueLow.map(j=>j.name)]).value();
 ====E47N81====>,remoteHauler-gammaRMine,====E46N82====>,reserver-alphaTRMine,remoteHauler-alphaLMine,remoteHauler-alphaTMine,remoteHauler-alphaRMine,remoteHauler-alphaTRMine,remoteHauler-betaOldMine,====E43N82====>,remoteHauler-deltaLMine,remoteHauler-deltaRMine
[8:22:40 PM]
```

chobobobo:
I'll look into that further tonight, and write some code that tracks the efficiency of remote rooms so that we can more readily catch issues like this in the future, there is an issue with #671 as well that may be affecting haulers where they go back to collect more energy and top-up after they have idled.  This is possibly the fault of remoteHaulers not using actions directly but I suggested a fix for it there.

whardeman:
Will merge in latest and test soon.

karlthepagan:
I've merged #671 into this branch.

If that merge is incorrect or needs to be removed please don't revert the merge... force update this branch to `94319e262e53ac28f8c37ed262624123ae31d901` and rebase your changes.

chobobobo:
Merge looks fine, I think that remoteHaulers should have some sort of limit on how many can claim a withdraw/pickup based on their capacity, currently in edge-cases they all see-saw between sources picking up small piles when they cannot reach capacity, I would prefer they each pick a source and sit there until they are full enough.

karlthepagan:
Much improved!

I only have one room out of sorts. I'll take a look when I can, but here's the data:

```
JSON.stringify(Memory.tasks.mining.E45N82)
[4:08:06 PM] E46N82 > Good night tractor-14!
 {"queued":{"remoteMiner":[],"remoteHauler":[{"room":"E46N82","name":"remoteHauler-alphaLMine","body":{"work":1,"carry":29,"move":15}}],"remoteWorker":[]},"carryParts":-8,"spawning":{"remoteMiner":[],"remoteHauler":[],"remoteWorker":[]},"running":{"remoteMiner":["remoteMiner-alphaLMine-1","remoteMiner-alphaLMine-2"],"remoteHauler":["remoteHauler-alphaLMine-1"],"remoteWorker":[]},"nextSpawnCheck":{"remoteHauler":0,"remoteMiner":0},"haulersChecked":1}
[4:08:10 PM] E47N81 > Off to work tractor-11!
 Task.mining.carry('E45N82')
 Task.mining overall hauler carry parts for E45N82 are decreased by 8. Currently 0% live
```

0% live is strange

karlthepagan:
>  I think that remoteHaulers should have some sort of limit on how many can claim a withdraw/pickup based on their capacity

This is something that I'm working towards in the strategy handlers. Let's move this PR toward release and open a new one for that improvement.

karlthepagan:
While this may or may not be ready for merging into dev it is becoming a very broad PR and will impact other PR's as long as it isn't merged.

whardeman:
To clarify my comment about better docs: 
- Provide command examples in the wiki
- In the command output, knowing what _"X% live"_ actually means would be helpful. I can intuit the meaning, but it should be explicit in the docs.

karlthepagan:
I think this is what you're looking for @whardeman 

https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/wiki/Task:-Mining#memory-commands

I'll find better language for the capacity percent.

whardeman:
@karlthepagan Yes, I think the docs there are good, but could really use more examples so people won't be afraid to use the commands. I'm adding some now under Task.mining.storage.

Also, I'm cleaning up the output messages a bit, to make them more consistent.

chobobobo:
Ok, added a diagnostic method Task.mining.checkCapacity(roomName - optional) which can give you some information about how your remote rooms are performing and identify any rooms that might need some TLC.  It's focused on haul capacity and energy left on the ground.

Aside from bugfixes, I will not be mucking with the feature-set of this branch anymore, I too would like to see it merged in soon.  @whardeman I recently updated some of the output messages, feel free to do whatever you need to to make them more readable/useful.

whardeman:
@chobobobo Those changes look great. And I love the checkCapacity feature. I was actually just looking at Task.mining.storage to do a couple things:
1. If the room passed isn't actually a room name, preferrably we would not store at that key (big waste of memory over time if typos happen).
2. If no room is supplied (eg, `Task.mining.storage()`) just return an overview of some kind.

Like you, I think all of that should wait until this PR is merged into dev.

whardeman:
While we're talking about features to be added in a later round:

- Every worker, and probably all haulers should have drive-by picking (similar to drive-by repair). If there's energy on the ground `distance < 2` and I have space available, I might as well pick it up, right?

karlthepagan:
I've also noticed that remote haulers robbing has complicated the flow of energy back to the room. This is because robbing does not have the minimum increment requirement that uncharging does (minimum 250).

I see a permanent solution in the cost/benefit system so I'm going to comment-out robbing from the hauler actions. I think that will make this behave as expected and feel more stable and ready for release.

karlthepagan:
looks like removing robbing is causing an issue, triaging... it might go back for release because of the capacity checks in remote hauler

PR #661, opened 2017-02-03
let travel along border rooms be to a specific position

karlthepagan:
reviews wanted, major change

The core philosophy here as described in #626 is that travel to reach targeted actions should result in an approach along the correct entrance to the target room. Problems actually navigating across that boundary are still being investigated.


karlthepagan:
This will spit errors for a few actions which don't have a pre-targeted destination.

karlthepagan:
This is now based on #650

chobobobo:
@karlthepagan what's the easiest way to test this feature?

karlthepagan:
@chobobobo set up conditions that produce a logical travel loop, easiest is blocking part of a split exit room like the right exit to https://screeps.com/a/#!/room/E74S7

creep.action.travelling.js
```diff
@@ -9,9 +9,24 @@ action.step = function(creep){
...
-        if( creep.target.id == creep.id ) {
-            targetRange = 24;
-            pos = new RoomPosition(25, 25, creep.data.travelRoom);
+        if( creep.target.id == creep.id) {
+            if( creep.data.travelPos ) {
+                pos = new RoomPosition(creep.data.travelPos.x, creep.data.travelPos.y, creep.data.travelPos.roomName);
+            } else if( !creep.data.travelRoom ) {
+                logError('no travel room', {creepName:creep.name, roomName:creep.room.name, Behaviour:creep.data.creepType, Action:'travelling'});
+            } else if( creep.room.getBorder(creep.data.travelRoom) ) {
+                logError('bad border target', {creepName:creep.name, roomName:creep.room.name, Behaviour:creep.data.creepType, Action:'travelling'});
```
chobobobo: This error message does not tell me what is expected, or why it is bad.

room.js
```diff
@@ -835,6 +835,12 @@ mod.extend = function(){
         },
     });
 
+    Room.prototype.getBorder = function(roomName) {
+        return _.findKey(Game.map.describeExits(this.name), function(name) {
+            return this.name === name;
```
chobobobo: Shouldn't this be comparing to roomName?  You have the names of all exits, and you want to find the exit that corresponds to the roomName you are looking for?

creep.action.travelling.js
```diff
@@ -9,9 +9,24 @@ action.step = function(creep){
...
+                targetRange = TRAVELLING_BORDER_RANGE;
+                pos = new RoomPosition(25, 25, creep.data.travelRoom);
+            } else {
+                targetRange = TRAVELLING_BORDER_RANGE;
+                pos = new RoomPosition(25, 25, creep.data.travelRoom);
+            }
+            if( creep.pos.roomName === pos.roomName ) {
```
chobobobo: This means that the creep will always only move one time into the room, so does targetRange have much meaning?

logxen:
My solution to traveling along room borders was to set all x and y == 0 and 49 positions to having path costs at least equal to swamp cost. You could even double it. They will still always find the exit when they need to.

creep.action.travelling.js
karlthepagan: yes, 25,25 is often un-movable so it's wasted CPU to search for that location

the signal to pathfinder is "how do i get to the next room" rather than "where exactly am i going" - this works very well in cases where the exact room location isn't known (when there is no vision)

room.js
karlthepagan: the exit name is the direction number as a string: 1-8

karlthepagan:
@logxen I like that for when we can set custom costmatrixes

creep.action.travelling.js
karlthepagan: good idea

PR #662, opened 2017-02-04
Task.mining.storage("flagRoom", "storageRoom");

whardeman:
Easy to use and only in effect with manual intervention.


karlthepagan:
documented!

PR #663, opened 2017-02-04
Fix when no creeps remain and large cost creep is in queue.

chobobobo:
If a room gets wiped, and a large cost worker is left in the queue, it gets stuck.  No point waiting for enough resources when they aren’t coming!


chobobobo:
@Spedwards quite right, fixed now.

PR #665, opened 2017-02-04
Feature/avoid room

karlthepagan:
this is a little more experimental. It's based on #647 but restores the room avoidance functionality

This is probably as good as I can do until I work out the custom costmatrix (which is a much bigger difference).

Test instructions:

set these parameters:
```
    ROUTE_PRECALCULATION: true,
    ROUTE_ROOM_COST: {
        E44N81: 31, // this room is 3x as expensive as an unowned room
        E45N82: 1, // i treat this room as if I own it
        E46N81: 11, // i treat this room as if I never own it
   }
```

### DIFF
--- a/room.js
+++ b/room.js
@@ -1414,6 +1414,9 @@ mod.getCostMatrix = function(roomName) {
     }
     // unset values / undefined reads from map
 };
+mod.validField = function(roomPos, checkWalkable = false, where = null) {
+    return mod.validFields(roomPos.roomName, roomPos.x, roomPos.x, roomPos.y, roomPos.y, checkWalkable, where);
+};
 mod.validFields = function(roomName, minX, maxX, minY, maxY, checkWalkable = false, where = null) {
     let look;
     if( checkWalkable ) {
--- a/room.js
+++ b/room.js
@@ -1386,11 +1386,21 @@ mod.getCostMatrix = function(roomName) {
         var costMatrix = new PathFinder.CostMatrix;
 
         if( DEBUG ) logSystem(roomName, 'Calulating cost matrix');
-        let setCosts = structure => {
-            if(structure.structureType == STRUCTURE_ROAD && _.isUndefined(structure.progress)) {
-                costMatrix.set(structure.pos.x, structure.pos.y, 1);
-            } else if(structure.structureType !== STRUCTURE_RAMPART || !(structure.isPublic || structure.my) ) {
-                costMatrix.set(structure.pos.x, structure.pos.y, 0xFF);
+        let setCosts = function(structure) {
+            switch(structure.structureType) {
+                case STRUCTURE_CONTAINER:
+                    return;
+                case STRUCTURE_ROAD:
+                    if (_.isUndefined(structure.progress)) {
+                        costMatrix.set(structure.pos.x, structure.pos.y, 1);
+                    }
+                    return;
+                case STRUCTURE_RAMPART:
+                    if (structure.isPublic || structure.my) {
+                        return;
+                    }
+                default:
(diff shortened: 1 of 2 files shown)


karlthepagan:
`[7:51:52 PM]17222157 Creep precalc failed {"creepName":"reserver-alphaLMine-1","pos":{"x":49,"y":40,"roomName":"E46N81"},"finalPos":{"x":24,"y":23,"roomName":"E45N82"},"targetPos":{"x":25,"y":25,"roomName":"E45N82"},"range":23,"maxRooms":5,"route":["E45N81","E45N82","E46N81"],"path":53,"getPath":"precalc failed","Creep":"getPath"}`

This is the only standard room that's giving me a problem now. Rebasing my fixes and looking at @chobobobo review.

whardeman:
@karlthepagan @chobobobo Is there a max for a room's cost?

karlthepagan:
I'll come back to this in a few days. It's pretty tricky.

karlthepagan:
@chobobobo @whardeman please re-review

This just went up to the next level. I heavily referenced traveler.js

karlthepagan:
I declare victory over pathfinding: https://screeps.com/a/#!/history/E44N80?t=17331140

karlthepagan:
This is not stable in rooms where construction sites are frequently placed.

karlthepagan:
I would like to promote other features into dev so that this can be tested in canary.

Almaravarion:
I was asked to note the following issue:

When creating new room with current canary build spawns create miners automatically, at which point all workers start ignoring the sources; While initially not that big of a problem given initial mining capabilities, same issue transfers to pioneers, forcing pioneers to wait for miners with half of their work part capabilities to obtain energy; Later on once spawners can take care of 5 work miners that is not an issue, as miner will mine entire potential but initally, extremely notably in RCL 2 and 1, this will weaken the room buildup.

chobobobo:
@Almaravarion you're quite right, I don't think that's related to this issue, but a fix for rooms with more than one access point to a source is #702 a more detailed fix would involve large work pioneers taking the role of miners for low RCL rooms.

chobobobo:
@karlthepagan I will try merging this into my testing branch again and profile once more.

chobobobo:
This now function as expected.  The CPU cost is still high in general on current builds, but I don't think this is contributing significantly to that increase.  The one area that still seems to hit us hard is honking because it involves regenerating a large route taking creeps into account (~8CPU vs ~2).

room.js
```diff
@@ -1389,9 +1371,48 @@ mod.roomDistance = function(roomName1, roomName2, diagonal, continuous){
...
+    if (room) {
+        var costMatrix = new PathFinder.CostMatrix;
+
+        if( DEBUG ) logSystem(roomName, 'Calulating cost matrix');
+        let setCosts = function(structure) {
+            switch(structure.structureType) {
+                case STRUCTURE_CONTAINER:
+                    return;
+                case STRUCTURE_ROAD:
+                    if (_.isUndefined(structure.progress)) {
```
logxen: I recommend simply dropping lines 1394 and 1396 in room.js. This will allow unbuilt roads to be treated just like normal roads. This is actually a good thing that helps pathing through broken rooms. It works particularly well when combined with the remote hauler drive by building feature.

logxen: Just thought of this ... you could alternately add "&& structure.room.controller && !structure.room.controller.my" to the condition to continue ignoring unbuilt roads in owned rooms.

karlthepagan:
on hold in favor of #751

karlthepagan:
Closing in favor of traveler. Will leave the branch around to learn from.

PR #668, opened 2017-02-05
Change/mining reservers

karlthepagan:
this requires #655 

an experimental change which de-prioritizes reservers until a remote mine's haulers have been built


karlthepagan:
Related to #621

karlthepagan:
after observing this casually I'm decreasing the weight requirement to 20%

karlthepagan:
rebased onto #655

pandabear41:
Line 237 and 243. The logic changed for reservation checking and current code was causing reservers to attempt to spawn for SK rooms. 
e72cb34d8462ee5a0baf32129a7c1ed92438d999

karlthepagan:
> but I don't believe people want to take into account queued creeps when considering capacity, only running or spawning creeps.

That's what happens here, but I guess it's not super clear. The only time queued creeps are considered is when calculating creeps to put into the queue.

karlthepagan:
The final requirement is 10% of capacity. This basically means that a miner and hauler will spawn first before the reserver, but once that is setup the reserver's medium priority takes over.

task.mining.js
```diff
@@ -421,15 +421,30 @@ function haulerCarryToWeight(carry) {
...
+
+            if( population < REMOTE_RESERVE_HAUL_CAPACITY ) {
+                // TODO if this room & all exits are currently reserved (by anyone) then use default to prevent Invaders?
+                if( DEBUG && TRACE ) trace('Task', {flagName:flag.name, pos:flag.pos, population, spawnParams:'population', [mod.name]:'spawnParams', Task:mod.name});
+                return function() { // TODO how to pass thru partial results?
+                    return {count: 0, priority: 'Low'};
+                };
+            }
+
+            // undefined means use defaults
```
karlthepagan: This is non-standard for strategies so I'm having a real good think about it before I merge into dev.

PR #671, opened 2017-02-06
added build nearby for remote haulers with range 1 to Creep.repairNearby

logxen:
This small addition allows remote haulers to build their own roads and containers. It is implemented along with Creep.repairNearby and only searches for construction sites within a range of 1 to save cpu time. Depending on strategies this can make the remote worker superfluous. Experimenting can be done with setting the REMOTE_WORKER_MULTIPLIER in the parameter file to 0 to save on spawn time.

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -310,16 +310,19 @@ mod.extend = function(){
         // if it has energy and a work part
         if(this.carry.energy > 0 && this.hasActiveBodyparts(WORK)) {
             let nearby = this.pos.findInRange(this.room.structures.repairable, 3);
-            if( nearby && nearby.length > 0 ){
+            if( nearby ){
                 if( DEBUG && TRACE ) trace('Creep', {creepName:this.name, Action:'repairing', Creep:'repairNearby'}, nearby[0].pos);
                 this.repair(nearby[0]);
             } else {
                 if( DEBUG && TRACE ) trace('Creep', {creepName:this.name, Action:'repairing', Creep:'repairNearby'}, 'none');
                 // enable remote haulers to build their own roads and containers
-                if( this.data && this.data.creepType == 'remoteHauler' ) {
+                if( REMOTE_HAULER_DRIVE_BY_BUILDING && this.data && this.data.creepType == 'remoteHauler' ) {
                     // only search in a range of 1 to save cpu
-                    let nearby = this.pos.findInRange(this.room.constructionSites, 1);
-                    if( nearby && nearby.length > 0 ){
+                    let nearby = this.pos.findInRange(this.room.constructionSites, 1, {filter: (site) =>{
+                        return site.structureType == STRUCTURE_CONTAINER ||
+					site.structureType == STRUCTURE_ROAD;
+                    }});
+                    if( nearby ){
                         if( DEBUG && TRACE ) trace('Creep', {creepName:this.name, Action:'building', Creep:'buildNearby'}, nearby[0].pos);
                         this.build(nearby[0]);
                     } else {
--- a/creep.js
+++ b/creep.js
@@ -310,7 +310,7 @@ mod.extend = function(){
         // if it has energy and a work part
         if(this.carry.energy > 0 && this.hasActiveBodyparts(WORK)) {
             let nearby = this.pos.findInRange(this.room.structures.repairable, 3);
-            if( nearby ){
+            if( nearby && nearby.length ){
                 if( DEBUG && TRACE ) trace('Creep', {creepName:this.name, Action:'repairing', Creep:'repairNearby'}, nearby[0].pos);
                 if( this.repair(nearby[0]) == OK && this.carry.energy <= this.getActiveBodyparts(WORK) * REPAIR_POWER / REPAIR_COST ) {
                     Creep.action.idle.assign(this);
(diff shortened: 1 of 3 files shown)


logxen:
In my experience having the miner do it causes a situation where the miner wastes half his time and so doesn't mine all the energy and refuses to share, blocking the chain and usefulness of any other creeps in the team. I drop the worker entirely and let the miner drop-mine until the container eventually finishes. I don't think that drop mining is really that big of a problem while the room is spinning up and the containers do get finished.

Yeah I can have it filter for roads and containers ... personally I don't mind the drive-by building but that's a good idea for the official repo.

I also have a couple other small updates in mind so don't merge this quite yet.

chobobobo:
Actually, I had misinterpreted this change thinking that they would stop and build.  Drive-by building is fine by me, and I agree that dropmining is plenty efficient until the container gets built, keeping the miner mining until the source is empty is most important.  This change, plus the miner building at the container after it has exhausted the source sounds great to me.

logxen:
Ok ... I made the containers and roads only thing an option just in case someone objects. In my opinion if the work part and the energy is there and it isn't busy then take advantage of the opportunity. :)

I added a feature to have the remote hauler go idle if they manage to drain themselves of energy. This can happen pretty easily in a swamp. On the bright side they build the road through the swamp pretty fast.

Miners should now also behave as workers if the source is empty as @chobobobo requested.

I think this has everything I wanted to add now and is ready to commit upstream.

creep.js
```diff
@@ -310,11 +310,29 @@ mod.extend = function(){
         // if it has energy and a work part
         if(this.carry.energy > 0 && this.hasActiveBodyparts(WORK)) {
             let nearby = this.pos.findInRange(this.room.structures.repairable, 3);
-            if( nearby && nearby.length > 0 ){
+            if( nearby ){
                 if( DEBUG && TRACE ) trace('Creep', {creepName:this.name, Action:'repairing', Creep:'repairNearby'}, nearby[0].pos);
-                this.repair(nearby[0]);
-            } else {
+                if( this.repair(nearby[0]) == OK && this.carry.energy <= this.getActiveBodyparts(WORK) * REPAIR_POWER / REPAIR_COST ) {
+                    Creep.action.idle.assign(this);
```
karlthepagan: Perhaps this should be an event instead of an immediate action assignment.

karlthepagan:
`canary` error

TypeError: Cannot read property 'pos' of undefined
    at Creep.repairNearby (creep:320:127)
    at Creep.run (creep:45:18)
    at run (creep:446:30)
    at Function.mod.execute (creep:447:7)
    at Object.module.exports.loop (main:269:11)

creep.js
```diff
@@ -310,11 +310,29 @@ mod.extend = function(){
...
+                };            } else {
```
karlthepagan: remove the trailing `;` from the if clause
put `} else {` on it's own line

creep.js
```diff
@@ -310,11 +310,29 @@ mod.extend = function(){
...
+                    let nearby = this.pos.findInRange(this.room.constructionSites, REMOTE_HAULER_DRIVE_BY_BUILD_RANGE, {filter: (site) =>{
+                        return site.my && REMOTE_HAULER_DRIVE_BY_BUILD_ALL ||
+                            (site.structureType == STRUCTURE_CONTAINER ||
+                            site.structureType == STRUCTURE_ROAD);
+                    }});
+                    if( nearby ){
+                        if( DEBUG && TRACE ) trace('Creep', {creepName:this.name, Action:'building', Creep:'buildNearby'}, nearby[0].pos);
+                        if( this.build(nearby[0]) == OK && this.carry.energy <= this.getActiveBodyparts(WORK) * BUILD_POWER ) {
+                            Creep.action.idle.assign(this);
+                        };
```
karlthepagan: remove `;` here, it does nothing

creep.js
```diff
@@ -310,11 +310,29 @@ mod.extend = function(){
         // if it has energy and a work part
         if(this.carry.energy > 0 && this.hasActiveBodyparts(WORK)) {
             let nearby = this.pos.findInRange(this.room.structures.repairable, 3);
-            if( nearby && nearby.length > 0 ){
+            if( nearby ){
```
karlthepagan: this needs to be `if( nearby && nearby.length )` see bug report

karlthepagan:
I had to take this out of canary because it was breaking me.

creep.js
logxen: yeah @Spedwards said it was redundant ... I'll put it back

karlthepagan:
Still missing one of the code cleanup fixes ( else on same line and not aligned properly)

are your carriage returns configured incorrectly?

karlthepagan:
I still see extra semicolons.

Also I'm running into issues with haulers getting stuck repairing containers: https://screeps.com/a/#!/history/E42N82?t=17279777

logxen:
OK I see the extra semicolon, bah. :)

That's strange behavior ... you are running on canary aren't you? I have only tested this on dev and that's very strange behavior. This code does not affect movement at all to my understanding.

karlthepagan:
Giving a creep over to a different behaviour loop will do that to the movement I think.

karlthepagan:
IMO this is ready. When a miner is present repairing the container won't block up a hauler.

logxen:
OK hold up I found the bug that is causing the behavior you mentioned.

PR #674, opened 2017-02-06
Feature/room visuals

Spedwards:
Add basic information using room visuals. Each have their own parameter, so you can customise freely.


karlthepagan:
I threw this into `canary`

Spedwards:
Latest commit includes a redesign of the "room" visual. It also includes WIP code for Storage and Terminal contents (just need to fix the positioning).

Tomorrow I'll work on heatmaps.

karlthepagan:
rampart visuals overlap other important structures, maybe the rampart stats should go on the left as a quick fix?

torin555:
ROOM_GLOBALS setting doesn't work, it still draws all the account information like GCL and CPU

karlthepagan:
would be really great to have a visual indication of recent market transactions above or below a terminal
maybe something as simple as `+${sellCredits}` or `-${buyCredits}`

karlthepagan:
feel free to rebase instead of reverting.

I know that it is slightly risky, but it makes automatic merging easier when fewer files are in the changeset.

Spedwards:
After enabling heatmaps in your parameters, run `Memory.heatmap = true;` to display the heatmap.

Spedwards:
Converted to ES6 class. We have to start upgrading somewhere. Functions the same, and doesn't affect viral files modifying it.

whardeman:
Yay! ES6! (Except I don't know ES6 yet!) 

*Googles ES6 tutorials*

visuals.js
```diff
@@ -0,0 +1,356 @@
...
+		if (controller.ticksToDowngrade < CONTROLLER_DOWNGRADE[controller.level]) {
+			let downgradeStyle = Object.assign({}, style, {color: '#FF0000'});
+			vis.text(`D: ${formatNum(controller.ticksToDowngrade)}`, BASE_X, y += 0.4, downgradeStyle);
+		}
+	}
+	
+	static highlightWeakest(room, type) {
+		const vis = new RoomVisual(room.name);
+		let weakest = _(room.find(FIND_STRUCTURES)).filter(s => s.structureType === type).min(s => s.hits);
+		if (weakest) {
```
karlthepagan: `		if (weakest && weakest.pos) {`

Spedwards: I was going to say that's strange, but then I remembered that `_.min` returns `Infinity` when undefined...

PR #677, opened 2017-02-06
patch to avoid statistics for SKs which causes an error

logxen:
This patch addresses internal issue # 185. Fix from georgeberkley.

### DIFF
--- a/room.js
+++ b/room.js
@@ -1148,6 +1148,7 @@ mod.extend = function(){
             this.memory.statistics = {};
 
         let registerHostile = creep => {
+            if(creep.owner.username == "Source Keeper") return;
             // if invader id unregistered
             if( !that.memory.hostileIds.includes(creep.id) ){
                 // handle new invader


Spedwards:
@karlthepagan @logxen I don't think this is the problem. I think it's _any_ hostile in the SK rooms. I've had two instances of getting the JSON error and having regular Invaders in the room.

whardeman:
@karlthepagan was your solution Task.reputation.NPC or something like that?

pandabear41:
This solution doesn't fix anything. Have been running it for a while and I still get the error all the time.

whardeman:
@pandabear41 I think the reality is that this is a generic error that correlates to certain conditions. This fix is for only one of the conditions under which it has been observed.

The error can occur _any time_ self- or circular-referential data is placed into Memory. An example of such came up just yesterday in @Spedwards work on visuals. He was assigning some property (eg `someProp`) to creep.data, then setting another object property equal to creep.data.someProp, then setting creep.data.someProp equal to the new object's someProp. When that happens, it breaks whenever the engine tries to write the creep's data to Memory because it can't serialize the self-reference.

Spedwards:
@whardeman I actually think it was a coincidence. There was an invader in the SK room at the same time I saved the visuals file.

karlthepagan:
I have experienced this while testing dev + traveler.js

The circular reference was resolved only after deleting all of my privateers who were gathering energy.

PR #683, opened 2017-02-07
Change/uncharging strategy

karlthepagan:
please merge https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/471 first

workers will now pick from the closest container that puts them to work
haulers continue to pick from best-fit container
partially implemented minimum energy amount for managed containers

PR #685, opened 2017-02-08
Change/picking strategy

karlthepagan:
Supersedes #683


karlthepagan:
rebased onto the rebased #683

creep.behaviour.remoteMiner.js
```diff
@@ -38,3 +38,20 @@ mod.approach = function(creep){
     }
     return range;
 };
+mod.strategies = {
+    defaultStrategy: {
+        name: `default-${mod.name}`,
+        // TODO don't move more than a few tiles to withdraw or pick
+        canWithdrawEnergy: function (creep) {
```
chobobobo: Is this going to make remoteMiners wander off currently?

creep.behaviour.remoteHauler.js
```diff
@@ -96,3 +96,43 @@ mod.gotoTargetRoom = function(creep){
...
+        name: `default-${mod.name}`,
+        canWithdrawEnergy: function (creep) {
+            const min = Math.min(creep.carryCapacity - creep.sum, 250);
+
+            return function (amount) {
+                return amount >= min;
+            };
+        }
+    },
+    uncharging: {
```
chobobobo: Is this identical to hauler, should it perhaps just be set to the hauler's strategy instead?

chobobobo:
@karlthepagan is this PR on hold?  If so, I will proceed with merging #1052 as it is required for SKMining haulers to pick up minerals

karlthepagan:
There's too many existing issues to test in 0.7.0 but I think this can fit into 0.7.1 if the change set is small.

The major issue is getting more new rooms and constructions observed.

PR #688, opened 2017-02-08
More sophisticated mapping of terminal to source

### LINKED ISSUE #681: Broken Source
One of my sources seems to be totally broken. I've cleared room memory, suicided the miners, and destroyed the container. (It's being rebuilt but the room doesn't have enough energy to finish it.)
screenshot_19

whardeman:
- potentially resolves #681

### DIFF
--- a/room.js
+++ b/room.js
@@ -957,7 +957,22 @@ mod.extend = function(){
         containers.forEach(add);
 
         if( this.terminal ) {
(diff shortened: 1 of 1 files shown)


whardeman:
@torin555 You've got the best live use case for this. If we can get it into a canary release for you, perhaps you can test and report back?

Issue #657, opened 2017-02-02
task.robbing cannot handle abandoned structures


karlthepagan:
https://screeps.com/a/#!/history/E48N82?t=17167802

If that data is saved you'll see a pair or robbers circling around the terrain at the center of the room. These were trying to walk onto the flag that was their travel target.

There are several links, storage, and an extension which should be priority robbing targets.

This was a room abandoned and I'm trying to recover the energy.

pandabear41:
@karlthepagan Was there a rampart under the structure?

karlthepagan:
there are no ramparts in my abandoned room

Dasku:
I'm testing this now with the proposed master branch and its working fine. Robbers emptied all the extensions and now are robbing the storage as expected.  Close for now?

karlthepagan:
:+1:

karlthepagan:
Must be a problem with canary. Some time after this my robbers sat and did nothing

https://screeps.com/a/#!/history/E71S7?t=17396239

karlthepagan:
Yeah I think that's the issue. I use robbing a lot to try to pull from abandoned rooms.

PR #693, opened 2017-02-10
Feature/delivery task

karlthepagan:
this is based on improved remote mining, and it might need to be rebased

regardless its fairly stable and has helped debug #665 

Use a Red/Yellow flag in an owned room to spawn a recycler from that room who will deliver your energy from that room to another room. Privateers / robbers are suppressed for friendly owned rooms.

Use a Green/Yellow flag to specify the destination.


Krazyfuq:
working well, you must have a container/storage in destination room. Did not work in a room with the white/yellow drop flag.

karlthepagan:
still need to tweak recycler behaviour - it should select the closest spawn as the default target instead of the home room

karlthepagan:
#657 was caused by this change probably

https://screeps.com/a/#!/history/E71S7?t=17396239 - robbers not withdrawing from structures.

karlthepagan:
looks like this isn't picking/uncharging before it travels, probably an oversight

karlthepagan:
privateers are still attracted to robbing/delivery flags and exhibit the issue in #657

karlthepagan:
Flag memory is getting stuck - I'm losing creeps when they start spawning. Looks like this needs to be re-validated a tick after spawning starts.

karlthepagan:
I have observed a few times today, creeps stuck in spawning and also creeps recycling in the room where they are born.

karlthepagan:
now depends on #788

chobobobo:
@karlthepagan this is still not working for me.  I created a red/yellow in E87S28 and a green/yellow in E89S29, 2 robbers spawned in E86S27, travelled to E87S28 and then returned to E86S27 to function as haulers in that room.  They later returned to E87S28 where there was some energy dropped on the ground, picked that up and returned it to E86S27 (their spawnRoom).  So it would appear they are correctly robbing the intended room, but their destination is not getting set properly.

As a caveat, this is Canary, so traveler is also included, it's not impossible that there is a problem related to that.

karlthepagan:
this has problems with traveler so I'm putting it on hold

creep.action.withdrawing.js
```diff
@@ -22,3 +16,11 @@ action.onAssignment = function(creep, target) {
     //if( SAY_ASSIGNMENT ) creep.say(String.fromCharCode(9738), SAY_PUBLIC);
     if( SAY_ASSIGNMENT ) creep.say('\u{1F4E4}\u{FE0E}', SAY_PUBLIC);
 };
+action.defaultStrategy.isValidAction = function(creep) {
+    return (
+        creep.room.storage &&
+        creep.room.storage.store.energy > 0  &&
+        creep.sum < creep.carryCapacity &&
+        (!creep.room.conserveForDefense || creep.room.relativeEnergyAvailable < 0.8)
+    ) || false;
```
TooAngel: IMHO the `|| false` doesn't do anything.

creep.behaviour.privateer.js
```diff
@@ -154,3 +155,17 @@ mod.exploitNextRoom = function(creep){
...
+    withdrawing: {
+        name: `withdrawing-${mod.name}`,
+        isValidAction: function(creep) {
+            return false;
+        },
+    },
+};
+mod.selectStrategies = function(actionName) {
+    return [mod.strategies.defaultStrategy, mod.strategies[actionName]];
+};
```
TooAngel: Nitpicking, new line at the end of the file.

creep.behaviour.recycler.js
```diff
@@ -19,3 +19,84 @@ mod.run = function(creep) {
...
+        Creep.action.feeding,
+        Creep.action.dropping,
+        Creep.action.recycling,
+        Creep.action.idle,
+    ];
+};
+mod.nextAction = function(creep){
+    const priority = mod.actionPriority();
+
+    for(var iAction = 0; iAction < priority.length; iAction++) {
```
TooAngel: You could also use `for (var action of priority) {`

creep.behaviour.recycler.js
```diff
@@ -19,3 +19,84 @@ mod.run = function(creep) {
...
+            return !creep.sum; // only recycle when empty
+        },
+    },
+    uncharging: {
+        name: `uncharging-${mod.name}`,
+        isValidAction: function(creep){
+            return (
+                creep.data.travelRoom && // only gather when on mission
+                creep.sum < creep.carryCapacity
+            ) || false;
```
TooAngel: Same here, is the `|| false` necessary?

room.js
```diff
@@ -897,36 +897,40 @@ mod.extend = function(){
...
+                let parsed = /^[WE]([0-9]+)[NS]([0-9]+)$/.exec(roomName);
+                isHighway = (parsed[1] % 10 === 0) || (parsed[2] % 10 === 0);
+            }
+            let isMyOrNeutralRoom = false;
+            if( checkOwner ){
+                let room = Game.rooms[roomName];
+                isMyOrNeutralRoom = room &&
+                    room.controller &&
+                    (room.controller.my ||
+                    (room.controller.owner === undefined));
```
TooAngel: I like the `preferHighway` idea. But I think the order should be `my/my reserved rooms > highway > neutral rooms`.
E.g. AzurStar blocks and well defends their reserved rooms, so avoiding these rooms is beneficial.

I even check and mark rooms which I can not pass, due to walls, and exclude them.

task.delivery.js
```diff
@@ -0,0 +1,153 @@
...
+//     if( !(entry && entry.destiny && entry.destiny.task === mod.name) ) {
+//         return;
+//     }
+//     const flag = Game.flags[entry.destiny.targetName];
+//     if (flag) {
+//         const running = mod.memory(flag.pos.roomName).running;
+//         const index = _.indexOf(running, creepName);
+//         running.splice(index, 1);
+//     }
+// };
```
TooAngel: Usually I say, commented code doesn't belong into a PR

creep.action.withdrawing.js
Spedwards: Have to agree. `|| false` will never equate. Either the initial statement is `true` and `|| false` doesn't matter, or it's `false` and you end up with `false || false`. Just omit the `|| false`.

karlthepagan: The result of this logical expression is not just true or false, but it can be any truthy or falsey value contained. This includes `undefined` when there is no storage.

`|| false` prevents the logic from ever equating to undefined and I would encourage everyone to use this as a safety check in all strategy handlers. This is very important because a return `=== undefined` looks like an error to the strategy API.

creep.behaviour.recycler.js
karlthepagan: When `creep.data.travelRoom === undefined`.

creep.action.withdrawing.js
Spedwards: That's not true though. If there is no storage, the first check fails because `creep.room.storage` will be falsey.

karlthepagan: And without `|| false` that will return undefined which violates the requirements of the API.

Try this on the REPL:
```
> undefined && true && true && true
< undefined
> (undefined && true && true && true) || false
< false
```

Spedwards: Yeah, I tested that once I realised what you meant. It's a weird interaction. Even `true && undefined && true` would return `undefined`.

karlthepagan:
I think the conflicts are fixed.

chobobobo:
I'll put this in canary today

chobobobo:
@karlthepagan I just merged dev in, but I'm concerned that there may now be duplicate code in Creep.Action.js, I'm going to leave this out of the current canary until you've had a chance to review the merge.

creep.action.withdrawing.js
```diff
@@ -51,9 +51,10 @@ action.assignDebounce = function(creep, outflowActions) {
...
 action.defaultStrategy.isValidAction = function(creep) {
     return (
-        creep.room.storage &&
-        creep.room.storage.store.energy > 0  &&
+        ((creep.room.storage && creep.room.storage.store[RESOURCE_ENERGY]) ||
+        (creep.room.terminal && creep.room.terminal.store[RESOURCE_ENERGY])) &&
+        creep.data.creepType !== 'privateer' &&
         creep.sum < creep.carryCapacity &&
         (!creep.room.conserveForDefense || creep.room.relativeEnergyAvailable < 0.8)
-    ) || false;
```
karlthepagan: strategy handlers must be guaranteed to never return undefined

karlthepagan:
@chobobobo details of the validation logic in delivery task

I tried very hard to make sure that creeps transitioned between spawning states accurately. This is backed up by intermittent spawning validation code.

Delivery task does not validate running each tick which should be a modest CPU savings if we extend this to all tasks.

creep.behaviour.recycler.js
```diff
@@ -3,19 +3,97 @@ module.exports = mod;
...
+        name: `recycling-${mod.name}`,
+        isValidAction: function(creep){
+            return !creep.sum; // only recycle when empty
+        },
+    },
+    uncharging: {
+        name: `uncharging-${mod.name}`,
+        isValidAction: function(creep){
+            return (
+                creep.data.travelRoom && // only gather when on mission
```
chobobobo: This use of travelRoom will be in conflict with Creep.action.travelling I believe

creep.behaviour.recycler.js
```diff
@@ -3,19 +3,97 @@ module.exports = mod;
...
+                creep.sum < creep.carryCapacity
+            ) || false;
+        },
+    },
+    withdrawing: {
+        name: `withdrawing-${mod.name}`,
+        isValidAction: function(creep) {
+            return (
+                creep.data.travelRoom && // only gather when on mission
```
chobobobo: also here

task.delivery.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    // assign destination flag?
+};
+mod.handleSpawningCompleted = function(creep) {
+    // ensure it is a creep which has been queued by this task (else return)
+    if ( !creep.data || !creep.data.destiny || !creep.data.destiny.task || creep.data.destiny.task != mod.name )
+        return;
+    creep.data.homeRoom = creep.data.destiny.homeRoom || creep.data.homeRoom;
+    creep.data.travelRoom = creep.data.destiny.targetRoom || creep.data.travelRoom;
```
chobobobo: creep.data.travelRoom is used by Creep.action.travelling and is deleted upon completion of that action, this will need to be renamed to function as intended I believe.

karlthepagan: These strategy functions are intended to specify the travel target for the delivery creep.

I'm open to suggestions for how to improve it or make this more clear.

chobobobo: deliveryRoom?

karlthepagan: That is the intention, but I'm willing to consider a change... maybe a strategy for newTarget

chobobobo:
seeing this 
```
629237 error handler returned undefined for args {"agent":"remoteWorker-W6N9-mine-1","strategyKey":["withdrawing","remoteWorker","mining"],"strategyName":["default-withdrawing"],"method":"isValidAction","args":"[creep remoteWorker-W6N9-mine-1]","stack":"Error\n    at prototype.getStrategyHandler (strategy:26:151)\n    at Action.action.isValidAction (creep.action.withdrawing:4:18)\n    at Object.mod.nextAction (creep.behaviour.remoteWorker:50:23)\n    at Object.mod.run (creep.behaviour.remoteWorker:18:14)\n    at Creep.run (creep:72:27)\n    at run (creep:382:19)\n    at /app/node_modules/lodash/index.js:3073:15\n    at baseForOwn (/app/node_modules/lodash/index.js:2046:14)\n    at /app/node_modules/lodash/index.js:3043:18\n    at Function. (/app/node_modules/lodash/index.js:3346:13)"}  
```

chobobobo:
I'm going to dig in here, update it further to match the centralized spawning checks introduced in the small task refactor and make it work.

chobobobo:
@karlthepagan please merge this if you approve of the small changes, I have now had a chance to test and run this and it works.

karlthepagan:
ok that works and I'm unavailable for the near term

I'm very very interested in moving task accounting toward something which does specific accounting (like the first version of delivery did - meaning find the one that died and splice it) and only using memory validation as a backup (a kind of integrity checkpoint in our fault-tolerant system).

Issue #698, opened 2017-02-10
remoteHaulers being set to workers


OverlordQ:
Somehow my remoteHaulers are getting their creepType set to 'worker' so they get confused, load up on energy, and try to start building with their 1 WORK part, effectively spending their entire lifetime trying to build the same object.

They should fall back to hauling not working.

chobobobo:
@logxen I think it's possible what happened here was related to #593, if their type has been set to worker, that means their population entry was missing for whatever reason and it defaulted.  I have seen this infrequently as well and currently the only fix is to manually change their creepType back to remoteHauler.

karlthepagan:
related to #593 - dying creep workflow

OverlordQ:
Yeah,  likely issue.

PR #699, opened 2017-02-10
cancel current actions instead of assigning remote haulers to idle

logxen:
This hopefully addresses issue #698

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -28,9 +28,9 @@ mod.extend = function(){
             }
         }
     };
-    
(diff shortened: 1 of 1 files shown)


chobobobo:
@logxen please see my comments in #698, setting them to idle should not affect their creepType as I understand it.

logxen:
Heh, ok. Thanks, @chobobobo. I think I just panicked and assumed it was my fault because my pr affecting haulers was merged that morning. I'll go ahead and close this. :)

PR #701, opened 2017-02-11
RCL 2&3 haulers will drop at spawn and not move around work parts

karlthepagan:
another respawn tweak, also adds the dropping action to in-room haulers

### DIFF
--- a/creep.behaviour.hauler.js
+++ b/creep.behaviour.hauler.js
@@ -35,6 +35,7 @@ mod.nextAction = function(creep){
             Creep.action.charging,
             Creep.action.fueling,
             Creep.action.storing,
+            Creep.action.dropping,
             Creep.action.idle];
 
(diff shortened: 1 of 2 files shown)


karlthepagan:
this needs strategies or the creeps start cycling on the energy at their feet

karlthepagan:
I really like the debounce API. I'm looking into this algorithm a little bit now.

karlthepagan:
@chobobobo please let me know which debounce strategy is capable of resolving this in 0.7

without a proposal I will have to find a quick fix

karlthepagan:
The alternative is that we must remove haulers from the default configuration until RCL3

chobobobo:
If you want to debounce, you would need something similar to withdraw's assignDebounce that checks all available outflow energy uses before it assigns pickup.  I think we should also find some way to cache those checks otherwise we are checking all outflow actions twice.

PR #711, opened 2017-02-11
Pick harvest source based on distance first

OverlordQ:
More of an issue on new rooms, looping niavely through list of sources can occassionally send creeps across room when a valid source is right next to them


karlthepagan:
:+1:

I'll get this into canary. Would you be interested in turning this into a score function so that miners and workers can behave differently?

PR #714, opened 2017-02-12
Change/unify mining

chobobobo:
This change merges the remote and local miner run loops because they share a majority of their logic.  Local miners will look for links as well.  Local miners will fortify & build structures within their range, remote miners will repair structures within their range.  They will also build their own containers if one is not present provided you set MINERS_AUTO_BUILD = true in your viral.parameters.js.  I also increased the work parts on local miners so that they can manage the capacity.


karlthepagan:
I assume that local haulers will be affected by `Task.mining.carry`

Having a look

karlthepagan:
https://screeps.com/a/#!/history/E74S7?t=17461855

workers are returning to harvest sources even when the miners are capable of exhausting that source

creep.behaviour.miner.js
```diff
@@ -29,7 +29,7 @@ mod.run = function(creep, params = {approach: mod.approach}) {
     }
 
     if( source ) {
-        if( !creep.action ) Population.registerAction(creep, Creep.action.harvesting, source);
+        Population.registerAction(creep, Creep.action.harvesting, source);
```
karlthepagan: I'm going to cherry-pick this into a master fix just in case.

karlthepagan: On second thought, registerAction does a lot of book keeping. Can you replace the check: `if( !creep.action || creep.action.name !== 'harvesting' )`

karlthepagan:
This one is next. Tagging.

karlthepagan:
I rebased this onto dev btw.

PR #715, opened 2017-02-12
Stop remote miners and haulers from border cycling causing a mess.

chobobobo:
Quick fix to prevent border cycling.  @whardeman can you try this out to see if it resolves your issues.  Any larger implementation to your issues would not be appropriate for a merge into RC so this will have to do for now.


chobobobo:
I believe this patch is ready now, it prevents the border cycling experienced without adding too much complexity.
remoteHauler - empty ? go to nearest source : idle
remoteMiner - (source ? go there : go to flag) then idle

PR #725, opened 2017-02-13
Rc/miner replacement

karlthepagan:
this is a bugfix / enhancement for the MASTER BRANCH

please review and I am going to merge to the new RC


creep.action.harvesting.js
```diff
@@ -4,9 +4,14 @@ action.renewTarget = false;
...
 };
-action.isValidTarget = function(target){
-    return (target != null && target.energy != null && target.energy > 0 &&
-        (target.targetOf === undefined || !_.some(target.targetOf, {'creepType': 'miner'}) ));
+action.isValidTarget = function(target) {
+    return (target !== null && target.energy !== null && target.energy > 0 &&
+        (target.targetOf === undefined || !_.some(target.targetOf, c =>
+            (c.creepType === 'miner' || c.creepType === 'remoteMiner')
+                && c.body.work >= 5
+                && (creep.ticksToLive || CREEP_LIFE_TIME) >= (creep.data.predictedRenewal || 0)
```
karlthepagan: this is new, going to move it into dev

karlthepagan: this is new, also note the inversion: not some are "miners with work parts and have big TTLS"

whardeman: `creep` is undefined, so `c.ticksToLive`

`c.data.predictedRenewal` will also need an undef check on `c.data`

karlthepagan: oh i'm bad thanks

creep.action.harvesting.js
```diff
@@ -9,7 +9,7 @@ action.isValidTarget = function(target) {
         (target.targetOf === undefined || !_.some(target.targetOf, c =>
             (c.creepType === 'miner' || c.creepType === 'remoteMiner')
                 && c.body.work >= 5
-                && (creep.ticksToLive || CREEP_LIFE_TIME) >= (creep.data.predictedRenewal || 0)
+                && (c.ticksToLive || CREEP_LIFE_TIME) >= (c.data.predictedRenewal || 0)
```
whardeman: c.data.predictedRenewal also throws errors if c.data is missing, but maybe we're ok with that (if there are other things handling the unfortunate "creep has lost it's mind, literally" situation)

karlthepagan: I'll review that in the other actions.

karlthepagan: Creep.prototype.data is a property so it won't be missing. We commonly check for creep.data.destiny but I think we'll be ok without the check here.

whardeman: I only brought it up because I got the error immediately upon adding my change. I think it is missing when creeps get the memory blown up, but I suppose we're addressing the cause of that elsewhere

karlthepagan: I see it, something that's not a creep must be targeting? Well... i'll add that fix for now.

PR #728, opened 2017-02-14
player reputation is case insensitive

### LINKED ISSUE #710: Whitelist doesn't work
My creeps will attack enemy creeps, friends within the whitelist, and even my own creeps.
This only happened after of the task.repuatition.

See this:
http://screeps.com/s/cRmmJx

karlthepagan:
I think this is the last must fix for 2017-2

this resolves #710

### DIFF
--- a/task.reputation.js
+++ b/task.reputation.js
@@ -26,15 +26,18 @@ const mod = {
         const reps = mod.cache('score');
         if( username === undefined ) {
             return reps;
-        } else if( reps[username] ) {
-            return reps[username];
+        }
+        const name = username && username.toLowerCase();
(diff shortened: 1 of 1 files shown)


karlthepagan:
I do need to test this

task.reputation.js
```diff
@@ -26,15 +26,18 @@ const mod = {
         const reps = mod.cache('score');
         if( username === undefined ) {
             return reps;
-        } else if( reps[username] ) {
-            return reps[username];
+        }
+        const name = username && username.toLowerCase();
```
chobobobo: You could move this up to line 27, but it still functions the way it is.

Issue #729, opened 2017-02-14
Workers should be disallowed from upgrading action at RCL8


whardeman:
Related to #727.

Non-upgrader creeps don't need to upgrade at RCL8 since the upgraders have to be limited in size anyway. (Also, if the room can't support a 15-WORK part upgrader, we have bigger problems that workers should be addressing.)

PR #730, opened 2017-02-14
Workers shouldn't upgrade at RCL8

whardeman:
#729

### DIFF
--- a/creep.action.upgrading.js
+++ b/creep.action.upgrading.js
@@ -9,7 +9,11 @@ action.isAddableAction = function(creep){
     // storage is leftover from invasion and has usable energy
     || (!creep.room.storage.my && creep.room.storage.store.energy > 0);
 };
-action.isAddableTarget = function(target){ return true; };
+action.isAddableTarget = function(target, creep){ 
+    // Limit to upgraders only at RCL8
+    if( target.level === 8 && creep.data.creepType != 'upgrader' ) return false;
+    return true;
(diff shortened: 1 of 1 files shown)


whardeman:
FWIW: This change has been tested live and appears to be working.

creep.action.upgrading.js
```diff
@@ -9,7 +9,11 @@ action.isAddableAction = function(creep){
     // storage is leftover from invasion and has usable energy
     || (!creep.room.storage.my && creep.room.storage.store.energy > 0);
 };
-action.isAddableTarget = function(target){ return true; };
+action.isAddableTarget = function(target, creep){ 
+    // Limit to upgraders only at RCL8
+    if( target.level === 8 && creep.data.creepType != 'upgrader' ) return false;
```
chobobobo: Perhaps just an undefined check creep.data && creep.data.creepType != 'upgrader

whardeman: If the creep has no data, we definitely don't want it upgrading at lvl8, but I'll add that `||` check so that works as well.

torin555:
Why should workers not upgrade at RCL 8? If you have excess energy you may as well pump it into your GCL

logxen:
@torin555 there is a cap of 15 control per tick once you hit RCL8 ... a large worker or upgrader will literally just stand doing nothing there unable to upgrade.

PR #731, opened 2017-02-14
Limit upgraders at Level 8 with separate setup

### LINKED ISSUE #727: Upgrading at RCL 8 is not being limited.
_A fully upgraded level 8 controller can't be upgraded over 15 energy units per tick regardless of creeps' abilities._ - [support.screeps.com](http://support.screeps.com/hc/en-us/articles/203013212-Creep)

At some point, the Creep.setup.upgrader default multiBody was changed and the limiter in maxMulti no longer does what it should.

A possible solution, as noted in a slack conversation, is to recognize that RCL8 is so unique in this way, it deserves it's own creep setup configuration.
```
setup.RC8 = {
    fixedBody: [CARRY, MOVE, MOVE, MOVE],
    multiBody: [WORK],
    minAbsEnergyAvailable: 300,
    minEnergyAvailable: 1,
    maxMulti: CONTROLLER_MAX_UPGRADE_PER_TICK / UPGRADE_CONTROLLER_POWER,
    maxCount: 1
};
setup.RCL.8 = setup.RC8;
```

This also allows for simplifying the maxMulti and maxCount methods since RCL8 conditions are no longer needed there.

whardeman:
resolves #727

### DIFF
--- a/creep.setup.upgrader.js
+++ b/creep.setup.upgrader.js
@@ -12,9 +12,8 @@ setup.maxMulti = function(room){
         let surplus = room.storage.store.energy - MAX_STORAGE_ENERGY[room.controller.level];
         multi += Math.ceil( surplus / 20000 ); // one more multi for each 20k surplus (+1)
     }
-    // at rcl 8 limit upgrading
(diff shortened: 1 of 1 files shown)


whardeman:
FWIW: This change has been tested live and appears to be working.

PR #735, opened 2017-02-14
Have creeps re-sign the controller if the message changes

OverlordQ:
I typo'd my original message and noticed creeps wont fix it if it's signed. Re-sign the controller if your message changes.


logxen:
I would give this an extra parameter ... I don't think everyone wants all rooms to say the same thing.

whardeman:
If CONTROLLER_SIGN_MESSAGE is an array, sign it with a random message from the list. 😄

logxen:
of course if one did BOTH of those the messages would start cycling every tick ...

PR #726, opened 2017-02-13
move reputation flush earlier in the main exec order

karlthepagan:
this is another MASTER BRANCH fix

maybe cyber had a point with reputation being lower level? maybe it just needs a flush? this might fix some inconsistencies with friend-or-foe code.

PR #738, opened 2017-02-14
OCS 0.6.0

karlthepagan:
This is the February 2017 release.

This is composed of dev plus the folowing RC fixes:
#725 
#726 
#728 
#737


karlthepagan:
You're missing #737 but that's a tiny edge case ;)

Issue #679, opened 2017-02-06
Resource Management


logxen:
~To prepare for labs our creeps need the ability to decide where resources should go (instead of just stuffing everything into the storage with overflow going to the terminal). This system will manage all minerals and compounds but probably not energy and power.~

~`Room.resources.setLoadingAmount(<structure>, <resource>, <amount>)`~

~This command populates a data object that stores what the player wants to happen with a room's resources.~

~`structure` is the lab, container, storage, or terminal to be controlled.~
~`resource` is the resource to set the target amount for.~
~`amount` is the amount of 'resource' that haulers will proceed to attempt to load into the structure.~

~To empty a structure of a resource set the amount to 0.~
~Running this command on a lab will set the loading amount to 0 for all other resources for that lab.~

~Hauler creeps will then attempt to fill structures to the requested loading amounts as part of action.reallocating. Structures will be filled in the following order of priority:~
~Lab -> Container -> Terminal -> Storage (Storage still overflows to the Terminal for auto-sell)~

~By default all structures are assumed to have loading amounts of 0. For the storage Structure any set loading amount is *added* to the savings amount set in parameters. As such, once this is in place everything should work as before until a setLoadingAmount command is actually executed.~

The info in this post is superseded by the comments below.

karlthepagan:
For spawning boosted creeps the requirements are slightly different, but this is good to get us started.

logxen:
Ok ... on further consideration I think it needs two functions. One that is a one-time order and another that is a continuous storage order. This would allow more careful manipulation of the resources. So:

room.placeOrder(<structure>, <resource>, <amount>)
room.setStore(<structure>, <resource>, <amount>)

This will extend the data structure a bit as well, changing `resourceType: targetAmount` to:
```javascript
    resourceType: {
        'orderAmount': orderAmount,
        'orderRemaining': orderRemaining,
        'storeAmount': storeAmount
    }
```

The resourceType.orderRemaining value would start equal to the total orderAmount but every time a hauler dropped off the resource it would subtract the amount from the orderRemaining. In this way the haulers know both how much of a one time order is left to be filled and they can look at the orderAmount plus the storeAmount and thereby would not immediately bring the resource back to storage.

To reset the value every time a hauler takes the resource from storage he should subtract that amount from orderRemaining. This will effectively let the one-time resources be used first.

logxen:
[...]

PR #739, opened 2017-02-14
Feature/resource management

logxen:
Ok I need more testing! Thanks a lot to @almaravarion for helping me with testing to get it this far. I made a clean transition from dev to this branch this morning and things went smoothly.

If you want to be able to move things around and are currently using dev please give this branch a try.

There are currently some docs here (just ignore the first post): https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/679

I will put together a wiki page once the system is generally accepted.

Cheers!

booooooostedupgrader


whardeman:
I would really love to try out your branch and test labs/boosting. Unfortunately, I'm trying to help finish testing RC3. Once that's done, I'll be checking this out and giving it a spin. My number one concern after trying canary lately is CPU, so if it doesn't drain my bucket, I'll really put it through its paces.

whardeman:
So, my only gripe right now: I started an order for G and while all the lab orders went into memory and the trading appears to be working, only ZK is being produce even though all the other labs are open. Also, I spawned an extra hauler to be sure that the room would be able to handle keeping extensions filled, etc but also move minerals as needed, yet the haulers refuse to bring U or L to their respective labs and instead, every 10 ticks they rush down to pick up 5 recently produced ZK and move it back to storage.

Dasku:
Just testing this out now, room has 6 labs, ordered 5000 GO (I think haha) - I used this
Game.rooms['E77N55'].placeReactionOrder('587d071e23457b202fbc986b', RESOURCE_GHODIUM, 5000);

Now a few orders have appeared, its started making ZK like with @whardeman. The room did have Z and K (200k and 100k respectively from robbing someone elses room), I don't have any U in any rooms so can't make GO (only just realised this while checking and writing this).

I have 1 master lab with order for G
1 master lab with order for ZK
2 slaves to ZK
2 idle (as no U can be found in any rooms I'm guessing)

If a mineral isnt available, should it be bought off the market? - or should it cancel that order so the lab can be used for something else instead?

Almaravarion:
why not make it configurable?
some may want to keep credits for later use, or shedule large lab order for stockpiling of goods later on

whardeman:
@Dasku If you go ahead and buy some U from the market, (do you also have L?) you can buy it directly into the room and then haulers _may_ eventually pick it up, but I had to go in and manually edit the Memory object once the order went into idle state. Not sure if that was required or if I'm just impatient.

Dasku:
I've emptied my bucket and had to go back to dev to fill back up. Working away tomorrow but will test more Friday and over the weekend.

logxen:
@Dasku it may have also just failed to find the needed slave labs ... the algorithm that finds the slave labs is pretty dumb right now. Basically, the master finds all labs within 2 ranged then takes the 2 that were built first. Those labs then do the same thing from where they are to put the first tier together. So ... right now that's pretty hit or miss.

parameter.js
```diff
@@ -5,6 +5,7 @@ let mod = {
     SAY_ASSIGNMENT: true, // say a symbol representing the assiged action
     SAY_PUBLIC: true, // creeps talk public
     DEBUG: true, // gimme some more details, use false not undefined to unset
+    DEBUG_LOGISTICS: false, // console log notes from hauler creeps running action.reallocating
```
karlthepagan: I intend to migrate these to debugTrace.

For example you would print these messages using `Memory.debugTrace.Action='reallocating'`.

logxen:
@karlthepagan FYI, I don't intend to keep those debug messages once action.reallocating is stable ... there's too much spam there and it's just haulers moving stuff.

karlthepagan:
I'd like to keep them, but I will move them into the trace syntax. There's less spam when you configure precisely what's displayed.

logxen:
Yeah ok. I can finish them then ... they are currently incomplete. I just need to study the trace system a bit.

PR #746, opened 2017-02-15
Fix undefined error

### LINKED ISSUE #721: RCL Displays Undefined in RCL8 Room
**Problem:** RCL Bar and progress next to controller display `undefined` when at RCL8.

**Expected Outcome:** Full bar with progress and level, and remove controller progress text (or just display current progress)

Spedwards:
Resolves #721

This commit also adds a feature allowing you to enable visuals for any room you have vision in that contains a controller.


Spedwards:
Need someone with an RCL8 room to make sure the RCL bar at top is correct, and the progress line next to the controller is removed.

Issue #722, opened 2017-02-13
Creep Paths display incorrectly when fatigued


Spedwards:
**Problem:** Creep paths display incorrectly when fatigued. An extra movement is drawn at the start.

**Expected Outcome:** Do not draw the extra movement.

PR #748, opened 2017-02-15
Fix for fatigued creeps.

Spedwards:
This doesn't fix creeps who stopped moving for reasons other than fatigue (i.e. hauler targeting extension). For that I need to add a check to see if it moves and add it with `if (creep.fatigue === 0)` so it won't change the `(x, y)` if it never moved.


Spedwards:
An extremely hacky fix would be to change all the `move` calls to set the creep's fatigue to `> 0` if it returned `OK`. Would be harmless, however I'm not sure if any other code uses the fatigue value.

Spedwards:
Forgot to mention, this (kind of) resolves #722

PR #751, opened 2017-02-15
Replace Creep.drive with Traveler.js

chobobobo:
After many frustrating hours trying to debug and optimize our routing code, I declare defeat.  

https://github.com/bonzaiferroni/bonzAI/wiki/Traveler-API is an existing free-to-use routing package that we can plug and play, and if we can get our routing CPU issues under control, we can focus on other, more important issues.

This branch is has been tested and should be free of major bugs.  However some routing choices may diverge from previous expectations.  Also, be aware that it's going to complain a lot about creep CPU usage, this is good because it identifies the big consumers.

Known Bugs
- ~~creeps cannot use public ramparts~~
- ~~creeps cannot path over hostile construction sites~~
-~~if the path is blocked in the last few steps, the creep will get stuck and not escape.~~
- ~~reservers will attempt to pathfind onto other reservers when they have arrived to replace them~~
- ~~dropping action will eat large amounts of CPU when a spawn in selected because it is trying to pathfind onto the spawn.~~
- ~~placing a remote mining flag in a wall breaks the creep pathing~~
- You will get incomplete path notifications when a creep is blocking the destination.
- ~~action.invade creeps will move towards their targets, but not attack.~~
- ~~currently it will avoid any path through a SK room and may take the long way around unless it has no choice~~ (this is as intended I believe)
- ~~currently if a worker starts building, but their ideal path to the constructionSite is through another room, they will border cycle.  The proposed fix is #766 and once that is approved I will merge it into this branch as well.~~
- ~~defenders seem to use a lot of cpu while traveling ~ 0.5-2CPU/tick~~

Untested
- ~~Avoiding rooms~~(set Memory.rooms.hostileRooms[roomName] = true)
- ~~Challenging geography~~
- ~~Routing through friendly rooms (likely not working at all yet)~~


karlthepagan:
We need to replace replacing `step` with `travelTo`

Spedwards:
I'm assuming new path visuals will need to be created for this?

chobobobo:
@spedwards Yes, I'll have a look at how you are currently doing them, but likely it will be as simple as reading the path from a different location in memory.

logxen:
Yeah I wanted to do that anyway to make it easier to make virals for the routeCallback

chobobobo:
@OverlordQ Yeah, it appears when I made that fix I checked it into the wrong branch.  I've found it and added it now, please review.

OverlordQ:
Looks like it needs a small rebase, likely conflicted with the hauler action changes.

OverlordQ:
meant to put this as a separate PR into this branch, but I stuffed it up, :face_with_head_bandage:

visuals.js
```diff
@@ -286,13 +286,13 @@ module.exports = class Visuals {
 	static drawCreepPath(room) {
 		const vis = new RoomVisual(room.name);
 		room.creeps.forEach(creep => {
-			if (creep.data && creep.data.path) {
-				const path = creep.data.path.substr(1);
+			if (creep.memory && creep.memory._travel) {
```
Spedwards: No need to check for `creep.memory`. It will always exist.

OverlordQ:
```
[11:07:03 PM]TRAVELER: incomplete path for hopper-4320-1
[11:07:03 PM]attempting path without findRoute was not successful
```

When creeps want to recycle, if they're in a hostile room, they can't path out of the hostile room.

OverlordQ:
This also breaks attackTrain

chobobobo:
@OverlordQ can you change the log output to                     
console.log(`TRAVELER: incomplete path for ${creep.name} from ${creep.pos} to ${destPos}`);
and let me know whta path it is trying to find?

chobobobo:
@OverlordQ also, since those are internal creeps, please create an issue on that repository and link to it from here

OverlordQ:
I thought about it, but they all use `Creep.action.travelling` to do the actual movement, so this will need to be fixed in the public code.

chobobobo:
Yes, you're quite right, my bad.  It's any military creep that needs to pathfind in a hostile room that is affected as well.

OverlordQ:
My attempt is #784, which will need merged before this PR is closed. Feel free to reject if you have a better idea.

chobobobo:
@OverlordQ see https://github.com/cyberblast/screeps.ocs.internal/pull/219 to go together with these changes, tested with a hopper and it worked as expected.

@karlthepagan could you review this to make sure I have correctly used strategies?  We will likely face further more minor bugs as we improve the granularity of who is allowed in a hostile room and when.

chobobobo:
Currently the invade action does not seem to be working, creeps move towards their targets but they are not attacking or healing.

traveler.js
```diff
@@ -0,0 +1,355 @@
...
+        if(!gOpts.installTraveler) {
+            global.traveler = new Traveler();
+            global.travelerTick = Game.time;
+        }
+
+        Creep.prototype.travelTo = function (destination, options = {}) {
+            if(global.traveler && global.travelerTick !== Game.time){
+                global.traveler = new Traveler();
+            }
+            options = this.getStrategyHandler([], 'moveOptions', options);
```
karlthepagan: Yup, this looks good.

creep.Action.js
```diff
@@ -102,7 +100,10 @@ let Action = function(actionName){
     this.onAssignment = (creep, target) => {};
     // empty default strategy
     this.defaultStrategy = {
-        name: `default-${actionName}`
+        name: `default-${actionName}`,
+        moveOptions: function(options) {
+            return options || {};
```
karlthepagan: 👍  this provides the default to all actions - which is the fundamental root for all strategy handlers.

The important requirement here is that handler functions must never return undefined (or it looks like the handler method does not exist).

karlthepagan:
I think the use of strategy handlers is correct. I'm still working thru the full details of this change.

karlthepagan:
Deleting flags is currently causing issues.

traveler.js
```diff
@@ -0,0 +1,356 @@
...
+                maxOps: options.maxOps,
+                roomCallback: callback
+            });
+        }
+        travelTo(creep, destination, options = {}) {
+            // register hostile rooms entered
+            let creepPos = creep.pos, destPos = (destination.pos || destination);
+            if (creep.room.controller) {
+                if (creep.room.controller.owner && !creep.room.controller.my &&
+                    !_.includes(PLAYER_WHITELIST, creep.room.controller.owner.toLowerCase())) {
```
karlthepagan: change to `creep.room.ally` - this also uses the reputation system

karlthepagan:
Really fine edge-case here: https://screeps.com/a/#!/history/E76S9?t=17566389

The hauler in the upper left is trying to get to the link, but it is actually a route 3 rooms long to get to the link. As a workaround I opened up the wall.

karlthepagan:
Privateer travels to SK room and then keeps trying to move into his flag: https://screeps.com/a/#!/history/E76S6?t=17567642

chobobobo:
@karlthepagan these changes should resolve the issues you are having.

karlthepagan:
```
36 PM] TypeError: Cannot read property 'controller' of undefined
    at Object.mod.run (creep.behaviour.privateer:23:55)
    at Creep.run (creep:50:39)
    at run (creep:338:30)
    at Function.mod.execute (creep:339:7)
```

haven't updated yet, but I caught this one as well

chobobobo:
Not at the computer now, but the likely fix for this is to use assignRoom rather than select a controller as a target.  Later this should look for towers in the room and try to go to them for a heal.

karlthepagan:
Interesting pathfinding choice - https://screeps.com/a/#!/history/E76S6?t=17568701 creeps in my SK room are pathing thru rooms without vision rather than directly south to a room which I own.

chobobobo:
This is currently not ready to be merged in with a few outstanding bugs.  The big thing to watch out for is any time we are trying to path to an unreachable location as that makes Traveler devour CPU (26-32CPU/computation)

karlthepagan:
Merged #801 into this.

karlthepagan:
Merged #714 into this

karlthepagan:
Merged #803 into this

Spedwards:
When first plugging in this branch:

```
TypeError: Cannot read property 'toUpperCase' of undefined
    at Traveler.findTravelPath (traveler:94:26)
    at Traveler.travelTo (traveler:223:32)
    at Creep.travelTo (traveler:348:29)
    at Action.action.step (creep.action.travelling:24:15)
    at Object.mod.run (creep.behaviour.remoteHauler:13:22)
    at Creep.run (creep:47:39)
```

karlthepagan:
58ad9747d0fffb4a1aa935df - creep stuck on border here https://screeps.com/a/#!/room/E76S10

it looks like he moved onto the border tile to pick up the energy & resources that an invader was carrying and now he's stuck trying to move TOP_RIGHT but that's a wall by the time he executes the move

karlthepagan:
merged #811

karlthepagan:
Of course a squash merge is how this feature will be integrated.

karlthepagan:
Are my own rampart construction sites being counted as obstacles?

image

My creeps are pathing around them.

karlthepagan:
My creeps are randomly deciding to path thru forkmantis's room https://screeps.com/a/#!/room/E77S7

That path is blocked by walls which I don't have vision to.

As well they are pathing into engineeryo's SK room to my west where they are blocked by walls that I have vision to.

I have set the route cost for these rooms to Infinity and haven't seen it happen again yet.

chobobobo:
Early RCL workers seem to eat a lot of CPU as well ~1200+ for their lifetime.

karlthepagan:
Room cost: Infinity has not resolved the route finding issue. I still have a few creeps who wander into forkmantis's room and get caught and go into an incomplete route cycle.

chobobobo:
@karlthepagan there is an efficiency issue with the avoid check, it consumes about 1CPU/tick on removeMiners, not sure why.  Also, I have commented out the code in traveler that rebuilds paths when it "thinks" the creep has moved.  There are likely edge cases that I have broken by this change, but it keeps CPU manageble for the time being.  Can you implement a check to see if the creep actually moved.

task.mining.js
```diff
@@ -349,6 +349,13 @@ mod.creep = {
         behaviour: 'remoteMiner',
         queue: 'Medium' // not much point in hauling or working without a miner, and they're a cheap spawn.
     },
+    skMiner: {
```
karlthepagan: This is way out of scope of the traveler PR and has contaminated the release process. There are now too many conflicts and I am being forced to remove traveler from canary.

As an alternative I am going to move this change into a new PR and hard reset traveler to the previous change.

karlthepagan: I created `major/traveler-pure` for this purpose.

karlthepagan:
There's a few places that we travel to 25,25 in a room without a travelTrange.

This results in incomplete path errors (wasted CPU) like this: `TRAVELER: incomplete path for remoteHauler-zetaRMine-2 from [room E82S11 pos 40,25] to [room E83S11 pos 25,25]. Route length 1.`

chobobobo:
@karlthepagan are you sure that's a result of travelRange not being set?  That error means that there is 1 room in the route length doesn't it, not a range of 1.

chobobobo:
Noticing that occasionally remoteHaulers will try to idle near their source when they are empty, but they also want to stay within a certain range (3 I think) so they idle off the road to 4 range and then move back into 3 range causing excessive CPU usage.

Dasku:
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/863

traveler.js
```diff
@@ -0,0 +1,374 @@
...
+            if (!this.structureMatrixCache[room.name]) {
+                let matrix = new PathFinder.CostMatrix();
+                this.structureMatrixCache[room.name] = Traveler.addStructuresToMatrix(room, matrix, 1);
+            }
+            return this.structureMatrixCache[room.name];
+        }
+        static addStructuresToMatrix(room, matrix, roadCost) {
+            for (let structure of room.find(FIND_STRUCTURES)) {
+                if (structure instanceof StructureRampart) {
+                    if (!structure.my) {
```
pandabear41: Need to change to. So creeps can go through public ramparts.

` if (!structure.my && !structure.isPublic ) {`

traveler.js
```diff
@@ -0,0 +1,374 @@
...
+            for (let site of room.find(FIND_CONSTRUCTION_SITES)) {
+                if (site.structureType === STRUCTURE_CONTAINER) {
+                    continue;
+                } else if (site.structureType === STRUCTURE_ROAD) {
+                    if (USE_UNBUILT_ROADS) matrix.set(site.pos.x, site.pos.y, roadCost);
+                    continue;
+                } else if (site.structureType === STRUCTURE_RAMPART) {
+                    continue;
+                }
+                matrix.set(site.pos.x, site.pos.y, 0xff);
```
pandabear41: Why are you setting construction site weight to 0xff? They should be the same weight as the tile itself or possibly lower weight if enemy construction sites, so creeps will destroy. 
Creeps can walk over their own construction sites and if they are enemy sites creeps will destroy them by moving over it.

It should be 
```
if (site.structureType === STRUCTURE_ROAD) {
	if (USE_UNBUILT_ROADS) matrix.set(site.pos.x, site.pos.y, roadCost);
    continue;
}
```
or (creep that prioritizes construction site trolling)
```
if (site.structureType === STRUCTURE_ROAD) {
	if (USE_UNBUILT_ROADS) matrix.set(site.pos.x, site.pos.y, roadCost);
} else if (!site.my) {
        // Kill enemy sites ;)
	matrix.set(site.pos.x, site.pos.y, roadCost);
}
```

Any input or objections?

karlthepagan: Can you test this with manual moves on top of a wall or extension site?

In my experience only walkable structures have walkable sites. Enemy cs's get bulldozed regardless tho. Maybe if we see a reputation.notHostile cs we should try to avoid it? (Cost 20 or higher?)

karlthepagan: Can you test this with manual moves on top of a wall or extension site?

In my experience only walkable structures have walkable sites. Enemy cs's get bulldozed regardless tho. Maybe if we see a reputation.notHostile cs we should try to avoid it? (Cost 20 or higher?)

traveler.js
chobobobo: This makes sense to me, however do we want all creeps using public ramparts by default, or would this be better to have a config setting for?

Dasku:
This may not be relevant here, however, in the picture below creeps are choosing a path through rooms (this isnt using traveller) (mine + enemy rooms). However, could we make pathing prioritise Highway Rooms? (there should never be enemies in highway rooms?) If you guys haven't addressed this already? (thinking more for powerMining - finding the shortest path to highway rooms should lead to the safest journey) 

image

traveler.js
pandabear41: You are correct. With testing you can only move over your own walkable construction sites (road, rampart, container). 

Sounds like the reputation would be the best bet for hostile construction sites.

chobobobo:
@Dasku I was running with highway rooms as the default, but it was causing some of my creeps to ignore my own rooms with roads in them in favour of highways without roads.  When your creeps first encounter a hostile room (or you manually set it) they will stop pathing through that room, so if they know not to use those rooms they will be finding routes that avoid hostile territory.

However, for power mining, you can just use creep.travelTo and set options.preferHighwayRooms(I think it's called that, check the code)

traveler.js
pandabear41: You could use a config setting, but you could use the reputation task instead and only allow if `notHostile` or `isAlly`?
Which do you think would work best?

chobobobo: I hadn't actually thought of the case of allies, reputation is the way to go I feel.

chobobobo:
In the avoid action, sometimes another creep will block a harvest spot( the only one at times) determinated spots on a source should be added to the spots to avoid when idling.

chobobobo:
idle creeps need to idle somewhere away from borders
screen shot 2017-03-01 at 8 02 09 pm

Spedwards:
@chobobobo Here's a lazy method I used in my old codebase to move idle creeps off roads. Could be easily modified to get them away from borders too.

```
Creep.prototype.isOnRoad = function() {
    return _.find(this.pos.lookFor(LOOK_STRUCTURES), i => i instanceof StructureRoad) !== undefined;
}

Creep.prototype.moveToIdlePosition = function() {
    if (this.isOnRoad()) {
        return this.move(Math.floor(Math.random() * 9)) === OK;
    }
    return true; // already off the roads
}

chobobobo:
There's a bug in using move(Direction) for the last few steps in creep.action.step we need to check for stuck and use traveler if stuck

chobobobo:
Also, not sure when it happened but room.routeCallback's second parameter seems to have changed from allowHostile to checkOwner, which changes the logic of that completely.  This explains why my creeps have been routing through enemy rooms.

chobobobo:
I'm also going to need to override the costmatrix callbacks using options in order to cache matrices like we currently do, provide access to public ramparts in the right cases, and path over hostile construction sites.

chobobobo:
9	
TypeError: Cannot read property 'structureType' of undefined
    at invalidObject (room:2175:128)
    at Array.filter (native)
    at Room.isWalkable (room:2177:21)
    at Action.step (creep.Action:71:32)
    at Object.mod.run (creep.behaviour.remoteWorker:17:22)
    at Creep.run (creep:64:39)
    at run (creep:358:30)

related to the most recent change, is it trying to check walkable when it's on a border perhaps?

karlthepagan:
`TRAVELER: incomplete path for guard-Flag4-1 from [room E83S11 pos 46,16] to [room E84S11 pos 13,22]. Route length undefined.`

Going from E83S11 to E84S11 there should be 4 rooms allowed and the naive pathfinding might to thru the wall at first. The fallback should be to go from the current room to route[1] which should be E83S11 to E84S10, but it looks like it just fails here.

I need to update to the latest canary and make sure my precalc is being obeyed.

karlthepagan:
I think a callback for permission to path may be needed for CPU management.

This could cover the stuck count as well as provide a much greater buffer for critical CPU.

karlthepagan:
@pandabear41 please respond to changes by approving them

chobobobo:
We also need to consider border-cycling as "stuck"

karlthepagan:
I need to remember to ignore the cost of a room in the findRoute callback if it is the destination room.

Perhaps also print warnings that list infinite room costs if a route can't be found.

karlthepagan:
https://screeps.com/a/#!/room/E85S12

Example of a room that's currently breaking me because I chose route precalculation.

findRoute believes that there is a valid exit to the south but the reality is that this produces an illegal route because it is blocked.

For this reason it could be justified to keep route precalc out of the recommendations and instead only use it for troubleshooting and for the fallback case.

chobobobo:
It's possible that the changes I made to room.costMatrix have negatively affected the fleeMove action, this needs further investigation.

chobobobo:
I think what may have happened is that the previously generated cost matrices were still being used and had not been regenerated, causing pathing issues, `delete Memory.pathfinder` will remove the cache and should resolve the issues you are seeing.  I will look at a way to migrate cleanly from the previous cache.

chobobobo:
I believe this is ready to merge now, @karlthepagan could you review my last commit related to border cycling for sanity?  Also, I believe your recent comments on this thread were more for your own reference, but I am not clear on your idea for a callback for permission to path.  I'm also not completely clear why E83S11 - E84S11 failed, were the highways not being considered?  It seems like it should have been a very simple route.

karlthepagan:
Looks good. @logxen take one more look and let's do tha thang

OverlordQ:
Something is broken in latest canary, guards wont enter hostile rooms.

chobobobo:
I will try to reproduce, this does demonstrate that we could do with better TRACE information in this code though, it would be helpful for us to be able to determine what action caused the unexpected move as well as where specifically the pathfinding broke down.

chobobobo:
@OverlordQ unfortunately I haven't been able to reproduce this, are you sure your internal branch is up to date, and Memory.modules is correct?  I created a guard flag, and used red/red inside the room, the guard moved inside the room and started attacking until it was destroyed.  I wonder if it's possible the bug is specific to targeting either a creep or a building and mine targeted the other.

karlthepagan:
@logxen 12 hour warning before I dismiss your hold. Let us know if you need more time.

chobobobo:
I believe there is a bug that may be what was hitting @OverlordQ if a creep targets a flag in a hostile room it returns no path and eats a very, very large amount of CPU (likely 20000 ops / tick)

chobobobo:
ok, potential fix for the issue @OverlordQ was having is in now, @karlthepagan it's a bugfix we will likely want to push back to the main traveler repo https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/751/commits/95a9b60e1f6ee0a03e9a1d99e5275a71d8d200a6

PR #755, opened 2017-02-16
Fill up worker from containers when picking

OverlordQ:
I've noticed things picking up dropped energy waste ***a lot*** of time making back and forth trips to pick up a partial load of dropped energy/minerals. Then, instead of grabbing from the container to fill up and make room for more to be stored, they wander off, leaving the miners to have to drop more energy on the ground.

### DIFF
--- a/creep.action.picking.js
+++ b/creep.action.picking.js
@@ -53,10 +53,10 @@ action.work = function(creep){
                 return result;
             }
         }
-        // Check for containers to uncharge also
+        // Check for containers to uncharge
         if( creep.sum < creep.carryCapacity) {
-            let containers = creep.pos.findInRange(creep.room.structures.container, 1, {
-               filter: (o) => Creep.action.uncharging.isAddableTarget(o, creep)
+            let containers = creep.pos.findInRange(creep.room.structures.container.in, 2, {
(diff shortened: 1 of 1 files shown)


creep.action.picking.js
```diff
@@ -53,6 +53,16 @@ action.work = function(creep){
                 return result;
             }
         }
+        // Check for containers to uncharge also
+        if( creep.sum < creep.carryCapacity) {
+            let containers = creep.pos.findInRange(creep.room.structures.container, 1, {
+               filter: (o) => Creep.action.uncharging.isAddableTarget(o, creep)
```
karlthepagan: right now uncharging.isAddableTarget is always true, check isValidTarget instead

creep.action.picking.js
```diff
@@ -53,6 +53,16 @@ action.work = function(creep){
                 return result;
             }
         }
+        // Check for containers to uncharge also
+        if( creep.sum < creep.carryCapacity) {
+            let containers = creep.pos.findInRange(creep.room.structures.container, 1, {
```
karlthepagan: A range of 2 is probably ok. Any further and I would want to use findClosestByPath.

karlthepagan: This one is optional. Use your judgement.

PR #757, opened 2017-02-16
standard .step now works and then finds a new target and moves in the…

logxen:
… same tick

Seriously kids, we are talking about close to doubling the speed of haulers filling extensions here.


creep.action.feeding.js
```diff
@@ -25,7 +25,12 @@ action.newTarget = function(creep){
     });
 };
 action.work = function(creep){
-    return creep.transfer(creep.target, RESOURCE_ENERGY);
+    let result = creep.transfer(creep.target, RESOURCE_ENERGY);
+    if (result == OK) {
+        creep.target = null;
+        this.assign(creep);
```
chobobobo: Has this been tested to ensure that it's not assigning the same target (the transfer hasn't completed yet, so it would still be empty right?)

creep.Action.js
```diff
@@ -61,8 +61,10 @@ let Action = function(actionName){
                 return;
             }
         }
-        if( creep.target )
+        if( creep.target ) {
+            range = creep.pos.getRangeTo(creep.target);
```
chobobobo: range is being regenerated because of the possibility that target has changed in creep.work correct?

logxen:
It has been tested and gets a new target but that's a good point about the old target not having actually been filled yet during the tick. Deserves a closer look.

And yes. Range needs to be recalculated because it may have changed.

creep.Action.js
logxen: Yep

creep.action.feeding.js
logxen: I looked at the code and I think it works reliably because the newTarget function in action.feeding uses isAddableTarget which already excludes the currently selected target with the targetof feature.

creep.Action.js
karlthepagan: range is cheap

logxen: sqrt and two pow

creep.action.feeding.js
```diff
@@ -25,7 +25,12 @@ action.newTarget = function(creep){
     });
 };
 action.work = function(creep){
-    return creep.transfer(creep.target, RESOURCE_ENERGY);
+    let result = creep.transfer(creep.target, RESOURCE_ENERGY);
+    if (result == OK && creep.carry.energy > creep.target.energyCapacity-creep.target.energy) {
```
karlthepagan: I'm going to think about this - could it be reconciled with is .. Valid/Addable .. Action/Target ?

karlthepagan:
Going to get this into canary for soak before I merge.

PR #760, opened 2017-02-16
MASTER avoid source crowding

karlthepagan:
this resolves the case where users manually assign miners to another action (yet they keep mining!)

also may fix other issues, but it's a pretty fine edge case.


OverlordQ:
Would address 756

PR #762, opened 2017-02-16
Fix upgrader determinatedSpot selection

chobobobo:
As well as forgetting previous bad spot selection and upgrading as soon as they are in range.  One possible change from old code is the potential range to a source, this checks to ideally choose a spot that is not adjacent to a source if possible.

We could add withdrawing from storage and storage to spot selection as well if we want to.

### DIFF
--- a/creep.behaviour.upgrader.js
+++ b/creep.behaviour.upgrader.js
@@ -45,12 +45,18 @@ mod.run = function(creep) {
             };
             let linkSpots = creep.room.structures.links.controller ? _.flatten(_.map(creep.room.structures.links.controller, getSpots)) : [];
             let containerSpots = creep.room.structures.container.controller ? _.flatten(_.map(creep.room.structures.container.controller, getSpots)) : [];
-            // priority = close to both > close to link only > close to container only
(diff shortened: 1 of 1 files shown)


chobobobo:
This seems to be working now, the only forseeable problem I see is that it is fairly naive about the position, so it will try to stand next to a link, failing that next to a container.  It could be better if it first searched for free positions next to both a container and a link, but that adds complexity.

jisaf:
Testing this in https://screeps.com/a/#!/room/E2N82 where the current logic failed to correctly place a 3rd upgrader.

chobobobo:
@jisaf I have added support for terminals and storage as well but it needs testing.  Can you let me know if that update allows your upgrader to withdraw from storage as intended?  @karlthepagan do you think there should be a check for min storage level before we allow them to draw from storage?

karlthepagan:
@chobobobo yes, do not consider the storage when it is below MIN_STORAGE_ENERGY[room.controller.level]

also do not consider the terminal when it is below 0.5 * TERMINAL_ENERGY

PR #766, opened 2017-02-17
Fix build action when transiting through room

chobobobo:
If a creep selects a build site, but paths through an external room to get there it loses its action because that room does not have any construction sites. If we want to prevent the action from being added when there are no construction sites, perhaps the check for constructionSites should be in the isAddableAction instead?

### DIFF
--- a/creep.action.building.js
+++ b/creep.action.building.js
@@ -4,7 +4,7 @@ action.maxPerTarget = 3;
 action.targetRange = 3;
 action.maxPerAction = 3;
 action.isValidAction = function(creep){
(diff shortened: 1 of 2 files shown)


karlthepagan:
Whoever merges this please squash merge. Reverted merge commits can be problematic.

chobobobo:
@karlthepagan would it be simplest for me to just reset fix/buildTravelling to the earlier commit, or are there any unforeseen side effects of doing that?

karlthepagan:
Yes that is ok. There's no problem reverting and force pushing branches if you're the only one working a branch.

it makes some people nervous tho, so I understand the hesitation and I don't push people into it.

PR #771, opened 2017-02-17
Stop trying to harvest if source has no free spaces

chobobobo:
Just add a check to validTarget to make sure the source still has enough spaces to harvest.  Prevents heavy CPU usage when trying to path to an unreachable source.

### DIFF
--- a/creep.action.harvesting.js
+++ b/creep.action.harvesting.js
@@ -6,10 +6,12 @@ action.isValidAction = function(creep){
 };
 action.isValidTarget = function(target) {
     return (target !== null && target.energy !== null && target.energy > 0 &&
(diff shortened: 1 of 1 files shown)


OverlordQ:
isn't this duplicating `isAddableTarget`?

chobobobo:
@OverlordQ Yes, but isAddable is only called when adding the target.  If for some reason the target has been added in error the creep won't ever release the target.

OverlordQ:
But since `action.renewTarget` is false, wont it still stick to the target?

chobobobo:
Had to look closer at the code to understand why, renewTarget refers to whether it should select a new target or not.  If the target comes up invalid, the creep will still abandon the action, it just won't select a new one immediately.

PR #779, opened 2017-02-18
room died creep is too big

karlthepagan:
I had two rooms die from the handleRoomDied creep.

The simplest fix is to make that this backup creep is only a 200 energy creep.


karlthepagan:
#783 helps address the customizability of room haulers ;)

PR #783, opened 2017-02-19
upgrader, miner, hauler maxCount dereference so that they are customizable

karlthepagan:
i figured out why I was having trouble customizing upgrader counts


karlthepagan:
example viral I use: viral.creep.setup.upgrader.js

```
const setup = {};
module.exports = setup;
setup.maxCount = function(room){
    let count = this.baseOf.viral.maxCount(room);
    switch( room.name ) {
        case 'E71S9': return count + 2;
        default: return count;
    }
};
```

PR #788, opened 2017-02-19
storage charge property, returns percent of the "full" range

karlthepagan:
The full range represents when energy between min and max.

use cases:
* when you are over max energy storage: room.storage.charge > 1
* when you are under minimum energy storage: room.storage.charge < 0

Numerical computing - this is biased around 100% full, the difference between current and max is the most important comparison in many existing methods.

If the room minimum === maximum then the response is either `Infinity` or `-Infinity`.

### DIFF
--- a/extensions.js
+++ b/extensions.js
@@ -177,6 +177,24 @@ mod.extend = function(){
             return this._sum;
         }
     });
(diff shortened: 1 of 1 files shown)


karlthepagan:
I intend to replace all references to room storage min/max with this charge property.

Eventually we can do the same thing with managed containers.

Issue #719, opened 2017-02-12
Improved Spawn Queue


Spedwards:
At the moment, this is just for me to work from so I don't forget anything.

---

 - Each setup will have it's own base value.
 - Each behaviour will have a base modifier.
 - Other modifiers will be considered when adding to the queue:
    - Spawn time
    - Distance from dest.
    - Remote or Home
 - Creep in queue with the lowest end value will get spawned next.
 - TBD: Re-evaluate queue every _x_ ticks to prevent useless creeps being spawned

Spedwards:
If you have an suggestions, criticism, etc. please feel free to share.

chobobobo:
Also see #620 this may be related

chobobobo:
@Spedwards I'm all for simplifying spawning and making it more predictable, but I am somewhat concerned about pruning every x ticks as that will reduce the likelihood of useless creeps, but not eliminate them.  Would it perhaps work if spawns uses event callbacks to poll receivers of spawns eg. flags/rooms and they reply with the information you need to determine the best next spawn?  That way we know a spawn is not being started that is not currently needed, and it would allow the receivers to be more deterministic about what is being spawned, and when to expect it.

karlthepagan:
Setup queues should be considered deprecated. I did not understand why until recently. #620 is on hold.

See #343

PR #790, opened 2017-02-20
Improved Spawn Queue

Spedwards:
Fulfils #719. This is just the base class. Still needs to replace the current system.

### DIFF
--- a/queue.js
+++ b/queue.js
@@ -1,0 +1,40 @@
+class Queue {
+	
+	constructor(room) {
+		if (typeof room === 'string') {
+			room = Game.rooms[room];
+		}
+		if (!(room instanceof Room)) throw new Error('Not a room');
+		this.room = room;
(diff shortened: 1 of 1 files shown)


karlthepagan:
There are time-sensitive spawns which need to be scheduled and I'm considering having those spawn at highest priority at a scheduled time (edited)

after that defensive spawns, after that new in-room mining and hauling, then in-room building and remote mining, then reserving, then remote hauling, then everything else (edited)

existing in-room mining and the last in-room hauler will be scheduled, maybe others

important factor being, that nothing spawned can overlap a scheduled higher priority item

karlthepagan:
maybe scheduled items should spawn early whenever there will be an overlap

karlthepagan:
Almost put this on hold, but I guess that indicates de-prioritization.

This is a priority, but it's still very early.

PR #791, opened 2017-02-20
Feature/critical cpu

logxen:
This patch adds CRITICAL_BUCKET_LEVEL and CRITICAL_ROLE parameters. If the bucket goes below the CRITICAL_BUCKET_LEVEL the system will stop processing creeps except for those roles listed in CRITICAL_ROLES.

In this way the bucket can be held above the CRITICAL_BUCKET_LEVEL in most circumstances by allowing the user to effectively choose which creeps to skip and allows all non-creep room processes (such as towers) to continue running every tick. Contrast this behaviour with a bucket that is actually empty which causes random parts of the colony not to function.

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -42,6 +42,9 @@ mod.extend = function(){
             if(!behaviour && this.data && this.data.creepType) {
                 behaviour = Creep.behaviour[this.data.creepType];
             }
+            if ( Game.cpu.bucket < CRITICAL_BUCKET_LEVEL && !CRITICAL_ROLES.includes(behaviour) ) {
+                return;
+            }
             this.repairNearby();
             if( DEBUG && TRACE ) trace('Creep', {creepName:this.name, pos:this.pos, Behaviour: behaviour && behaviour.name, Creep:'run'});
             if( behaviour ) behaviour.run(this);
--- a/creep.js
+++ b/creep.js
@@ -42,7 +42,7 @@ mod.extend = function(){
             if(!behaviour && this.data && this.data.creepType) {
                 behaviour = Creep.behaviour[this.data.creepType];
             }
-            if ( Game.cpu.bucket < CRITICAL_BUCKET_LEVEL && !CRITICAL_ROLES.includes(behaviour) ) {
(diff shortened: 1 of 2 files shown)


creep.js
```diff
@@ -42,6 +42,9 @@ mod.extend = function(){
             if(!behaviour && this.data && this.data.creepType) {
                 behaviour = Creep.behaviour[this.data.creepType];
             }
+            if ( Game.cpu.bucket < CRITICAL_BUCKET_LEVEL && !CRITICAL_ROLES.includes(behaviour) ) {
```
karlthepagan: you're comparing a string to an object here

logxen:
Heh whoops ... last minute change to the parameterized style. :)

parameter.js
```diff
@@ -129,5 +129,7 @@ let mod = {
...
     // Don't attack. Must be a member of OCS for permanent whitelisting in git repository. But you can change your own copy... Please ask if you are interested in joining OCS :)
     DEFENSE_BLACKLIST: [], // Don't defend those rooms (add room names). Blocks spawning via defense task (will not prevent offensive actions at all)
+    CRITICAL_BUCKET_LEVEL: 1000, // take action when the bucket drops below this value to prevent the bucket from actually running out
+    CRITICAL_ROLES: [ 'melee', 'ranger', 'healer', 'hauler', 'upgrader' ], // when the bucket drops below the critical bucket level only these creep roles will be executed
```
Spedwards: Not a requirement, but it may be worth adding `miner` to the list, otherwise people may end up running out of energy without realising what's happening.

logxen:
In most cases I've used it in so far it only happens for about 1 or 2 ticks at a time ... skipping all the creeps tends to get one out of the cpu deficit pretty quickly. Doesn't mean the code won't dip you right back in ... but in general all the skipped creeps still get to run at least every other tick.

I would like to discuss the list of CRITICAL_ROLES with everyone. We might for example want some of the internal creeps in an internalViral.parameter.js ...

parameter.js
logxen: yeah that's a good idea. the core local economy should run. it's generally remote stuff that is killing the cpu.

karlthepagan:
An email notification option is something to consider here.

Spedwards:
I'd disagree with an email option. At least by default. It's the type of thing that will spam you.

PR #795, opened 2017-02-21
storage charge

karlthepagan:
still need to test this, but it's simple tho there are many many cases so have at reviewing this

The purpose of this is to normalize how we address energy storage as a relative number. This hopefully makes our logic more comprehensible and consistent.

"charge" is just the percent of "full" - so 200% full means "i've got too much energy and need to spend it now!"

### DIFF
--- a/creep.setup.hauler.js
+++ b/creep.setup.hauler.js
@@ -20,7 +20,7 @@ setup.maxCount = function(room){
     if( miners > 0  || ( cont > 0 && workers > 2 )) {
         count += Creep.setup.upgrader.maxCount(room);
         if( room.structures.links.all.length < 3 ||
-           (room.storage && room.storage.store.energy > MAX_STORAGE_ENERGY[room.controller.level] &&
+           (room.storage && room.storage.store.charge > 1 &&
             room.structures.container.controller && _.sum(room.structures.container.controller, 'store.energy') == 0 )) count++;
         /* Add hauler when there is energy on the ground
         let dropped = 0;
--- a/creep.setup.hauler.js
+++ b/creep.setup.hauler.js
@@ -20,7 +20,7 @@ setup.maxCount = function(room){
     if( miners > 0  || ( cont > 0 && workers > 2 )) {
         count += Creep.setup.upgrader.maxCount(room);
         if( room.structures.links.all.length < 3 ||
-           (room.storage && room.storage.store.charge > 1 &&
+           (room.storage && room.storage.charge > 1 &&
             room.structures.container.controller && _.sum(room.structures.container.controller, 'store.energy') == 0 )) count++;
         /* Add hauler when there is energy on the ground
         let dropped = 0;
--- a/creep.setup.upgrader.js
+++ b/creep.setup.upgrader.js
@@ -3,11 +3,12 @@ module.exports = setup;
 setup.minControllerLevel = 2;
 setup.maxMulti = function(room){
     let multi = 0;
-    if( !room.storage || room.storage.store.energy > MIN_STORAGE_ENERGY[room.controller.level])
+    const charge = room.storage.store.charge;
+    if( !room.storage || charge > 0)
         multi++;
-    if( !room.storage || room.storage.store.energy > ((MAX_STORAGE_ENERGY[room.controller.level]-MIN_STORAGE_ENERGY[room.controller.level])/2)+MIN_STORAGE_ENERGY[room.controller.level])
+    if( !room.storage || charge > 0.5)
         multi++;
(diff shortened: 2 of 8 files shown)


logxen:
Hmmmm ... maybe the charge property or something similar would be a good way of implementing that part of action.reallocating.

Can we make a similar object that has a value for each resource type instead of just energy? Then that could be implemented for each of the different structure types that hold thins with a similar api interface.

creep.setup.hauler.js
```diff
@@ -20,7 +20,7 @@ setup.maxCount = function(room){
     if( miners > 0  || ( cont > 0 && workers > 2 )) {
         count += Creep.setup.upgrader.maxCount(room);
         if( room.structures.links.all.length < 3 ||
-           (room.storage && room.storage.store.energy > MAX_STORAGE_ENERGY[room.controller.level] &&
+           (room.storage && room.storage.store.charge > 1 &&
```
Spedwards: Should be `room.storage.charge`.

creep.setup.upgrader.js
```diff
@@ -3,11 +3,12 @@ module.exports = setup;
 setup.minControllerLevel = 2;
 setup.maxMulti = function(room){
     let multi = 0;
-    if( !room.storage || room.storage.store.energy > MIN_STORAGE_ENERGY[room.controller.level])
+    const charge = room.storage.store.charge;
```
Spedwards: Should be `room.storage.charge`.

creep.setup.worker.js
```diff
@@ -8,10 +8,10 @@ setup.maxWorker = room => {
         return 1;
     // constructionsites present & no strorage or storage > min
     if( room.constructionSites.length > 0 && (!room.storage
-        || room.storage.store && room.storage.store.energy > MIN_STORAGE_ENERGY[room.controller.level]))
+        || room.storage.store && room.storage.store.charge > 0))
         return 1;
     // storage full & base fortifyable
-    if( room.storage && room.storage.store.energy > MAX_STORAGE_ENERGY[room.controller.level] && room.structures.fortifyable.length > 0 )
+    if( room.storage && room.storage.store.charge > 1 && room.structures.fortifyable.length > 0 )
```
Spedwards: Should be `room.storage.charge`.

room.js
```diff
@@ -649,7 +649,7 @@ mod.extend = function(){
         'conserveForDefense': {
             configurable: true,
             get: function () {
-                return (this.my && this.storage && this.storage.store.energy < MIN_STORAGE_ENERGY[this.controller.level]);
+                return (this.my && this.storage && this.storage.store.charge < 0);
```
Spedwards: Should be `room.storage.charge`.

room.js
```diff
@@ -1396,7 +1396,7 @@ mod.extend = function(){
             }
         }
         if( this.controller.level == 8 && !transacting &&
-            this.storage.store.energy > MAX_STORAGE_ENERGY[this.controller.level] * 0.8 &&
+            this.storage.store.charge > 0.8 &&
```
Spedwards: Should be `room.storage.charge`.

karlthepagan:
@logxen how about something like `object.getCharge(RESOURCE_GHODIUM)`

would this interact with the resource management / orders?

logxen:
yes, both sides of the system could use this and it would be much cleaner than what I've currently got

logxen:
the other neat trick would be to also have it manage hysteresis. so would I would like is getNeeds(<resourceType>) which returns 0 when there's nothing to do, negative value when needs emptying, and positive value when needs filling. to do hysteresis that would also need to somehow track if the container is being filled or emptied ... and it would return 0 until the threshold was crossed.

karlthepagan:
Do you mean that we should track the current material flow as it relates to sources and sinks? E.g. a container attains a negative value as soon as a nearby source has a miner allocated. Then as haulers are allocated the flow / hysteresis score balances.

logxen:
no just for each container and resource that would be getting filled or emptied ... this concept doesn't work as well for the intermediate containers in the energy flow that are generally dealt with in charge and discharge if I'm not mistaken. this would be for back end reallocating.

creep.setup.upgrader.js
```diff
@@ -3,11 +3,12 @@ module.exports = setup;
 setup.minControllerLevel = 2;
 setup.maxMulti = function(room){
     let multi = 0;
-    if( !room.storage || room.storage.store.energy > MIN_STORAGE_ENERGY[room.controller.level])
+    const charge = room.storage.charge;
```
chobobobo: This needs an undefined check

Spedwards:
@chobobobo Want to approve this, since the changes were made?

PR #796, opened 2017-02-21
changing the mineralMiner's body

Perrytheplatypus43:
changing creep fixedBody and multiBody on line 17 and 18.
 
the old fixedBody is WORK, WORK, WORK, CARRY, MOVE 
the new fixedBody is MOVE, MOVE, WORK, WORK, WORK, CARRY

the old multiBody is WORK, WORK, WORK, MOVE
the new multiBody is MOVE, MOVE, WORK, WORK, WORK, WORK, CARRY


Spedwards:
Can you provide any reasons for this? Why would you want more than 1 `CARRY` part?

Perrytheplatypus43:
good point i changed it to 1 carry

Dasku:
Why so much move? It only needs to move from spawn to mineral 1 time? (on roads too so 1/3 move max, but thats still a waste?)

Dasku:
I think long term the change here needs to be a calculated body size based on amount of mineral left - and then change role to upgrader or recyle if any life left when mineral depletes.

creep.setup.mineralMiner.js
```diff
@@ -14,8 +14,8 @@ setup.maxCount = function(room){
     return max;
 };
 setup.default = {
-    fixedBody: [WORK, WORK, WORK, CARRY, MOVE],
-    multiBody: [WORK, WORK, WORK, MOVE],
+    fixedBody: [MOVE, MOVE, MOVE, WORK, WORK, WORK, WORK, CARRY],
+    multiBody: [MOVE, MOVE, MOVE, WORK, WORK, WORK, WORK, CARRY],
     minAbsEnergyAvailable: 750,
```
karlthepagan: the previous configuration has the effect of minMulti: 1

Perrytheplatypus43:
i changed it so the fixedBody and the multiBody both have 2 moves

Dasku:
@Perrytheplatypus43 Hi. Thank you for contributing. I think I see why you are wanting to change this, however as Karl suggested I think you may want to change this in your personal setup using viral settings. I don't think it would benefit the "general user" and therefore wouldn't be changed in the standard codebase. 

For background: MOVE parts only really matter when moving long distances. You need 1/2 a body of move parts to move every tick without roads (not including swamps). On a road you only need 1/3 body as move to move each tick. MineralMiners however don't need to walk very often, only once in fact from spawn to the mineral. Therefore, missing 5 ticks in travel time (from fatigue, not having enough MOVE) from spawn to mineral acctually saves a lot of wasted energy spent on un-needed move parts. This discussion did take place at the time we introduced mineral miners and general agreement was don't waste move parts. 

I will close this PR for now, but please dont hestitate to raise any new PR's you may have or want to submit. We can always discuss them before anything gets committed so no hard done in suggesting changes. 

Thanks again.

Dasku:
p.s. Miners and mineral miners also only require 1 CARRY part to enable them to put the energy in a container instead of drop mining.

PR #797, opened 2017-02-21
Save RCL8 Structures

Spedwards:
Add Room properties for observer, nuker, and power spawn.

This contains no functionality.


Dasku:
Need to load the save memory lines too. 

Will see if I can update from work

Spedwards:
@Dasku, you mean the things I just pushed?

Dasku:
Yep, cheers

room.js
```diff
@@ -1004,6 +1040,21 @@ mod.extend = function(){
             this.memory.spawns = _.map(spawns, id);
         } else this.memory.spawns = [];
     };
+    Room.prototype.saveObserver = function() {
+        [this.memory.observer] = this.find(FIND_MY_STRUCTURESS, {
```
karlthepagan: `FIND_MY_STRUCTURESS` is undef ;)

Dasku:
Also can we please add Power Banks to this?

room.js
Spedwards: Nice catch.

Spedwards:
Power Bank added @Dasku, and request filled @karlthepagan.

room.js
```diff
@@ -389,6 +425,23 @@ mod.extend = function(){
                 return this._sources;
             }
         },
+        'powerBank': {
+            configurable: true,
+            get: function() {
+                if (_.isUndefined(this.memory.powerBank)) {
+                    [this._powerBank] = this.find(FIND_STRUCTURES, {
```
karlthepagan: the powerbank needs a game time expiration added

Spedwards: I'm not sure what you mean. Power Banks already have a `ticksToDecay` property.

Dasku: I don't even think decay time is important. I can already see the power in the powerBank with this and if we find them from observers we will always have >4950 ticks which is more than enough.

logxen: @dasku 4500 but yeah still plenty of time

@spedwards neat thanks ... I was trying to figure out what the ticksToDecay property's name is. :P

karlthepagan: The object in memory must know when to be garbage collected.

karlthepagan:
@logxen https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/817 is on hold

I encourage you to merge this change into the needs refactor so that both can move forward. This is the smaller change and has less risk.

The right management decision is to accept this PR and merge it into the higher risk feature.

logxen:
Ok this merged fine, sorry for the trouble. I should have tried that first. I was just worried because I already had these implemented with a bunch of code on top of them already. And this clearly states that there is no actual function here. :P

karlthepagan:
I'm still not 100% fine with how powerBanks are expired, but I'm going to move this into dev so we can focus on that problem specifically.

logxen:
Ugh nope apparently this is broken. Nukers and Powerspawns are not being saved in the same format as everything else. The memory stucture is `<structureType> : [ <structureId> ]` but it is only population '<structurType> : <structureId>` so pulling this in and dropping the existing implementations I had broke it. Any chance you can fix the implementation? Otherwise I need to add special handlers for the non-standard memory entries. :P

PR #801, opened 2017-02-21
call Population.registerAction more often

karlthepagan:
if assignment changes from one flag/spawn to another

feels like a possible bug waiting to bite us

### DIFF
--- a/creep.Action.js
+++ b/creep.Action.js
@@ -89,7 +89,7 @@ let Action = function(actionName){
         if( target === undefined ) target = this.newTarget(creep);
         if( target != null ) {
             if( DEBUG && TRACE ) trace('Action', {creepName:creep.name, assign:this.name, target:!target || target.name || target.id, Action:'assign'});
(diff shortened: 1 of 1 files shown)


creep.Action.js
```diff
@@ -89,7 +89,7 @@ let Action = function(actionName){
         if( target === undefined ) target = this.newTarget(creep);
         if( target != null ) {
             if( DEBUG && TRACE ) trace('Action', {creepName:creep.name, assign:this.name, target:!target || target.name || target.id, Action:'assign'});
-            if( creep.action == null || creep.action.name != this.name || creep.target == null || creep.target.id != target.id ) {
+            if( creep.action == null || creep.action.name != this.name || creep.target == null || creep.target.id != target.id || creep.target.name != target.name ) {
```
Spedwards: These should really be `===` and `!==` comparisons. However if you do change them, make sure the values are `null` as `undefined` wouldn't match.

Otherwise, all is fine.

karlthepagan: #751 was merged this way, I kept the existing code style for now

PR #803, opened 2017-02-21
creeps may not self target

karlthepagan:
this attempts to fix the SK circular reference problem

my current hypothesis is that circular references are more likely to occur in SK rooms and this is possible outside that scenario (defenders or privateers traveling to rooms without vision)

the core mechanic is that all creeps use the "special flag" to denote the old self-targeting conditions (idle and travel)

### DIFF
--- a/flagDir.js
+++ b/flagDir.js
@@ -187,7 +187,6 @@ mod.specialFlag = function(create) {
     if (create) {
         if (!flag) {
             return _(Game.rooms).values().some(function (room) {
-                room.createFlag()
                 new RoomPosition(49, 49, room.name).createFlag(name, COLOR_WHITE, COLOR_PURPLE);
                 return true;
             });
--- a/flagDir.js
+++ b/flagDir.js
@@ -143,7 +143,6 @@ mod.flush = function(){
     delete this._hasInvasionFlag;
 };
 mod.analyze = function(){
-    const specialFlag = mod.specialFlag(true);
     let register = flag => {
         flag.creeps = {};
         if( flag.cloaking && flag.cloaking > 0 ) flag.cloaking--;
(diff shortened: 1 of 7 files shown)


flagDir.js
```diff
@@ -179,3 +181,17 @@ mod.cleanup = function(){
     let clearMemory = flagName => delete Memory.flags[flagName];
     this.stale.forEach(clearMemory);
 };
+mod.specialFlag = function(create) {
+    const name = '_OCS';
+    const flag = Game.flags[name];
+    if (create && !flag) {
+        _(Game.rooms).values().some(function(room) {
+            new RoomPosition(49, 49, room.name).createFlag(name, COLOR_WHITE, COLOR_PURPLE);
```
Spedwards: I'm confused. Would this not just create a single flag across all rooms? Maybe it should be `createFlag(room.name + name,`?

Or does it not matter that there is only one flag? In which case, why iterate over all the rooms?

karlthepagan: chain.some will stop iterating as soon as one is found. (tested!)

Spedwards: Ahh, of course.

karlthepagan:
I'm still testing this for stability. Flag creation is delayed and that's a little scary.

karlthepagan:
The creeps that run off to the center of the map? That's this change... need to work out the issues with it in traveler.

karlthepagan:
@OverlordQ says he got the circular reference error with the latest canary: 76982c3...6aee7d5

if that includes noSelfTarget then this doesn't resolve that issue

chobobobo:
I also had one recently when removing an invade.robbing flag.. I believe part of the issue may have been that one/more of my non-privateer creeps had been turned into privateers, I short circuited my main-loop and suicided all of my creeps and that resolved it.

Spedwards:
@karlthepagan Probably stable enough to get merged. Circular reference errors seem to be rare enough that it should be fine for dev.

Wouldn't be the first error of this kind to get merged.

PR #808, opened 2017-02-22
Make Level 5 And Long Haul Rooms More Robust

TafThorne:
Level 5 rooms frequently collapse. They are at a sensitive point where the code base switches from non-storage to storage base behaviour and where links are introduced.  If the player does not sequence the construction of these facilities or if one is destroyed it can cause a room to struggle or fail.  

Rooms with long haulage distances from the energy source to the spawn/extension/storage also suffer greatly as we rely heavily on haulers and workers to do the transport of energy.  If there are no working link facilities the workers and haulers can spend all their time hauling energy and recharging the spawn.  In the worst cases they do this at a rate too low to support the continued generation or workers, haulers and miners.

PR #809, opened 2017-02-22
Robust RCL5 & Smarter Hauler Spawn

chobobobo:
I cherrypicked this commits for @TafThorne to clean up the previous PR.

Quoted from #808 
Level 5 rooms frequently collapse. They are at a sensitive point where the code base switches from non-storage to storage base behaviour and where links are introduced. If the player does not sequence the construction of these facilities or if one is destroyed it can cause a room to struggle or fail.

Rooms with long haulage distances from the energy source to the spawn/extension/storage also suffer greatly as we rely heavily on haulers and workers to do the transport of energy. If there are no working link facilities the workers and haulers can spend all their time hauling energy and recharging the spawn. In the worst cases they do this at a rate too low to support the continued generation or workers, haulers and miners.

As requested by @chobobobo in a slack chat the other day, the above are my changes that make long haul and level 5 rooms more stable. There should be no net effect on any rooms that have (not got piles of energy laying about) or (lots of incomplete buildings and stored energy).

I believe this changes are now ready for merging into canary for wider testing.


TafThorne:
Thank you for doing that.  I am in no hurry so whenever you get the time to review and test the changes is fine with me.

creep.setup.hauler.js
```diff
@@ -17,12 +17,14 @@ setup.maxCount = function(room){
     let miners = (room.population.typeCount['miner']||0);
     let workers = (room.population.typeCount['worker']||0);
     let cont = room.structures.container.in.length + room.structures.links.storage.length;
-    if( miners > 0  || ( cont > 0 && workers > 2 )) {
+    if( miners > 0  || ( cont > 0 && workers > Setup['worker'].maxCount) ) {
```
karlthepagan: the changed comparison is always true I think.

`Creep.setup.worker.maxCount(room)` is what you need here

TafThorne: Is it not false when there are more workers than worker.maxCount?  I though this is what the hard coded 2 related to.  Without this chance an increased allowance of workers would reduce the number of haulers.  

It could be that my JavaScript is not very good and `Setup['worker'].maxCount` resolves to `true` rather than some integer value.

chobobobo: Yes, maxCount is a function which takes a room object as a parameter.

TafThorne: So can you pass a `(room)` to it and solve it all then?

chobobobo: Yes, the code is currently returning true because it is checking for the existence of the function, if you call the function with the room it will give you the intended result

TafThorne: I do not know how to do that.  

`Setup['worker'].maxCount(room)` ?
`room.Setup['worker'].maxCount` ?

I have no IDE and minimal JS experience.  This code base is a basically a pile of magical extensions to some existing classes to me with little in the way of recognisable patterns.

chobobobo: Yes, it takes some time to get used to JS depending on the language you are coming from.  I also have rather limited JS experience but my understanding is that basically the function maxCount is a property of Setup['worker'] which means you can set and compare it much like you would a variable.  In JS there are a great many things that can resolve to false, and in this case the current code is checking whether the property maxCount is defined and since the property is then it returns true.  What needs to be done instead is to cal l it directly to get the result so room.Setup['worker'].maxCount(room) will call the function and resolve to whatever value the function returns.

creep.setup.hauler.js
```diff
@@ -17,12 +17,14 @@ setup.maxCount = function(room){
     let miners = (room.population.typeCount['miner']||0);
     let workers = (room.population.typeCount['worker']||0);
     let cont = room.structures.container.in.length + room.structures.links.storage.length;
-    if( miners > 0  || ( cont > 0 && workers > 2 )) {
+    if( miners > 0  || ( cont > 0 && workers > room.Setup['worker'].maxCount(room) ) ) {
```
chobobobo: room.Setup does not exist, I think you mean Creep.setup.worker.maxWorker(room)

creep.setup.worker.js
```diff
@@ -1,19 +1,23 @@
 let setup = new Creep.Setup('worker');
 module.exports = setup;
 setup.maxWorker = room => {
+    let count = 1;
```
TafThorne: That would exactly work against the idea of this patch.  You do always need a worker present because they are your backup when the haulers, miners and everything else expire in an untimely manner.  What we have is a fragile system and what we need a robust one.  No they are not efficient but they are much more efficient than a dead room.

creep.setup.hauler.js
TafThorne: I do not know what I mean.  I kept being told to change that line but had no idea what to change it to.  If that works and returns the currently calculated maximum value for Workers in the room then please apply it. I can type all kind of combinations into that code line that silently fail, or pass, or who knows what so there is little value in my writing anything off the top of my head.

creep.setup.worker.js
chobobobo: Wouldn't having two haulers up instead of one hauler and one worker provide an equally useful guard against collapse?  I don't like the idea of using an inefficient creep all the time just because we might collapse.. better to resolve the actual reason for collapse.  We currently have emergency workers that spawn when a room has been detected as collapsed, and that could be improved upon I imagine.

creep.setup.hauler.js
chobobobo: @TafThorne I think you are going to need to try running this code and debugging it yourself to figure out how it works and how is it supposed to work.

creep.setup.worker.js
TafThorne: > Wouldn't having two haulers up instead of one hauler and one worker provide an equally useful guard against collapse? 
No.  Two haulers do not fix the issue of both miners dying.  That leaves you with two haulers and no energy income.  You have to have a worker around as it is the only multi-purpose creep we have.  The alternative would be to get haulers to mine when there are no miners so there is at least a small trickle of energy coming in.  

> I don't like the idea of using an inefficient creep all the time just because we might collapse.. better to resolve the actual reason for collapse.  

This is exactly what this change is doing.  The reason for collapse is that rooms that are too far from the ideal (long paths, lots of building sites, just invaded and lost a creep) are shattered.  Then they recover by spawning 1 worker that builds a miner, that gets some haulers then the whole thing dies and you repeat the cycle.  Until there is a:

- huge pile of energy in a storage unit (that might not exist at early level 5)
- and working links to transfer energy without long haulage times (that also might not exist and level 5) 
- and containers to prevent drop mining of energy and later minerals (which expire at room collapse) 

Some rooms at best enter a cycle of recovery and collapse.  You need a creep generating steady income, completing building projects (to get the storage, links or containers in place) and healing the rapidly expiring infrastructure.  The creep we have that can do that is the worker.  The hauler is better at moving mined energy but it cannot mine it, the miner is better at extracting energy but it cannot transport it.  

> We currently have emergency workers that spawn when a room has been detected as collapsed, and that could be improved upon I imagine.

You get a few workers, then a miner appears and then the whole thing enters the fragile stage again.  These changes are that improvement.  You cannot spawn your extra worker after collapse, you have no stored energy, you have no income.

creep.setup.hauler.js
TafThorne: @chobobobo I have run this code and it seemed fine.  It worked.  It stopped my struggling rooms from collapsing.  It let two new rooms pass through level 5 without a hickup.  The only issue identified is that I may have wrongly attempted to look at the maximum number of workers calculated for a room.  I have repeatedly said that I am not a JS expert and asked for help in doing one thing specific to this code base.  There are no errors generated in what I do.  Is it too much to ask for one person in the whole of OCS to take the time to say what the code line should be to look up the current calculated value of the maximum number of workers should be from within another file?  Apparently it is.  Karlthepagan and yourself have only told me it is wrong, that is not collaborating it is criticising.  If you do not want my contributions to this code base just say so.  Do not waste my time by dragging things out for two weeks after requesting I make the changes in a slack chat then changing your mind.

creep.setup.worker.js
chobobobo: Yes, I think setting it to 1 for RCL1-7 and 0 for RCL8 would be a good solution.

creep.setup.hauler.js
chobobobo: My apologies, I was coming to this issue in stops and starts and hadn't realized that I was the one who initially gave you the misinformation about what specifically to change the line to.  After realizing that, I understand your frustration.

chobobobo:
@karlthepagan I have made the changes to fix the errors and rebased this onto dev, I will put it into canary to soak.

PR #811, opened 2017-02-22
Feature/action avoiding

karlthepagan:
Right now this lets remoteMiners avoid source keepers.

It's a foundation for more danger avoiding cases in the future.


creep.action.avoiding.js
```diff
@@ -0,0 +1,74 @@
+let action = new Creep.Action('avoiding');
```
Spedwards: I know it's consistent with other action files, but we have to start somewhere. This should be `const`.

creep.action.avoiding.js
```diff
@@ -0,0 +1,74 @@
...
+};
+action.isValidTarget = function(target, creep){
+    return Task.reputation.hostileOwner(target) && action.isActiveLair(target);
+};
+action.newTarget = function(creep) {
+    delete creep.data.determinatedSpot;
+
+    if (Room.isSKRoom(creep.pos.roomName)) {
+        const target = _.first(creep.room.find(FIND_STRUCTURES, {filter: function (t) {
+            return !_.isUndefined(t.ticksToSpawn) && action.isActiveLair(t) && creep.pos.getRangeTo(t.pos) < 15;
```
Spedwards: Should check if the structure is an instance of `StructureKeeperLair`, so you aren't iterating over every structure in the room.

Spedwards: Actually, the undefined check should be fine. Checking the instance would however be more clear.

karlthepagan: yeah undefined check is duck typing! ;)

Spedwards: God forbid another structure is added with a `ticksToSpawn` property, it'd be best to still check for the instance. Future proof, and clear as to what's happening. `t.structureType === STRUCTURE_KEEPER_LAIR` would work too.

karlthepagan: I've got to keep pushing back tho. Any hostile structure with a defined ticksToSpawn is going to be a possible target for this algorithm.

karlthepagan:
I will declare a national holiday.

PR #812, opened 2017-02-22
all creeps using flag & setup spawn will react to predicted renewal

karlthepagan:
I'm using this to better implement predictedRenewal to have my SKK's respawn at the needed time.


chobobobo:
@karlthepagan this has been a hairy one, is there any way you would suggest to test whether it is functioning as intended?  Check predictedRenewal in creep.data?

karlthepagan:
source killers are really effective at testing this

karlthepagan:
my last fix has solved my over spawning problem

PR #814, opened 2017-02-22
Task/power mining

Dasku:
Bare bones task. Struggling with the attacking the structure, trying to use invade action. 

Pre-requisites is the RCL8 Structures branch. 

First train spawns, arrives but wont attack the structure. Haulers shouldn't spawn until hp is low.

Miner and healers currently spawn with 2 parts each for ease until its working.


Dasku:
changed to powerMine action but still will not attack the power bank. Need to find out if it requires something special.

Dasku:
This should be working now, Red/Brown flag on the powerbank, should auto complete.

karlthepagan:
I need to reconcile this with actionAvoiding (which should be ready soon).

Dasku:
@karlthepagan ok thanks.

Branch is ready for testing - I'm running it and have collected 2 powerBanks so far. Feels good.

Dasku:
TypeError: Cannot read property 'roomName' of undefined
    at Action.action.step (creep.action.travelling:19:42)
    at Object.mod.run (creep.behaviour.powerHealer:9:22)
    at Creep.run (creep:47:39)
    at run (creep:447:30)
    at Function.mod.execute (creep:448:7)

This happens when 3rd miner spawns (just before the 1st miner dies). Evidence I need to update this to traveller and use assignRoom in the behaviours once this is merged in canary

Dasku:
Today haulers idled at the edge of the room they entered from instead of waiting 4 squares away from the powerBank - could be traveller related

Spedwards:
To add support from observers (#828), create a method in room such as the below, and call it in `mod.analyze`

```
Room.prototype.checkPowerBank = function() {
    if (!this.powerBank) return; // no power bank in room
    // some other checks
    this.createFlag(this.powerBank.pos, null, FLAG_COLOR.invade.powerMining.color, FLAG_COLOR.invade.powerMining.color);
}
```

PR #816, opened 2017-02-22
random missing object check in room

logxen:


logxen:
nevermind this was caused by something else.

PR #820, opened 2017-02-23
more accurate sign message

karlthepagan:
Tested!

image

This might be "undefined" for a few ticks after first spawn, but meh.


parameter.js
```diff
@@ -114,7 +114,7 @@ let mod = {
     COST_MATRIX_VALIDITY: 1000,
     CONSTRUCTION_PRIORITY: [STRUCTURE_SPAWN,STRUCTURE_EXTENSION,STRUCTURE_LINK,STRUCTURE_STORAGE,STRUCTURE_TOWER,STRUCTURE_ROAD,STRUCTURE_CONTAINER,STRUCTURE_EXTRACTOR,STRUCTURE_WALL,STRUCTURE_RAMPART],
     CONTROLLER_SIGN: false,
-    CONTROLLER_SIGN_MESSAGE: "Territory of the Open Collaboration Society! (https://github.com/ScreepsOCS)",
+    CONTROLLER_SIGN_MESSAGE: `Territory of ${_.chain(Game.spawns).values().first().get('owner.username').value()}, an Open Collaboration Society user! (https://github.com/ScreepsOCS)`,
```
Spedwards: Not sure why you're using `_.chain` instead of `_` but the changes are nonetheless fine.

karlthepagan: unfortunately not using _.chain means that .first() returns a literal value which will not provide .get()

Spedwards: So use `_(Game.spawns).values().first().owner.username`?

karlthepagan: .get will be undef safe!

Spedwards: Live on the edge a little. ;)

PR #824, opened 2017-02-24
universal room object cloaking

karlthepagan:
simple enough, try it!

`Game.getObjectById(sourceId).cloak = 100`

Every part of OCS will ignore that source for 100 ticks. (edit: that is if the sources are accessed by room.sources)

`Game.getObjectById(sourceId).cloak = true`

That source will be ignored forever.

Data is stored in `Memory.cloaking`.


Spedwards:
Wouldn't the correct use be `Game.getObjectById(sourceId).cloak = true`?

karlthepagan:
Oops. Yes. :)

karlthepagan:
Documented here https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/wiki/Settings-&-Tuning

PR #826, opened 2017-02-24
changing position some visuals

Perrytheplatypus43:
I am changing the position of the Storage Contents visual and the Terminal Contents visual. The new position of the visuals is lined up with the energy visual.


Spedwards:
I'm going to have to disagree with this change. It's pointless. It seems as though you just want to change it for you, in which case you should be creating a viral file.

karlthepagan:
@spedwards I see down the rabbit hole of customizing visuals... I hate to say that everything should be parameterized, but I'm not sure what a quick fix would look like

PR #828, opened 2017-02-24
Feature/observer

Spedwards:
Depends on #797


Spedwards:
Please don't judge me by my commits :(

karlthepagan:
need to finish process power feature before I can canary this

logxen:
I pushed process power in #823 

I have to say though ... I'm not really sure about this plan for observers. It's going to consume extra cpu for no benefit. I'm still waiting to hear a reason to scan anything other than corridors.

Second ... prioritizing corridors is pointless. Scanning the corridors once every room tick is more than sufficient to find all the power banks in plenty of time to deal with them.

Spedwards:
@logxen @Dasku see the latest commit.

Dasku:
Error in room.js (Room.prototype.loop) for "E74N54": 
TypeError: Cannot read property 'rooms' of undefined
    at Room.controlObserver (room:1934:35)
    at getEnvironment (room:2046:18)
    at /opt/engine/node_modules/lodash/index.js:3073:15
    at /opt/engine/node_modules/lodash/index.js:3043:18
    at Function. (/opt/engine/node_modules/lodash/index.js:3346:13)
    at Function.mod.analyze (room:2053:7)
    at Object.module.exports.loop (main:267:10)
    at __module (__mainLoop:1:52)
    at __mainLoop:2:3
TypeError: Cannot read property 'rooms' of undefined
    at Room.controlObserver (room:1934:35)
    at getEnvironment (room:2046:18)
    at /opt/engine/node_modules/lodash/index.js:3073:15
    at /opt/engine/node_modules/lodash/index.js:3043:18
    at Function. (/opt/engine/node_modules/lodash/index.js:3346:13)
    at Function.mod.analyze (room:2053:7)
    at Object.module.exports.loop (main:267:10)
    at __module (__mainLoop:1:52)
    at __mainLoop:2:3

 and 

Error in room.js (Room.prototype.loop) for "E79N55": 
TypeError: Cannot read property 'observeRoom' of undefined
    at Room.controlObserver (room:1947:18)
    at getEnvironment (room:2046:18)
    at /opt/engine/node_modules/lodash/index.js:3073:15
    at /opt/engine/node_modules/lodash/index.js:3043:18
    at Function. (/opt/engine/node_modules/lodash/index.js:3346:13)
    at Function.mod.analyze (room:2053:7)
    at Object.module.exports.loop (main:267:10)
    at __module (__mainLoop:1:52)
    at __mainLoop:2:3
TypeError: Cannot read property 'observeRoom' of undefined
    at Room.controlObserver (room:1947:18)
    at getEnvironment (room:2046:18)
    at /opt/engine/node_modules/lodash/index.js:3073:15
    at /opt/engine/node_modules/lodash/index.js:3043:18
    at Function. (/opt/engine/node_modules/lodash/index.js:3346:13)
    at Function.mod.analyze (room:2053:7)
    at Object.module.exports.loop (main:267:10)
    at __module (__mainLoop:1:52)
    at __mainLoop:2:3

room.js
```diff
@@ -1971,13 +2026,23 @@ mod.isMine = function(roomName) {
...
 };
 
+mod.calcCardinalDirection = function(roomName) {
+    const parsed = /^([WE])[0-9]{1,2}([NS])[0-9]{1,2}$/.exec(roomName);
+    return [parsed[1], parsed[2]];
+};
+mod.calcGlobalCoordinates = function(roomName, callBack) {
+    if (!callBack) return null;
+	const parsed = /^[WE]([0-9]+)[NS]([0-9]+)$/.exec(roomName);
+	const x = +parsed[1];
```
karlthepagan: make sure I haven't missed this correction elsewhere

Spedwards: I correct this in lines 2001 and 2007 of the same file. Changing it in this method would also require changing it in `calcCoordinates`.

Dasku:
Does range work like the diagram below? If so we may want to reduce default max range to 3?
image

Spedwards:
If it works the same as room coordinates does, @Dasku, then yes, it would work like that. And after seeing that image, I'm inclined to agree with you. Until any other functionality is added for observers, 3 is probably the optimal range.

logxen:
How about a range of 4? that would guaranteed that an inner room next to the SKs would still be able to see the highway.

main.js
```diff
@@ -241,6 +241,7 @@ module.exports.loop = function () {
     if (Memory.cloaked === undefined) {
         Memory.cloaked = {};
     }
+    Memory.observerSchedule = [];
```
chobobobo: I think this should go in Room.flush(), with a check for existence the first time it is used.

Spedwards: @chobobobo It doesn't need a check for existence. It needs to be reset each tick. As for moving it to `Room.flush()`, that's for room specific memory stuff. `Memory.observerSchedule` is global across all rooms, thus fits better where it is.

Perhaps you could dismiss this so @karlthepagan could merge it?

chobobobo: Room.flush() is used to reset each room, but can also be used to reset data relevant to all rooms.  The reason I suggested this is that not everyone will have/use observers, so it doesn't seem like something that should be in our main run loop as it is data that will only be used by rooms that have observers in them. It is not used by or relevant to Population, FlagDir, etc, so keeping it in rooms in the place where memory is currently reset on each tick would make it easier to find for someone not familiar with how it works.  If it was moved to Room.flush() it would need a check for existence otherwise it would be undefined on first use.

Spedwards:
Changes made meet @chobobobo's request. All ready to merge.

PR #836, opened 2017-02-25
creeps heal when fleeing

karlthepagan:
ranger is the reference for healing


karlthepagan:
Still need to investigate my changes to internal healing, but this works for public.

PR #832, opened 2017-02-25
Shared logic for all miners, with separate source selection.

chobobobo:
This lays the foundation for being able to define all kinds of miners as "Miners" with dynamic source selection (energy/minerals) and location (local/remote).

PR #837, opened 2017-02-25
Reduce Miner CPU Usage

chobobobo:
Reduce how often miners look for repairs and construction sites, every tick is a bit excessive.


chobobobo:
This was merged into #832

PR #840, opened 2017-02-25
attempt to solve routing problems first using transit of adjacent room

karlthepagan:
this adds the OCS room-by-room pathfinding to traveler as a fallback when things go wrong

### DIFF
--- a/traveler.js
+++ b/traveler.js
@@ -341,7 +341,7 @@ module.exports = function(globalOpts = {}){
             let offsetY = [0, -1, -1, 0, 1, 1, 1, 0, -1];
             return new RoomPosition(origin.x + offsetX[direction], origin.y + offsetY[direction], origin.roomName);
         }
-    }    
+    }
 
     if(gOpts.installTraveler){
         global.Traveler = Traveler;
@@ -364,6 +364,7 @@ module.exports = function(globalOpts = {}){
             if (_.isUndefined(options.preferHighway)) options.preferHighway = true;
             if (_.isUndefined(options.allowHostile)) options.allowHostile = false;
             if (_.isUndefined(options.routeCallback)) options.routeCallback = Room.routeCallback(destination.roomName, options.allowHostile, options.preferHighway);
+            if (_.isUndefined(options.useFindRoute)) options.useFindRoute = global.ROUTE_PRECALCULATION;
             return traveler.travelTo(this, destination, options);
         };
     }
--- a/traveler.js
+++ b/traveler.js
@@ -26,6 +26,7 @@ module.exports = function(globalOpts = {}){
         maxOps:            20000,
         defaultStuckValue: 3,
         reportThreshold:   50,
+        roomRange:         22,
     });
     class Traveler {
         constructor() {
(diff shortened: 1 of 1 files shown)


karlthepagan:
@chobobobo it was intentional: I use global.ROUTE_PRECALCULATION in order to avoid the undef error and make the parameter optional.

karlthepagan:
I'm definitely hitting his case from reservers who arrive from the room to the west of this: E81S11

traveler.js
```diff
@@ -26,6 +26,7 @@ module.exports = function(globalOpts = {}){
         maxOps:            20000,
```
TooAngel: In the description is written `6 options` but there are actually 8.
And `roomRange` is missing in the description.

traveler.js
```diff
@@ -130,12 +132,14 @@ module.exports = function(globalOpts = {}){
                 }
                 return matrix;
             };
-            return PathFinder.search(origPos, { pos: destPos, range: options.range }, {
+            const ret = PathFinder.search(origPos, { pos: destPos, range: options.range }, {
```
TooAngel: Define `ret` as const and change it afterwards `ret.route`. Is that fine for objects?

traveler.js
```diff
@@ -231,8 +235,17 @@ module.exports = function(globalOpts = {}){
                     console.log(`TRAVELER: heavy cpu use: ${creep.name}, avg: ${travelData.cpu / travelData.count}, total: ${_.round(travelData.cpu, 2)}, pos: ${creep.pos}`);
                 }
                 if (ret.incomplete) {
-                    console.log(`TRAVELER: incomplete path for ${creep.name} from ${creep.pos} to ${destPos}`);
-                    if (ret.ops < 2000 && options.useFindRoute === undefined && travelData.stuck < gOpts.defaultStuckValue) {
+                    const route = ret.route && ret.route.length;
+                    console.log(`TRAVELER: incomplete path for ${creep.name} from ${creep.pos} to ${destPos}. Route length ${route}.`);
+                    if (route > 1) {
+                        ret = this.findTravelPath(creep, new RoomPosition(25, 25, ret.route[1].room),
```
TooAngel: I like that, we do that, too.

traveler.js
karlthepagan: 👍

traveler.js
karlthepagan: Is is permitted. Only frozen objects are immutable in ES6.

traveler.js
karlthepagan: What target ranges have you tested? I'm preferring 22 (so 47 -> 3).

TooAngel: I chose 22, too. I know that someone played around with the value, not sure what was the conclusion.

karlthepagan: 24 leads to arriving on a swamp even with vision. I'd still entertain 23.

chobobobo:
Unfortunately, for the room pathing issues I'm having no route is returned so this does not help.  Are there any cases you've found where an incomplete route is returned and this finds a route?

The way I'm reading it right now, this will simply get you to the next room on the route returned, but will ultimately end up in the same room that returned an incomplete route in the first place resulting in a dead end.

karlthepagan:
With precalc on it seems reliable.

karlthepagan:
Since this is going into traveler branch I'm going to merge. it's an improvement and hopefully gets traveler closer to dev.

PR #853, opened 2017-02-26
Change target selection from ByPath to ByRange

chobobobo:
This comes with the possibility that there will be some unforseen base layout that causes creeps to select a target that is unreasonably far away.  If you have one of these issues, please report it with details.


karlthepagan:
Going to let this one bake in canary for a while.

karlthepagan:
I've been in a bunch of odd situations and I don't think this has caused me great harm.

PR #855, opened 2017-02-26
Reduce defender CPU usage

chobobobo:
Make sure we are travelling to the room, not the creep when we are not in the same room, the creep can move and that means rebuilding the path!

### DIFF
--- a/creep.action.defending.js
+++ b/creep.action.defending.js
@@ -21,6 +21,7 @@ action.newTarget = function(creep){
 };
 action.step = function(creep){
     if(CHATTY) creep.say(this.name, SAY_PUBLIC);
+    if (creep.target.pos.roomName !== creep.room.name) return Creep.action.travelling.assignRoom(creep, creep.target.pos.roomName);
     this.run[creep.data.creepType](creep);
 };
 action.run = {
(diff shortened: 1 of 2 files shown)


creep.action.defending.js
```diff
@@ -21,6 +21,7 @@ action.newTarget = function(creep){
 };
 action.step = function(creep){
     if(CHATTY) creep.say(this.name, SAY_PUBLIC);
+    if (creep.target.pos.roomName !== creep.room.name) return Creep.action.travelling.assignRoom(creep, creep.target.pos.roomName);
```
karlthepagan: This requires traveler.

Also as we discussed assignRoom should terminate at the border room if you have vision.

so this will need to be something like `if (target.roomName != creep.roomName && assignRoom(creep)) return true;`

karlthepagan:
Maybe I can backport assignRoom into dev.

creep.action.defending.js
chobobobo: Right, I can just merge this into the traveler branch instead if it looks good.

karlthepagan: +1 put this into traveler, it's a smaller change

chobobobo:
Merged into traveler.

PR #856, opened 2017-02-26
RemoteHaulers use storage in more cases

chobobobo:
Previously when remoteHaulers found a potentially chargable target closer than storage, they would potentially ignore the storage.


Dasku:
why is charging a bad thing if its closest? surely where the energy goes should be based on distance not container type?

chobobobo:
@Dasku this is specifically the case when target is an object that is not storage, and then trying to assign the charging action on that target returns false.  In this case, the remoteHaulers were defaulting to dropping, when they should have been putting the energy in storage if possible.

PR #858, opened 2017-02-26
pioneers learned to build up neutral spaces

karlthepagan:
i found this extremely useful, there's better detection for the neutral room error case and `PIONEER_UNOWNED: true` enables neutral pioneering


karlthepagan:
I owe my new room to these pioneers. But please try it.

PR #859, opened 2017-02-26
Allow energy transfer to RCL8 rooms

Dasku:


logxen:
Some hysteresis of at least 50-100k should alse be added here.

A situation could occur where two terminals could send energy back and forth (though it would only happen every 500 ticks) because the line for sending is > 80% and the line for receiving is < 80%. Maybe change the qualifier for receiving to < 60% and all should be well.

PR #875, opened 2017-02-28
optimize hasActiveBodyparts

karlthepagan:
iterate starting at the damage level


creep.js
```diff
@@ -33,9 +33,23 @@ mod.extend = function(){
...
+        const limit = body.length;
+        const position = Math.ceil(this.hits * 0.01);
+        if(Array.isArray(partTypes)) {
+            for (let i = limit - position; i < limit; i++) {
+                if (partTypes.includes(body[i].type)) {
+                    return true;
+                }
+            }
+        } else {
+            for (let i = limit - position; i < limit; i++) {
```
Spedwards: Duplicate loop. Move array check inside the loop.

creep.js
```diff
@@ -33,9 +33,23 @@ mod.extend = function(){
     // Accepts a single part type (like RANGED_ATTACK) or an array of part types. 
     // Returns true, if there is at least any one part with a matching type present and active.
     Creep.prototype.hasActiveBodyparts = function(partTypes){
```
Spedwards: Potentially change to `function(...partTypes)`, then you can just use `for (let part of partTypes)`, as it will always be an array, and it makes the array check later obsolete.

Spedwards: Only problem with this change is, any current use of this method where an array is used will need to be updated.

creep.js
karlthepagan: Can you express what the tradeoff is?

DRY for its own sake is insufficient, see http://joelabrahamsson.com/the-dry-obsession/

creep.js
karlthepagan: I opted to keep the scope the same.

creep.js
Spedwards: The benefits are that it's clearer as to what's going on. There's no getting confused over a duplicate loop. The downside is that you'd be running the array check each iteration, unless the result it stored before the loop.

PR #876, opened 2017-02-28
only avoid offensive military creeps

karlthepagan:
don't run from civilians

### DIFF
--- a/creep.action.avoiding.js
+++ b/creep.action.avoiding.js
@@ -14,7 +14,12 @@ action.isAddableAction = function(creep) {
     return true;
 };
 action.isValidTarget = function(target, creep){
-    return Task.reputation.hostileOwner(target) && action.isActiveLair(target);
+    if (Task.reputation.npcOwner(target)) {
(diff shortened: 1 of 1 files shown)


creep.action.avoiding.js
```diff
@@ -14,7 +14,12 @@ action.isAddableAction = function(creep) {
     return true;
 };
 action.isValidTarget = function(target, creep){
-    return Task.reputation.hostileOwner(target) && action.isActiveLair(target);
+    if (Task.reputation.npcOwner(target)) {
+        return action.isActiveLair(target);
+    } else if (Task.reputation.hostileOwner(target) && target.hasActiveBodyparts) {
+        return target.hasActiveBodyparts([ATTACK,RANGED_ATTACK]);
```
Spedwards: Might be worth checking all parts, not just active ones. A healer could be following the target creep. Our creeps should get a head-start.

karlthepagan: I'll think about making that a new feature.

PR #881, opened 2017-03-01
allow sort value set in creep body params to function

logxen:


creep.js
```diff
@@ -501,6 +501,7 @@ mod.partsComparator = function (a, b) {
 };
 // params: {minThreat, maxWeight, maxMulti}
 mod.compileBody = function (room, params, sort = true) {
+    if (params.sort !== undefined) sort = params.sort;
```
Spedwards: I don't see the point in this. Could you explain why this would be used instead of just passing in the `sort` parameter?

logxen: I mostly noticed that the new powerMiner tried to do that but wasn't getting sorted. :)

PR #883, opened 2017-03-01
Attempt to avoid exits and sources as well when idling.

chobobobo:


creep.js
```diff
@@ -298,9 +298,13 @@ mod.extend = function(){
         if( here && here.length > 0 ) {
             let path;
             if( !this.data.idlePath || this.data.idlePath.length < 2 || this.data.idlePath[0].x != this.pos.x || this.data.idlePath[0].y != this.pos.y || this.data.idlePath[0].roomName != this.pos.roomName ) {
-                let goals = _.map(this.room.structures.all, function(o) {
+                let goals = _.chain(this.room.structures.all.map(function(o) {
```
Spedwards: If you're going to use a chain, `_.chain(this.room.structures.all),map`. Would be easier than having the initial map inside the chain.

chobobobo:
@Spedwards there, I've cleaned up the code, how is it now?

creep.js
```diff
@@ -298,9 +298,13 @@ mod.extend = function(){
         if( here && here.length > 0 ) {
             let path;
             if( !this.data.idlePath || this.data.idlePath.length < 2 || this.data.idlePath[0].x != this.pos.x || this.data.idlePath[0].y != this.pos.y || this.data.idlePath[0].roomName != this.pos.roomName ) {
-                let goals = _.map(this.room.structures.all, function(o) {
+                let goals = this.room.structures.all.map(function(o) {
                     return { pos: o.pos, range: 1 };
-                });
+                }).concat(this.room.sources.map(function (s) {
+                    return { pos: s.pos, range: 2 };
+                })).concat(this.pos.findInRange(FIND_EXIT, 2).map(function (e) {
```
karlthepagan: At first this made me nervous about performance, but I see the bounds of this are reasonable.

Issue #869, opened 2017-02-27
Cached Paths


Spedwards:
So I've had this thought for cached paths.

```
Memory.rooms['<roomName>'].paths = {
    someDestID: {
        route: [
            {x: 10, y: 20, roomName: 'W0N0', nextDirection: 2},
            {x: 11, y: 19, roomName: 'W0N0', nextDirection: 8},
        ],
    }
}
```

You would have a set list of paths. The object key would be either the ID for a structure, or a flag name. The route would contain a list of room positions with the direction required for the creep to get to the next position.

The paths should be set from a spawn. If a creep starts them from any other position, they should look for the closest position in the route array, move there, and start their path.

The creep's data property should contain the index for the route the creep is up to, and the ID for the route. So using the above format, you could get it down to a one-liner once the creep is on the path:

`creep.move(creep.room.paths[creep.data.pathID].route[creep.data.pathIndex++].nextDirection);`

Spedwards:
By storing the room positions, you are able to also efficiently go through rooms. The only problem with the system laid out, is when `creep.room` does not equal the room the path belongs to.

chobobobo:
I like this, my current thinking is to break down the overall route even further, into a list of rooms that need to be traversed, then as the creep is going room by room it could load paths from a cache like this to get the directions to get it from the entrance to the exit within that room.

Spedwards:
Another thought which would solve the problem aforementioned. Change the structure to:

```
Memory.paths = {
    someDestID: {
        startRoomA: {
            route: [
                {x: 10, y: 20, roomName: 'W0N0', nextDirection: 2},
                {x: 11, y: 19, roomName: 'W0N0', nextDirection: 8},
            ],
        },
        startRoomB: {
            // ...
        }
    }
}
```

Then all you would have to do is additionally save the starting room to the creep's data. Allows a path to have multiple starting points. Since paths can go through rooms and different rooms can have paths to the same dest, it seems better this way.

chobobobo:
Would this allow two paths ABC and ABCD to share the same route information for rooms BC without duplication?

logxen:
they should be able to share B ... C maybe not though

Spedwards:
It could. You could have separate smaller routes from room to room, which the more specific routes could use if need be.

chobobobo:
[...]

PR #885, opened 2017-03-01
Feature/traveler caching

chobobobo:
A work in progress branch to cache all routes that use travelBy and reuse them. Currently if there is not a saved route from the creep's position one is generated, if the generated route crosses an existing route to that destination they merge.

There is a good chance there are CPU issues with this form of pathfinding not respecting ranges, any creep that tracks a live/moving target will likely fail as well (caching should be turned off for them).  Moving a flag will likely break as well.

The memory usage might be a bit obscene as well.

This will likely conflict with assignRoom, and perhaps will eventually replace the need for it.

### DIFF
--- a/room.js
+++ b/room.js
@@ -987,6 +987,24 @@ mod.extend = function(){
         });
     };
 
+    Room.prototype.getPath = (start, finish, options) => {
+        let path = this.memory.paths[creep.pos.x + ',' + creep.pos.y][destID];
+        if (!path) {
+            if(global.traveler && global.travelerTick !== Game.time){
+                global.traveler = new Traveler();
+            }
+            const ret = traveler.findTravelPath(start, finish, options);
+            if (!ret || ret.incomplete) {
+                logError('Room.getPath incomplete path from' + start + finish);
+                return;
+            } else {
+                path = Traveler.serializePath(start, ret.path);
+            }
+            this.memory.paths[creep.pos.x + ',' + creep.pos.y][destID] = path;
+        }
+        return path;
+    };
+
     Room.prototype.getBestConstructionSiteFor = function(pos, filter = null) {
         let sites;
         if( filter ) sites = this.constructionSites.filter(filter);
--- a/room.js
+++ b/room.js
@@ -987,20 +987,29 @@ mod.extend = function(){
         });
     };
 
-    Room.prototype.getPath = (start, finish, options) => {
-        let path = this.memory.paths[creep.pos.x + ',' + creep.pos.y][destID];
-        if (!path) {
-            if(global.traveler && global.travelerTick !== Game.time){
-                global.traveler = new Traveler();
-            }
-            const ret = traveler.findTravelPath(start, finish, options);
+    Room.prototype.getPath = function(startPos, destination, options) {
+        if (_.isUndefined(this.memory.paths)) this.memory.paths = {};
+        const startID = startPos.x + ',' + startPos.y;
+        const destPos = destination.pos || destination;
+        const destID = destination.id || destination.x + ',' + destination.y;
+        if (_.isUndefined(this.memory.paths[destID])) this.memory.paths[destID] = {};
+        let path = this.memory.paths[destID];
+        if (!path || !path[startID]) {
+            const ret = traveler.findTravelPath(startPos, destPos, options);
             if (!ret || ret.incomplete) {
-                logError('Room.getPath incomplete path from' + start + finish);
+                logError('Room.getPath incomplete path', 'from ' + startPos + ' to ' + destPos);
                 return;
             } else {
-                path = Traveler.serializePath(start, ret.path);
+                const directions = Traveler.serializePath(startPos, ret.path);
+                path[startID] = directions[0];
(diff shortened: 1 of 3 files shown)


room.js
```diff
@@ -987,6 +987,35 @@ mod.extend = function(){
         });
     };
 
+    Room.prototype.getPath = function(startPos, destination, options) {
+        if (_.isUndefined(this.memory.paths)) this.memory.paths = {};
+        const startID = startPos.x + ',' + startPos.y;
```
Spedwards: Maybe use the same system that `roadConstructionTrace` and `Visuals.heatmap` uses.

See [Visuals L360](https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/blob/dev/visuals.js#L360) for a method to set, and the next method down to retrieve.

room.js
```diff
@@ -987,6 +987,35 @@ mod.extend = function(){
...
+        if (!path || !path[startID]) {
+            const ret = traveler.findTravelPath(startPos, destPos, options);
+            if (!ret || ret.incomplete) {
+                logError('Room.getPath incomplete path', from + startPos + finish);
+                return;
+            } else {
+                //console.log(startPos, destPos, ret.path);
+                path = this.memory.paths[destID];
+                const directions = Traveler.serializePath(startPos, ret.path);
+                path[startPos.x+','+startPos.y] = directions[0];
```
Spedwards: `path[startID]`.

chobobobo:
We will need to find a clever way to implement getting 'unstuck', what I have done in the past is generate roomPositions from the next 5-10 moves and then use them as goals for pathfinding with ignoreCreeps=false.  Once you reach that position, you can continue following the path again.

Another potential way to do it is to randomly try to move on a diagonal and back again, this might clear jams as well.
screen shot 2017-03-01 at 10 39 21 pm

room.js
chobobobo: I'll take a look at it

room.js
```diff
@@ -987,6 +987,35 @@ mod.extend = function(){
...
+        if (_.isUndefined(this.memory.paths[destID])) this.memory.paths[destID] = {};
+        let path = this.memory.paths[destID];
+        if (!path || !path[startID]) {
+            const ret = traveler.findTravelPath(startPos, destPos, options);
+            if (!ret || ret.incomplete) {
+                logError('Room.getPath incomplete path', from + startPos + finish);
+                return;
+            } else {
+                //console.log(startPos, destPos, ret.path);
+                path = this.memory.paths[destID];
```
Spedwards: `this.memory.paths` is not populated between this call and the previous, and `destID` never changes, so `path` or `path[startID]` would still be undefined.

room.js
chobobobo: Why is it that we prepend that key with the other characters?

Spedwards: The ``${String.fromCharCode(32+x)}${String.fromCharCode(32+y)}` part is so you can convert the position number to a single string character. Convert it back and forth easily. (No idea why 32 is added).

The last part (``${String.fromCharCode(32+x)}${String.fromCharCode(32+y)}`) is appended I think if you wanted to manually check and know what you're looking at.

room.js
```diff
@@ -1000,14 +1000,12 @@ mod.extend = function(){
...
-                //console.log(startPos, destPos, ret.path);
-                path = this.memory.paths[destID];
                 const directions = Traveler.serializePath(startPos, ret.path);
-                path[startPos.x+','+startPos.y] = directions[0];
+                path[startID] = directions[0];
                 for (let i = 0; i < ret.path.length; i++) {
                     const p = ret.path[i];
                     // use existing path
-                    if (_.isUndefined(path[p.x+','+p.y])) path[p.x+','+p.y] = directions[i+1];
+                    if (_.isUndefined(path[startID])) path[startID] = directions[i+1];
```
Spedwards: I believe you went too far on this change. Now `p` is defined but never used. Almost as if you changed it without thinking 😛

room.js
chobobobo: Ok, so the idea is that it optimizes for memory encoding?

Spedwards: Not sure what the benefits are. It would be consistent with the rest of the codebase. This would be an optional change though, I'll leave it up to you.

traveler.js
```diff
@@ -357,14 +357,31 @@ module.exports = function(globalOpts = {}){
...
+                if (path){
+                    const next = path[this.pos.x + ',' + this.pos.y];
+                    if (next) {
+                        //console.log(this.name, this.pos, 'cached', next);
+                        return this.move(next); // take next step
+                    }
+                    else { // TODO:find closest place to get on the path
+                        console.log(this.name, 'could not generate or use cached route, falling back to traveler.');
+                        options.cacheRoutes = false;
+                        return this.travelTo(destination, options);
```
Spedwards: Instead of calling the function again, remove this line, move the `else` branch on L379 down to a new if-statement, with the condition `!options.cacheRoutes`.

chobobobo:
This is based on #869

chobobobo:
Will revisit this once traveler has stabilized and merged.

PR #889, opened 2017-03-01
Fix upgrader/hauler for RCL8.

pandabear41:
Had a situation where 6 haulers spawned because upgrader.maxCount was
returning 5 during RCL8. Only one hauler was spawning due to #731. See [creep.setup.hauler.js#L22](https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/blob/dev/creep.setup.hauler.js#L22)
Also had upgradeBlocked and upgraders were still spawning.


karlthepagan:
Is this improved by https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/891

I would be happier if we knew what the root cause was.

pandabear41:
It didn't improve by #891.

 This problem is caused by #731
 The line in creep.setup.hauler line 22 accesses upgrader maxCount.
`count += Creep.setup.upgrader.maxCount(room);`

But creep.setup.upgrader doesn't use maxCount at RCL8, but set manually to 1. Lines 70-77 and 86.
```
setup.level8 = {
    fixedBody: [CARRY, MOVE, MOVE, MOVE],
    multiBody: [WORK],
    minAbsEnergyAvailable: 300,
    minEnergyAvailable: 1,
    maxMulti: CONTROLLER_MAX_UPGRADE_PER_TICK / UPGRADE_CONTROLLER_POWER,
    maxCount: 1
};
```
`    8: setup.level8`

The code in Creep.setup.upgrader.maxCount didn't exclude RCL8 from count calculation. 

In my situation I had full storage of energy (like 900k) so Creep.setup.upgrader.maxCount was requesting like 5 upgraders, thus 6 haulers were spawning (5 for upgraders, 1 for mineral miner).

chobobobo:
I wrote a patch for this I believe.. let me find it.

karlthepagan:
I'll let this go. I'm fairly certain that #891 will also resolve this problem, and it is the underlying root cause: two different parts of the setup flow are retrieving the value using different methods.

chobobobo:
Yes, I believe the root cause is that the maxHaulers uses creep.setup.upgrader.maxCount(room) but the actual setup was hard-coded to use 1.  creep.setup.upgrader.maxCount(room) was not written to handle RCL 8 rooms.

What I'm still not clear on is why upgraders were still spawning for him and what upgradeBlocked meant.

PR #891, opened 2017-03-01
creep.Setup property dereferencing is more consistent

karlthepagan:
This resolves some major inconsistencies with creep setup customization.

Hopefully this will help us realize the features needed for Task-based creep body definitions, or perhaps help us realize the ways we can extend and customize creep body definitions as we transition away from the Setup queue.

### DIFF
--- a/creep.Setup.js
+++ b/creep.Setup.js
@@ -26,20 +26,21 @@ let Setup = function(typeName){
     this.sortedParts = true;
     this.mixMoveParts = false;
 
+    this.rclProperty = Setup.rclProperty;
     this.SelfOrCall = function(obj, param) {
         if( obj == null ) return null;
         if (typeof obj === 'function' )
             return obj(param);
         else return obj;
     };
-    this.fixedBody = Setup.rclProperty(this, 'fixedBody');
-    this.multiBody = Setup.rclProperty(this, 'multiBody');
-    this.minAbsEnergyAvailable = Setup.rclProperty(this, 'minAbsEnergyAvailable');
-    this.minEnergyAvailable = Setup.rclProperty(this, 'minEnergyAvailable'); // 1 = full
-    this.minMulti = Setup.rclProperty(this, 'minMulti');
-    this.maxMulti = Setup.rclProperty(this, 'maxMulti');
-    this.maxCount = Setup.rclProperty(this, 'maxCount');
-    this.maxWeight = Setup.rclProperty(this, 'maxWeight');
+    this.fixedBody = this.rclProperty('fixedBody');
+    this.multiBody = this.rclProperty('multiBody');
+    this.minAbsEnergyAvailable = this.rclProperty('minAbsEnergyAvailable');
+    this.minEnergyAvailable = this.rclProperty('minEnergyAvailable'); // 1 = full
+    this.minMulti = this.rclProperty('minMulti');
+    this.maxMulti = this.rclProperty('maxMulti');
+    this.maxCount = this.rclProperty('maxCount');
+    this.maxWeight = this.rclProperty('maxWeight');
 
     this.buildParams = function(spawn){
         var memory = {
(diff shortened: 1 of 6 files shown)


karlthepagan:
I need 4 of the 6 requested reviews before this can move forward. I will be putting this into my personal build for a few days before it gets to canary.

@logxen @forkmantis @torin555 @Spedwards @Dasku @chobobobo

Dasku:
19:58:29]TypeError: Cannot read property '8' of undefined
    at rcl (creep.Setup:228:38)
    at Setup.SelfOrCall (creep.Setup:33:20)
    at Setup.isValidSetup (creep.Setup:71:42)
    at probe (spawn:18:26)
    at arraySome (/opt/engine/node_modules/lodash/index.js:1493:13)
    at Function.some (/opt/engine/node_modules/lodash/index.js:7051:14)
    at Spawn.execute (spawn:24:22)
    at run (spawn:109:33)

karlthepagan:
@Dasku fixed with the latest

I forgot that SelfOrCall did not provide `this` - so I fixed it :)

creep.Setup.js
```diff
@@ -78,8 +78,8 @@ let Setup = function(typeName){
             return false;
         }
 
-        let maxCount = this.SelfOrCall(rcl.maxCount, room);
```
karlthepagan: I believe this is the line of code which fixes https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/889

chobobobo: I don't think this will fix that issue, because it is explicitly calling Creep.setup.upgrader.maxCount(room) but the hard limit of 1 upgrader is set in the RCL object.  Since maxCount is overridden in creep.setup.upgrader, the maxCount property set here is not used correct?

creep.Setup.js
```diff
@@ -217,3 +215,24 @@ Setup.maxPerFlag = function(flagFilter, maxRoomRange, measureByHome) {
         return max;
     };
 };
+Setup.rclProperty = function(property) {
+    return function(room) {
+        const creepSetup = this;
+        let rcl;
+        if (typeof creepSetup.RCL === 'function') {
+            rcl = function(room) {
```
chobobobo: We do not currently have any RCL functions do we?  Is this in preparation for future changes?

Dasku:
Not sure if this is related, but just spawned a power train, and its spawned 4 healers and 1 miner, all 4 healers are behaving as they should, however the healer count is only showing as 2 in console.log.

creep.Setup.js
karlthepagan: That does make sense. As we reimplement this we should look at how to provide extendable methods.

creep.Setup.js
karlthepagan: Yes. The RCL object property can't be extended. A function won't have that limitation.

chobobobo:
I'm seeing tiny upgraders in canary at the moment, since the RCL object should have the correct value of 15, is it possible that this change is causing it to use the maxCount function instead of the RCL value as it is supposed to?

chobobobo:
Ok, I'm currently testing the refactor to fix the issue with overridden methods.

creep.Setup.js
```diff
@@ -28,7 +28,7 @@ let Setup = function(typeName){
 
     this.rclProperty = Setup.rclProperty;
     this.SelfOrCall = function(obj, param) {
-        if( obj === null ) return null;
+        if( obj == null ) return null;
```
karlthepagan: I should know better than to have done this.

creep.Setup.js
```diff
@@ -84,9 +84,9 @@ let Setup = function(typeName){
             if (DEBUG && TRACE) trace('Setup', {setupType:this.type, room:room.name, maxCount, maxWeight, Setup:'isValidSetup'}, 'too many creeps');
             return false;
         }
-        if( maxCount === null )
+        if( maxCount == null )
             maxCount = Infinity;
-        if( maxWeight === null )
+        if( maxWeight == null )
```
karlthepagan: upgraders and miners had undefined maxWeight, so yup this fixes it

Spedwards:
@karlthepagan @chobobobo How we looking here?

karlthepagan:
Waiting for release candidate. This changes api behavior

chobobobo:
@karlthepagan with RC branched, are we good to merge now?

PR #896, opened 2017-03-02
added defending strategy for basic target selection customization

karlthepagan:
basic way to parameterize this

also I think the healing filter was broken


karlthepagan:
this has been extremely reliable for me at E75S6, anyone else try it yet?

chobobobo:
Yes, I have been using this and it has been working well, had a 5 invader spawn yesterday that it handled well.

PR #898, opened 2017-03-03
if a viral has an install function then call on load!

karlthepagan:
unfortunately this is required to extend most strategy functions with the conventions I've started implementing

I'll try to move away from it, but it's a necessary evil at the moment


main.js
```diff
@@ -105,6 +105,9 @@ global.load = (modName) => {
         // load viral overrides 
         mod = infect(mod, 'internalViral', modName);
         mod = infect(mod, 'viral', modName);
+        if( mod.install ) {
```
Spedwards: Check if `mod.install` is also a function.

karlthepagan: I'll throw an error explaining the API contract if it's not a function.

Spedwards: Sounds good.

karlthepagan:
Thinking about the object merge feature I'm going to make this function be the place where that feature could be disabled. Changes still coming, I've had a tough week.

Spedwards:
That actually sounds awesome @karlthepagan.

karlthepagan:
Not done, but getting close.

karlthepagan:
I've been overwhelmed by other systematic changes and I'm abandoning this for the foreseeable future.

PR #900, opened 2017-03-03
haulers and miners are vital spawns, and should be easy to spawn.

chobobobo:
RCL8 rooms are fragile because these are set too high.  If a room gets into 'collapsed' and only has the emergency worker to fill extensions, and then both miners die, no haulers can spawn.  With remote creeps still spawning the tiny worker has no real chance of getting energy capacity back up to the level that a hauler or miner could spawn.

Miner was set to 30% full, which is 3870 energy at RCL8
Hauler at 40% is 5160 which is also excessive.

This change sets the limit for miners to 1290 and haulers to 2580.

### DIFF
--- a/creep.setup.hauler.js
+++ b/creep.setup.hauler.js
@@ -54,6 +54,15 @@ setup.default = {
     maxCount: room => setup.maxCount(room),
     maxWeight: room => setup.maxWeight(room),
 };
+setup.high = {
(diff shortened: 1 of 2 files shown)


Perrytheplatypus43:
What is the reason for this?

karlthepagan:
Makes sense. Simple.

PR #904, opened 2017-03-03
Energy cap is 15 energy per tick at RCL8

Dasku:


OverlordQ:
That is what `CONTROLLER_MAX_UPGRADE_PER_TICK / UPGRADE_CONTROLLER_POWER` comes out to.

Dasku:
ahh ok.. I may viral it then as im getting 2 work upgraders and my CP/Hour has dropped to nothing

Dasku:
what is UPGRADE_CONTROLLER_POWER ? and why is it divided? CONTROLLER_MAX_UPGRADE_PER_TICK should be 15 shouldn't it?

logxen:
there is a constant ... therefor we use the constant just in case it changes in the future

OverlordQ:
`UPGRADE_CONTROLLER_POWER` is how much work is done per `WORK` part and how much energy it costs. `CONTROLLER_MAX_UPGRADE_PER_TICK` is max amount of energy you can pump in per tick before it's ignored.

PR #908, opened 2017-03-04
Move our costMatrix caching to memory segments

chobobobo:
Also includes a potential way for us to manage memory segments.  The idea is that if a memory segment is loaded, you can choose how you wish to act on it.  For cost matrices, if the segment loads then the cache gets updated, and the segment currently only gets loaded when the codebase gets loaded.  This populates the cache stored in a global object.

The reasoning behind doing this is that we currently serialize & deserialize the costmatrix each time we use it, which carries a cost, and storing it into normal memory also carries a heavy serialize/deserialize cost.  Now the cost matrices are cached as objects in the global scope.


chobobobo:
I still need to add code to make the transition smooth from the existing cache and clean it up.

ocsMemory.js
```diff
@@ -0,0 +1,56 @@
...
+};
+mod.processSegments = () => {
+    const costMatrixCache = RawMemory.segments[SEGMENTS.COSTMATRIX_CACHE];
+    if (costMatrixCache) {
+    	if (DEBUG) logSystem('RawMemory', 'loading pathfinder cache..');
+        Room.loadCostMatrixCache(JSON.parse(costMatrixCache));
+        mod.deactivateSegment(SEGMENTS.COSTMATRIX_CACHE);
+    }
+};
+mod.numSaved = 0;
```
Spedwards: Keep fields at the top of the file.

global.js
```diff
@@ -178,6 +178,9 @@ mod.LAB_REACTIONS = {
     [RESOURCE_CATALYZED_GHODIUM_ACID]: [RESOURCE_GHODIUM_ACID, RESOURCE_CATALYST],
     [RESOURCE_CATALYZED_GHODIUM_ALKALIDE]: [RESOURCE_GHODIUM_ALKALIDE, RESOURCE_CATALYST],
 };
+mod.SEGMENTS = {
```
Spedwards: Maybe name it `MEMORY_SEGMENTS` or `MEM_SEGMENTS`. Or at the very least add a comment stating that this is for memory segments and not something else.

chobobobo:
There's a potential problem I hadn't considered when implementing this.  We move between cloud instances quite rapidly, rarely the same instance twice, so if I wait a tick to load the cache I am likely loading it into a global on the wrong instance.

chobobobo:
I also need to confirm that when we reload the code the current costMatrix in the segment is the most up to date.

chobobobo:
@Spedwards I believe this meets the changes requested now, when I merged this into traveler I did notice an increase in incomplete paths, that still needs to be debugged before this is ready.

chobobobo:
I have changes for this that make it work, but it won't merge into dev cleanly until traveler is in so it's on hold until traveler( #751 ) is merged.

chobobobo:
Also, I need to store the deserialized cost matrices somewhere else, otherwise the benefit of caching them deserialized gets lost when I have to serialize them all again before they are stored.

chobobobo:
This still has conflicts with #918 once that is merged it can be added to canary.

chobobobo:
Ok, hit the limit for segments faster than I had expected, this should not be put into canary until code to resolve memory limits is implemented.  We also shouldn't be caching empty cost Matrices.

chobobobo:
This needs to be tested when the cache moves into 2 segments before it can be merged into dev

global.js
```diff
@@ -153,6 +153,12 @@ for(let a in REACTIONS){
         mod.LAB_REACTIONS[REACTIONS[a][b]] = [a,b];
     }
 }
+mod.MEM_SEGMENTS = {
+    COSTMATRIX_CACHE: {
+        start: 0,
+        end: 5
```
karlthepagan: would it be possible to start at 99 and work down?

chobobobo:
Yes, I was just thinking about that myself, I will make that change in the morning

chobobobo:
@karlthepagan please have a look, I am quite confident now that this is ready for dev, running it currently and tested transition over 100kb boundary.

PR #910, opened 2017-03-04
Better Creep Birth/Death detection

chobobobo:
Makes our detection more robust, and less prone to errors when we miss a tick for whatever reason.

### DIFF
--- a/population.js
+++ b/population.js
@@ -176,13 +176,13 @@ mod.analyze = function(){
             if( creep.spawning ) { // count spawning time
                 entry.spawningTime++;
             }
-            else if( creep.ticksToLive > 0 && !creep.memory.spawned ){ // spawning complete
-                creep.memory.spawned = true;
+            else if( creep.ticksToLive > 0 && !creep.data.spawned ){ // spawning complete
+                creep.data.spawned = true;
                 this.spawned.push(entry.creepName);
                 if (Game.spawns[entry.motherSpawn]) this.spawnsToProbe.push(entry.motherSpawn); // only push living spawns
             }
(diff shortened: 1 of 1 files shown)


population.js
```diff
@@ -176,17 +176,19 @@ mod.analyze = function(){
             if( creep.spawning ) { // count spawning time
                 entry.spawningTime++;
             }
-            else if( creep.ticksToLive ==  ( creep.data.body.claim !== undefined ? 499 : 1499 ) ){ // spawning complete
+            else if( creep.ticksToLive > 0 && !creep.memory.spawned ){ // spawning complete
+                creep.memory.spawned = true;
```
Spedwards: Use `creep.data` instead. We don't use `creep.memory` for anything (for reasons beyond me).

population.js
```diff
@@ -176,17 +176,19 @@ mod.analyze = function(){
...
                 this.spawned.push(entry.creepName);
                 if (Game.spawns[entry.motherSpawn]) this.spawnsToProbe.push(entry.motherSpawn); // only push living spawns
             }
-            else if(creep.ticksToLive == ( entry.predictedRenewal ? entry.predictedRenewal : entry.spawningTime)) { // will die in ticks equal to spawning time or custom
+            else if(creep.ticksToLive <= ( entry.predictedRenewal ? entry.predictedRenewal : entry.spawningTime) && !creep.memory.nearDeath) { // will die in ticks equal to spawning time or custom
+                creep.memory.nearDeath = true;
```
Spedwards: Here too.

chobobobo:
@Spedwards I have made the requested changes, how does it look now?

PR #911, opened 2017-03-04
Upgraders should not repair, miners have their own checks.

chobobobo:
I think upgraders should focus on upgrading, this becomes especially important at RCL8 when you want to spend a majority of your ticks upgrading.

This also removes the drive by repair from miners as they have their own means of repairing.

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -56,7 +56,9 @@ mod.extend = function(){
                     return;
                 }
             }
-            this.repairNearby();
+            if (this.data && !_.contains(['remoteMiner', 'miner', 'upgrader'], this.data.creepType)) {
+                this.repairNearby();
+            }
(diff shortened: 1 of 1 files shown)


creep.js
```diff
@@ -56,7 +56,9 @@ mod.extend = function(){
                     return;
                 }
             }
-            this.repairNearby();
+            if (this.data && !_.contains(['remoteMiner', 'miner', 'upgrader'], this.data.creepType)) {
+                this.repairNearby();
```
karlthepagan: I'm going to make this a strategy as an example.

chobobobo: ok, I will hold off on merging this then.

karlthepagan: well, after next release... I think this is important enough to let it go

PR #914, opened 2017-03-04
Fix Visuals Undefined Error

### LINKED ISSUE #913: Visuals
TypeError: Cannot read property 'run' of undefined
    at Object.module.exports.loop (main:290:75)
    at __mainLoop:1:52

visuals do not work

Spedwards:
Resolves #913


Perrytheplatypus43:
will the visuals stop showing when the the CPU bucket is critical?

karlthepagan:
yes, and code loading makes there be lag when they come back

Perrytheplatypus43:
@karlthepagan ok thanks

PR #918, opened 2017-03-05
Feature/profiling

chobobobo:
Basic profiling information, if PROFILE is true it gives you average creep & flag usage per tick.  If DEBUG is true it gives you just average usage in and out of the runloop as well as your current bucket.  Otherwise it defaults to noop functions so consumes no CPU.

Testing between enabled/disabled revealed between 1-2% overhead while enabled.

To reset the profiling data type resetProfiling() into the console.


Spedwards:
To convert the entire profile to a HTML table, store each check into an array, only storing what you need (identifier, size, etc.).

Then display it the same as you would the memory method by iterating through the array at the bottom of the tick.

Storing it in an array also makes sorting it by the size easy.

global.js
```diff
@@ -392,4 +392,97 @@ mod.guid = function(){
...
+            totalCPU: 0,
+            totalTicks: 0,
+            types: {},
+            validTick: Game.time
+        };
+    }
+    mod.profiler = Memory.profiler;
+};
+mod.startProfiling = function(name, startCPU) {
+    let checkCPU = function(localName, limit, type) {};
```
karlthepagan: constants for type please

you can define them inside this module as to not pollute global:

`p.ANALYZE_CPU, p.EXECUTE_CPU` etc

chobobobo: type represents creep/flag type in this case and is used for avg usage/type.  Do you mean that you would like constants in parameters.js to allow people to turn on/off certain aspects of the profiling?

karlthepagan: No, definitely not in parameters.

I would like it to be expressive so that someone reading main.js has a chance of getting the gist of what's going on.

Either each type should have its own method or each type should be a constant or each type a string instead of a number (an opaque / not expressive value)

karlthepagan:
ypeError: Cannot read property 'undefined' of undefined
    at Object.checkCPU (global:450:57)
    at triggerFound (flagDir:175:15)
    at Array.forEach (native)
    at Object.mod.execute (flagDir:178:15)
    at Object.module.exports.loop (main:281:13)

TypeError: Cannot read property 'worker' of undefined
    at Object.checkCPU (global:450:57)
    at Creep.run (creep:117:15)
    at run (creep:476:30)

looking at a fix

global.js
```diff
@@ -392,4 +392,97 @@ mod.guid = function(){
...
+    if (reset) {
+        logSystem('Profiler', 'resetting profiler data.');
+        Memory.profiler = {
+            totalCPU: 0,
+            totalTicks: 0,
+            types: {},
+            validTick: Game.time
+        };
+    }
+    mod.profiler = Memory.profiler;
```
karlthepagan: this will resist viral extension - e.g. viral.global.js

IMO profiler could be injected into the global, but why not just keep referring to Memory.profiler?

chobobobo: The idea was to minimize any overhead of enabling the profiler, and to avoid using memory so that we could profile the deserialization costs but I think you're right that it adds unnecessary complexity.. we could just do the comparison before we touch memory the first time.

karlthepagan:
I'm going to get this merged. I'm hitting strange CPU overflows and want to start reintroducing my bailout mechanism (stop executing at low CPU and record last execution step for logging).

PR #933, opened 2017-03-06
STRUCTURE_POWERSPAWN is undefined

karlthepagan:
such a simple change, but nope... that constant does not exist


logxen:
it should be STRUCTURE_POWER_SPAWN

PR #939, opened 2017-03-06
controllers at range 4 from links or containers 

karlthepagan:
these are possible managed energy structures

There are also cases where these should *not* be managed structures, but I feel that the logic needs some work to be perfect.

For now I'm making these on by default, and I might add a property definition which saves a memory setting to add or exclude these from the list of controller structures.

### DIFF
--- a/room.js
+++ b/room.js
@@ -1085,13 +1085,15 @@ mod.extend = function(){
             structure => structure.structureType == STRUCTURE_CONTAINER
         );
         let add = (cont) => {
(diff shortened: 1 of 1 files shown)


chobobobo:
I also made a change to avoid setting .controller in unowned rooms see #926 perhaps these should be merged into one.

chobobobo:
I have merged this into #926

karlthepagan:
:+1: let there be one branch!

PR #940, opened 2017-03-07
Change flag colour reservation

Dasku:
as per Karls request for auto building


karlthepagan:
holding change for 7.0

PR #942, opened 2017-03-07
storage debounce improved - hold withdrawing / storing until an output action target is available

karlthepagan:
This is very similar to your implementation @logxen but I wanted to clean it up and make room for a standard impl that uses our APIs.

### DIFF
--- a/creep.action.feeding.js
+++ b/creep.action.feeding.js
@@ -15,6 +15,9 @@ action.isAddableTarget = function(target){
         (!target.targetOf || _.filter(target.targetOf, {'actionName':'feeding'}).length < this.maxPerTarget));
 };
 action.newTarget = function(creep){
+    if (creep.room.energyAvailable === creep.room.energyCapacityAvailable) {
+        return null;
+    }
     var that = this;
     return creep.pos.findClosestByPath(creep.room.structures.all, {
         filter: (structure) => {
--- a/creep.action.withdrawing.js
+++ b/creep.action.withdrawing.js
@@ -22,3 +22,18 @@ action.onAssignment = function(creep, target) {
     //if( SAY_ASSIGNMENT ) creep.say(String.fromCharCode(9738), SAY_PUBLIC);
     if( SAY_ASSIGNMENT ) creep.say('\u{1F4E4}\u{FE0E}', SAY_PUBLIC);
 };
+action.debounce = function(creep, outflowActions, callback, thisArg) {
+    let shouldCall = false;
+    if (creep.data.lastAction === 'storing' && creep.data.lastTarget === creep.room.storage.id) {
+        // cycle detected
+        shouldCall = _.chain(outflowActions).invoke('newTarget', creep).some().value(); // cycle broken if valid target
+    } else {
+        shouldCall = true;
+    }
+
+    if (shouldCall) {
+        return _.invoke([thisArg], callback, this)[0];
+    }
+
+    return undefined;
+};
--- a/creep.behaviour.hauler.js
+++ b/creep.behaviour.hauler.js
@@ -20,28 +20,27 @@ mod.nextAction = function(creep){
         Creep.action.travelling.assign(creep, Game.rooms[creep.data.homeRoom].controller);
         return;
     }
-    let priority;
-    if( creep.sum < creep.carryCapacity/2 ) {
+    const outflowPriority = [
+        Creep.action.feeding,
+        Creep.action.charging,
+        Creep.action.fueling,
+    ];
+    let priority = outflowPriority;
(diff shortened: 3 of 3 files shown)


creep.action.feeding.js
```diff
@@ -15,6 +15,9 @@ action.isAddableTarget = function(target){
         (!target.targetOf || _.filter(target.targetOf, {'actionName':'feeding'}).length < this.maxPerTarget));
 };
 action.newTarget = function(creep){
+    if (creep.room.energyAvailable === creep.room.energyCapacityAvailable) {
```
karlthepagan: This is a CPU savings.

creep.action.withdrawing.js
```diff
@@ -22,3 +22,18 @@ action.onAssignment = function(creep, target) {
     //if( SAY_ASSIGNMENT ) creep.say(String.fromCharCode(9738), SAY_PUBLIC);
     if( SAY_ASSIGNMENT ) creep.say('\u{1F4E4}\u{FE0E}', SAY_PUBLIC);
 };
+action.debounce = function(creep, outflowActions, callback, thisArg) {
+    let shouldCall = false;
+    if (creep.data.lastAction === 'storing' && creep.data.lastTarget === creep.room.storage.id) {
+        // cycle detected
+        shouldCall = _.chain(outflowActions).invoke('newTarget', creep).some().value(); // cycle broken if valid target
```
karlthepagan: This line is specific to withdrawing, but we could parameterize it.

logxen:
@karlthepagan alternatively I can reverse things back to having the screeps idle full of energy ... and then just give reallocating the ability to break out of idle. having them idle empty requires that all three of feed charge and fuel need to be able to break out with the "pretargeting". I can also follow through with the reallocating side and make it remember the potential target that it found which I need to do anyway.

karlthepagan:
Feel free to use this as the basis for your fix. I would like it to be expressive and I believe the debounce function achieves that.

logxen:
sounds good I didn't care for my implementation too much. :P

creep.action.feeding.js
chobobobo: :tada:

creep.behaviour.hauler.js
```diff
@@ -20,28 +20,27 @@ mod.nextAction = function(creep){
...
     }
-    let priority;
-    if( creep.sum < creep.carryCapacity/2 ) {
+    const outflowPriority = [
+        Creep.action.feeding,
+        Creep.action.charging,
+        Creep.action.fueling,
+    ];
+    let priority = outflowPriority;
+    if( creep.sum * 2 < creep.carryCapacity ) {
```
chobobobo: The intention here is to check if the current carry is more than half o the creep's carry right?  I find it clearer to leave it as carryCapacity/2 as it reads closer to the intention.

creep.behaviour.hauler.js
```diff
@@ -20,28 +20,27 @@ mod.nextAction = function(creep){
...
-        let potentialFuelTarget = Creep.action.fueling.newTarget(creep);
-        let potentialChargeTarget = Creep.action.charging.newTarget(creep);
-        if( creep.data.lastAction !== 'storing' || !creep.room.storage || creep.data.lastTarget !== creep.room.storage.id ||
-            creep.room.relativeEnergyAvailable < 1.0 || potentialFuelTarget !== null ||
-                ( potentialChargeTarget && ( (potentialChargeTarget.structureType === STRUCTURE_LINK) ? true : ( potentialChargeTarget.storeCapacity - potentialChargeTarget.sum ) >  Math.min( creep.carryCapacity, 500 ) ) ) ) {
-            priority.push(Creep.action.withdrawing);
-        }
+            Creep.action.picking,
+        ];
+        Creep.action.withdrawing.debounce(creep, outflowPriority, function(withdrawing) {
```
chobobobo: Can you add a comment here to the effect of 'only withdraw if we have an outflow task' to make it clear what is happening here for a first time reader

chobobobo:
@logxen if this looks good to you, feel free to merge it in.

creep.action.withdrawing.js
chobobobo: I had to change this line to  
```shouldCall = _.some(outflowActions, a => a.isValidAction(creep) && a.isAddableAction(creep) && a.newTarget(creep));```
because it was thinking it could charge a managed container when it failed lower in behaviour.hauler.nextAction

logxen: you want me to replace line 29?

logxen: ok that's taken care of. :)

PR #946, opened 2017-03-09
Improve needsDebounce

chobobobo:
Add debounce to workers, fix debounce checking, don’t try the same action more than once.

Didn't want to jump into your branch without speaking to you, but this has resolved the storage cycling and has been running well for me.


creep.action.withdrawing.js
```diff
@@ -26,7 +26,7 @@ action.debounce = function(creep, outflowActions, callback, thisArg) {
     let shouldCall = false;
     if (creep.data.lastAction === 'storing' && creep.data.lastTarget === creep.room.storage.id) {
         // cycle detected
-        shouldCall = _.chain(outflowActions).invoke('newTarget', creep).some().value(); // cycle broken if valid target
+        shouldCall = _.some(outflowActions, a => a.isValidAction(creep) && a.isAddableAction(creep) && a.newTarget(creep));
```
karlthepagan: isValidAction fails for any action that relies on the creep state - e.g. creep.sum < creep.carryCapacity

that's why I skipped it initially and just used newTarget

creep.behaviour.worker.js
```diff
@@ -21,64 +21,50 @@ mod.nextAction = function(creep){
...
+        Creep.action.fueling,
+        Creep.action.fortifying,
+        Creep.action.charging,
+        Creep.action.upgrading
+    ];
+    if (creep.data.creepType === "pioneer") {
+        outflowPriority.unshift(Creep.action.building); // let's get that base built!
+        if (creep.room.controller && creep.room.controller.level < 2) // build up the room to RCL2 quickly
+            outflowPriority.unshift(Creep.action.upgrading);
+        outflowPriority.unshift(Creep.action.feeding); // highest priority so we don't spend our time repairing existing walls
```
karlthepagan: having the entire pioneer list expressed is a lot of repeating yourself, but it points to the chance we have to make outflowPriority into a strategy function

it can stay this way for now, but some day strategy handler!

creep.action.withdrawing.js
chobobobo: Ah yes, well the only time I saw this failing before was when haulers selected a nearly full managed container as a charging target but then couldn't add the target in the next tick.  I will instead look at whether a fix can be focused there instead

chobobobo: Ah yes, well the only time I saw this failing before was when haulers selected a nearly full managed container as a charging target but then couldn't add the target in the next tick.  I will instead look at whether a fix can be focused there.

creep.behaviour.worker.js
chobobobo: Yeah, but keeping the list made adding debounce much messier.  I can revert it to put the if statements deciding on outflow priority, or I could add a TODO :)

karlthepagan: I think @logxen has integrated these changes so the point is moot for now ;)

chobobobo:
Merged, so closed.

PR #951, opened 2017-03-10
Allow Viral Tasks

Spedwards:
Now to add a viral task, in addition to previous method, call `Task.addTasks(youTask)`.


task.js
```diff
@@ -1,42 +1,21 @@
 let mod = {};
 module.exports = mod;
+mod.tasks = [];
+mod.populate = function() {
+    Task.addTasks(Task.guard, Task.defense, Task.claim, Task.reserve, Task.mining, Task.pioneer, Task.attackController, Task.robbing, Task.reputation);
```
karlthepagan: one line per task for easier merges

also please alphabetize the task names

Spedwards: Sure thing.

Dasku:
yey!!!

task.js
```diff
@@ -2,15 +2,17 @@ let mod = {};
...
-    Task.addTasks(Task.attackController);
-    Task.addTasks(Task.claim);
-    Task.addTasks(Task.defense);
-    Task.addTasks(Task.guard);
-    Task.addTasks(Task.mining);
-    Task.addTasks(Task.pioneer);
-    Task.addTasks(Task.reputation);
-    Task.addTasks(Task.reserve);
-    Task.addTasks(Task.robbing);
+    Task.addTasks(...[
```
Spedwards: An array so you can have a following comma on the last item, making for easier merges.

task.js
```diff
@@ -1,42 +1,31 @@
...
-    }
+        Task.reserve,
+        Task.robbing,
+    ]);
+};
+mod.addTasks = (...task) => Task.tasks.push(...task);
+// load task memory & flush caches
+mod.flush = function () {
+    Task.tasks.forEach(task => {
+        if (task.flush) task.flush();
```
chobobobo: Do we want to make sure this is actually a function as well while we're at it?

task.js
```diff
@@ -1,42 +1,31 @@
...
-        Task.robbing,
-        Task.reputation,
-    ];
-    for (let i = tasks.length - 1; i >= 0; i--) {
-        if (tasks[i].register) {
-            tasks[i].register();
-        }
-    }
+    Task.tasks.forEach(task => {
+        if (task.register) task.register();
```
chobobobo: as above.

task.js
karlthepagan: I'm ambivalent about strictly checking the type like that.

task.js
karlthepagan: :+1:

PR #952, opened 2017-03-11
avoiding should not return a non-attack creep from newTarget

karlthepagan:
quick fix


creep.action.avoiding.js
```diff
@@ -33,7 +33,9 @@ action.newTarget = function(creep) {
     }
 
     if (creep.room.situation.invasion) {
-        const target = _.chain(creep.room.hostiles).map(function(target) {
+        const target = _.chain(creep.room.hostiles).filter(function(target) {
+            return action.isValidTarget(target);
```
Spedwards: Just realised, this could have been `filter(action.isValidTarget)` instead.

karlthepagan: Ah I think you're right in this case.

A few times I was burned by doing function references buy I realize that was declarative and this is in a function scope.

Today I learned something.

PR #955, opened 2017-03-11
[WIP] Mobile Storage Creep

Spedwards:
Mobile Storage Creep. Currently a WIP.

Waiting for the viral tasks PR to get merged so I can add this task to the list.

TODO:

 - Delivering Action: A creep will request a resource, and the mobile storage will move to that creep and transfer to it. Ideally, taking the distance into account.
 - Storage Action: Similar to guarding action. Idles at the flag placed.
 - Add Mobile Storage as a valid target to reallocate energy.

### DIFF
--- a/global.js
+++ b/global.js
@@ -62,6 +62,10 @@ mod.FLAG_COLOR = {
         }
     },
     //COLOR_PURPLE,
+    economy: {
+        color: COLOR_PURPLE,
+        secondaryColor: COLOR_PURPLE,
+    },
     //COLOR_BLUE,
     //COLOR_CYAN - Reserved
     //COLOR_GREEN
--- a/task.mobileStorage.js
+++ b/task.mobileStorage.js
@@ -1,0 +1,141 @@
+const mod = {};
+module.exports = mod;
+
+mod.register = () => {
+    Flag.found.on(Task.mobileStorage.handleFlagFound);
+    
+    Creep.spawningStarted.on(Task.mobileStorage.handleSpawningStarted);
+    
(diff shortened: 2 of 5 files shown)


global.js
```diff
@@ -62,6 +62,10 @@ mod.FLAG_COLOR = {
         }
     },
     //COLOR_PURPLE,
+    economy: {
+        color: COLOR_PURPLE,
+        secondaryColor: COLOR_PURPLE,
+    },
```
karlthepagan: white/yellow flag exists and is suitable for this purpose

Spedwards: Yeah, the Purple/Purple was more of a placeholder for the time being.

task.mobileStorage.js
```diff
@@ -0,0 +1,141 @@
...
+            spawning: [],
+            running: [],
+        };
+    }
+    return flag.memory.tasks.mobileStorage;
+};
+
+mod.creep = {
+    fixedBody: [CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,CARRY,MOVE,MOVE,MOVE,MOVE,MOVE,MOVE,MOVE],
+    multiBody: [CARRY,CARRY,MOVE],
```
karlthepagan: my previous thinking on creep design was 1 MOVE and the rest carry

however, this is highly dependent on the actions

Perrytheplatypus43:
What will this be used for, will it be used as a storage for remote mining rooms?

Spedwards:
@Perrytheplatypus43 Uses could be for anything. The task is mostly a basis for viral tasks. I'm sure we can come up with an idea for uses in public.

PR #957, opened 2017-03-11
Change/rebuild ally

karlthepagan:
p.s. my spawns are gone and I need a rebuild

### DIFF
--- a/room.js
+++ b/room.js
@@ -537,7 +537,15 @@ mod.extend = function(){
                 return this._constructionSites;
             }
         },
-
+        'myConstructionSites': {
+            configurable: true,
+            get: function() {
+                if( _.isUndefined(this._myConstructionSites) ) {
+                    this._myConstructionSites = this.find(FIND_MY_CONSTRUCTION_SITES);
+                }
+                return this._myConstructionSites;
+            }
+        },
         'creeps': {
             configurable: true,
             get: function() {
@@ -1991,6 +1999,7 @@ mod.flush = function(){
         if( global.isNewServer ) {
             delete room._my;
             delete room._constructionSites;
+            delete room._myConstructionSites;
             delete room._maxPerJob;
             delete room._minerals;
             delete room._structures;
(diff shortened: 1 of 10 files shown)


room.js
```diff
@@ -532,7 +532,7 @@ mod.extend = function(){
             configurable: true,
             get: function() {
                 if( _.isUndefined(this._constructionSites) ) {
-                    this._constructionSites = this.find(FIND_MY_CONSTRUCTION_SITES);
+                    this._constructionSites = this.find(FIND_CONSTRUCTION_SITES);
```
chobobobo: Need to consider any possible side-effects in the code from this change, eg. If I am taking over a room with buildings left over from someone else, will my workers start building those buildings?

task.reputation.js
```diff
@@ -12,7 +12,7 @@ const CONST = {
 
 const mod = {
     name: 'reputation',
-    myName: () => _.values(Game.spawns)[0].owner.username,
+    myName: () => _.chain(Game.rooms).map('controller').flatten().filter('my').map('owner.username').first().value(),
```
chobobobo: This change is already in #956 will that cause conflicts?

room.js
karlthepagan: I agree. This is problematic.

karlthepagan: Possibly resolved by the ally checking in building action.

task.reputation.js
karlthepagan: No, same branch because I'm taking a break and my dev discipline broke down.

room.js
chobobobo: I think perhaps adding myConstructionSites which is simply a filter to this.constructionSites would work.. I noticed that miner just does a check on constructionSites, upgrader looks at constructionSites.length, worker spawns are based on the same, getBestConstructionSiteFor also looks at them all.. perhaps a good change here would be to simply take FIND_CONSTRUCTION_SITES and filter for ally owner?

chobobobo:
Ok, added room.myConstructionSites and converted all existing usage.

chobobobo:
This was merged into major/traveler to get some further testing in canary given it will be merged soon.

PR #962, opened 2017-03-13
haulers should fill the closest towers first.

chobobobo:
At some point we should update fueling so that they fill towers based on their proximity to hostiles, but this stops them from walking across your room to fill 1 tower and making the invasion even worse.


karlthepagan:
Some day I'll add this to strategies... but NOT TODAY!

PR #963, opened 2017-03-14
energy bar can be over 100%

karlthepagan:
quick visuals fix, already applied to rc1


Spedwards:
Wait. How can it be over 100%?

karlthepagan:
it's over 100% when the room has downgraded

Spedwards:
Wouldn't have thought that non-active extensions would count.

karlthepagan:
Also if I remember from taking over a dead room, un-owned extensions are counted but you can't use that energy.

PR #968, opened 2017-03-14
Task/lab tech new

logxen:
This implements Task.labTech which responds to a purple/white flag by spawning a specialized hauler that prioritizes action.reallocating to serve the local room.


creep.behaviour.labTech.js
```diff
@@ -0,0 +1,68 @@
...
+    let priority = outflowPriority;
+    if( creep.sum * 2 < creep.carryCapacity ) {
+        priority = [
+            Creep.action.uncharging,
+            Creep.action.picking,
+        ];
+        Creep.action.withdrawing.debounce(creep, outflowPriority, function(withdrawing) {
+            priority.push(withdrawing);
+        });
+        priority.push(Creep.action.reallocating);
```
Dasku: This won't work will it? The reason I'd made a new behaviour for labTech was to ensure reallocating was always the first chosen action (using the old priority action list). If we just use current hauler logic it wouldn't act any different.

chobobobo:
What's the best way to test this feature?

Dasku:
drop a flag in any room you want to fill powerSpawn, labs or empty terminals. I drop a flag in most RCL6+ rooms (setStore energy 0 on terminal so high rooms send lots and labTech empties for faster upgrading), and at 7-8 I add lab work and powerSpawns too

chobobobo:
purple/white flag, and drop it anywhere, or on the labs?

chobobobo:
I think this is a perfect use case for behaviours, from what I can see a labTech is essentially just a hauler with different priorities.  If the priorities of the hauler were defined by a behaviour, we could override it with the labtech's priorities and we wouldn't need to copy all of the hauler code into labTech.  Having two copies of the same code will introduce errors over the long term as changes are made to haulers but not reflected in the labtech code.

chobobobo:
Also, I have idling labTech's when my powerSpawns are full of power but have no energy, not sure why.

chobobobo:
I've pushed a temporary fix for the idling labTechs

Dasku:
> Also, I have idling labTechs when my powerSpawns are full of power but have no energy, not sure why.

Does your room have enough energy? It wont take it if below 50% between your MIN/MAX energy store

chobobobo:
Yes, there were a few issues, my commits should have resolved them.  The lab techs are now happily draining my rooms to 50% store.

Dasku:
found a couple labTech's idling at storage with 6-150 energy on them when they should be taking power to the spawns. Couldn't work out if its in reallocate or not. Its like having energy on them is stopping them getting a new action, even thou they are less than half capacity so should be doing something.

chobobobo:
Yeah it's action.reallocating isAddableAction currently they have to be empty, but they also end up idling if you remove that so I don't have a fix yet

PR #971, opened 2017-03-15
Moving extractor up in priority

Perrytheplatypus43:
Moving the extractor up in priority in be tween road and container.


Spedwards:
Might I ask the reason for this? It seems like something more for virals.

Perrytheplatypus43:
When i was trying to build my extractor it took forever so thing i should be built faster.

Spedwards:
I'm going to have to disagree with this then. Here are my reasons:

1. Roads get built first to serve creeps walking back and forth to the mineral while building the extractor and container.
2. Container is built second so it's ready as soon as the extractor done.
3. Extractor is one of the lowest priority items in the game. Sure there's only one per room so getting it over and done with is a valid argument, however I don't agree with moving it up.

As stated in my first comment, I'd recommend you move this to a viral.

Perrytheplatypus43:
I thing thoughs are great arguments I will close this pull reqest

PR #982, opened 2017-03-16
Move Traveler from memory to data property

Spedwards:
Before traveler we weren't using the memory property.

Not a fan of inconsistencies.

### DIFF
--- a/visuals.js
+++ b/visuals.js
@@ -477,12 +477,12 @@ module.exports = class Visuals {
         const vis = new RoomVisual(room.name);
         room.creeps.forEach(creep => {
             if (creep.data._travel && creep.data._travel.path) {
-                const path = creep.memory._travel.path.substr(1);
+                const path = creep.data._travel.path.substr(1);
                 const style = Visuals.creepPathStyle(creep);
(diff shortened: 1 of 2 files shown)


chobobobo:
Also not a fan, this will depend on @karlthepagan though as he wants to minimize changes to traveler proper.

karlthepagan:
The location of Traveler.js storage is a contract of that API. Users who migrate from an OCS branch to another AI should respect each other's pathing information if it is available.

I am willing to entertain an accessor function to make integration for visuals simpler, but not a change in Traveler.js functionality.

visuals.js
```diff
@@ -476,13 +476,13 @@ module.exports = class Visuals {
     static drawCreepPath(room) {
         const vis = new RoomVisual(room.name);
         room.creeps.forEach(creep => {
-            if (creep.memory && creep.memory._travel && creep.memory._travel.path) {
```
karlthepagan: I would prefer an accessor utility method instead of this refactor.

chobobobo:
@Spedwards do you intend to resolve these changes for the 0.7.0 RC?

PR #985, opened 2017-03-17
Observer Invalid Room Fix

Spedwards:
Previous code didn't cause an error. Just had the observer do nothing that tick.


room.js
```diff
@@ -2203,7 +2203,10 @@ mod.extend = function(){
         } while (Memory.observerSchedule.includes(nextRoom) || nextRoom in Game.rooms);
         this.memory.observer.lastLookedIndex = lastLookedIndex;
         Memory.observerSchedule.push(nextRoom);
-        OBSERVER.observeRoom(nextRoom); // now we get to observe a room
+        const r = OBSERVER.observeRoom(nextRoom); // now we get to observe a room
+        if (r === ERR_INVALID_ARGS) { // room has not yet been created / off the map
+            Memory.observerSchedule.splice(Memory.observerSchedule.indexOf(nextRoom), 1); // remove invalid room from list
```
chobobobo: If I understand the code correctly, this will remove an invalid entry from the schedule, is it not possible to prevent putting an invalid entry in in the first place?  Also, this will still skip a tick for that observer correct?  Do we want the observer to try observing the next room in the schedule?

Spedwards: Unfortunately there's no other way to check if the room actually exists. What this does is remove the room from the list so on the next loop, it won't look at it.

And after thinking about it, I may be able to get it to look at the next room.

karlthepagan: There's this: http://support.screeps.com/hc/en-us/articles/203079191-Map#isRoomAvailable

However this is an improvement. Make a new PR for that feedback. :)

PR #997, opened 2017-03-18
attack controllers with gusto!

chobobobo:


task.attackController.js
```diff
@@ -165,8 +165,10 @@ mod.nextAction = creep => {
 };
 mod.creep = {
     attackController: {
-        fixedBody: [CLAIM, MOVE,CLAIM, MOVE,CLAIM, MOVE,CLAIM, MOVE,CLAIM, MOVE],
-        multiBody: [],
+        fixedBody: [],
+        multiBody: [CLAIM, MOVE,CLAIM, MOVE,CLAIM, MOVE,CLAIM, MOVE,CLAIM, MOVE],
```
Spedwards: Keep fixed body the same, set multi body to `[CLAIM, MOVE]`. That way it should automatically adjust.

chobobobo: It has to be multiples of 5 claim parts to attack controllers.

Spedwards: Oh my bad. I thought it was a minimum of 5 parts.

PR #1003, opened 2017-03-18
attempt to check for circular references before the backend find them

karlthepagan:
Guess what... this one will throw an exception. BECAUSE we want to halt execution before the engine can mess us up.

Not sure what I should do to repair these circulars until I see one. I might be able to delete the keys after we gather a few samples.

This is disabled by default. In order to enable it you MUST add the following code to your viral.mainInjection.js
```
mod.execute = function(){
    Population.findCircular();
};
```

### DIFF
--- a/population.js
+++ b/population.js
@@ -271,6 +271,7 @@ mod.execute = function(){
 mod.cleanup = function(){
     let unregister = name => Population.unregisterCreep(name);
     this.died.forEach(unregister);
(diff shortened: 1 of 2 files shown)


karlthepagan:
p.s. because this is disabled by default it should be good to merge any time. No, I don't want a config to ask people to enable this. I want anyone debugging it to understand virals ;)

PR #1012, opened 2017-03-19
closed borders diamond iterator

karlthepagan:
this does what it says, try it!

If you would like to add the an open-border (dotted?) strictly diamond shape to this code please do so. I'll be using it for pathfinding so it had to be closed.

### DIFF
--- a/util.diamond.iterator.js
+++ b/util.diamond.iterator.js
@@ -1,0 +1,71 @@
+/**
+ * A simple diamond pattern iterator which ends after one loop.
+ *
+ * Test case: [...load("util.diamond.iterator").loop({x:19,y:7},3)].map(i=>
+ *               Game.rooms.W29N19.createConstructionSite(i.x,i.y,STRUCTURE_ROAD))
+ */
+class DiamondIterator {
+    // public y;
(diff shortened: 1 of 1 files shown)


karlthepagan:
image

demo of this code: `[...load("util.diamond.iterator").loop({x:19,y:7},3)].map(i=>Game.rooms.W29N19.createConstructionSite(i.x,i.y,STRUCTURE_ROAD))`

karlthepagan:
This has a bug. Resolving.

karlthepagan:
turns out something else I'm doing on the loop uses a lot of CPU... so this is ok

karlthepagan:
This algorithm has been pretty stable, but I do have an optimization that reduces the number of adds per loop to 4.

chobobobo:
In the process of writing the questions I think I have come to understand this. The intention here is just to create a loop which produces a diamond in room positions of the specified radius?  Open borders is the diagonal traversal of the outer edge, where closed is the one you have which consists of only vertical and horizontal movements.  Is that an accurate assessment?

karlthepagan:
You got it!

PR #1013, opened 2017-03-19
Observer Simplification

Spedwards:


chobobobo:
reduced the cost of this from ~10CPU/room to <= 2CPU/room

chobobobo:
@karlthepagan this deprecates OBSERVER_PRIORITISE_HIGHWAY but I don't know that anyone would have been using it, otherwise I think it's ready to merge now.

PR #1014, opened 2017-03-19
Bulldoze Action

Spedwards:
Really hope I've done this right.

### DIFF
--- a/creep.action.bulldozing.js
+++ b/creep.action.bulldozing.js
@@ -1,0 +1,29 @@
+const action = new Creep.Action('bulldozing');
+module.exports = action;
+action.isValidAction = function(creep) {
+    return true;
+};
+action.isValidTarget = function(target) {
+    return target instanceof ConstructionSite && Task.reputation.notAlly(target.owner.username);
+};
+action.newTarget = function(creep) {
+    return _(creep.room.constructionSites)
+        .filter(action.isValidTarget)
+        .max(target => {
+            let rating;
+            if (target.structureType === STRUCTURE_SPAWN) {
+                rating = 20000;
+            } else {
+                rating = 10000;
+            }
+            rating += target.progress / target.progressTotal * 10000;
+            rating -= creep.pos.getRangeTo(target);
+            return rating;
+        });
+};
+action.work = function(creep) {
+    return creep.travelTo(creep.target);
+};
+action.onAssignment = function(creep) {
+    if (SAY_ASSIGNMENT) creep.say(String.fromCodePoint(0x1F4CC), SAY_PUBLIC);
+};
--- a/creep.action.bulldozing.js
+++ b/creep.action.bulldozing.js
@@ -1,8 +1,5 @@
 const action = new Creep.Action('bulldozing');
 module.exports = action;
-action.isValidAction = function(creep) {
-    return true;
-};
 action.isValidTarget = function(target) {
     return target instanceof ConstructionSite && Task.reputation.notAlly(target.owner.username);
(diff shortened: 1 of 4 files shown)


creep.action.bulldozing.js
```diff
@@ -0,0 +1,29 @@
+const action = new Creep.Action('bulldozing');
+module.exports = action;
+action.isValidAction = function(creep) {
+    return true;
+};
+action.isValidTarget = function(target) {
+    return target instanceof ConstructionSite && Task.reputation.notAlly(target.owner.username);
```
karlthepagan: reputation.hostile is also a good choice: that will let you not bulldoze people in the 1-199 reputation range

creep.action.bulldozing.js
```diff
@@ -0,0 +1,29 @@
...
+};
+action.newTarget = function(creep) {
+    return _(creep.room.constructionSites)
```
karlthepagan: does implicit chaining ever return a falsey value? I'm still not sure

creep.action.bulldozing.js
```diff
@@ -0,0 +1,29 @@
+const action = new Creep.Action('bulldozing');
+module.exports = action;
```
chobobobo: if you add action.renewTarget=true it will automatically select new targets for you

creep.action.bulldozing.js
```diff
@@ -0,0 +1,29 @@
...
+            } else {
+                rating = 10000;
+            }
+            rating += target.progress / target.progressTotal * 10000;
+            rating -= creep.pos.getRangeTo(target);
+            return rating;
+        });
+};
+action.work = function(creep) {
+    return creep.travelTo(creep.target);
```
chobobobo: I think this can just be a creep.move(creep.pos.getDirectionTo(creep.target)) since you will be within range 1 of the target at this point.

creep.action.bulldozing.js
chobobobo: oh, the default is true.

creep.action.bulldozing.js
Spedwards: Not necessarily? Couldn't you be on the other side of the room? If so, walls would get in the way.

creep.action.bulldozing.js
Spedwards: Since I end with `max`, the only falsey value it can return is `-Infinity`, which wouldn't pass the `isValidTarget` check.

karlthepagan: Ok I did some console testing, seems fine:
```
_([0,1,2,3,4]).reject(i=>i%2===1).first()
 0
 _([0,1,2,3,4]).reject(i=>i%2===1).min()
 0
 typeof _([0,1,2,3,4]).reject(i=>i%2===1).min()
 number
 !!_([0,1,2,3,4]).reject(i=>i%2===1).min()
 false
```

My previous experience lead me to doubt that the type would be number in this case.

creep.action.bulldozing.js
karlthepagan: Work only comes into play when you have sufficient range. Before then you get step movement.

creep.action.bulldozing.js
Spedwards: It only bulldozes people `< 100`.

creep.action.bulldozing.js
DasBrain: Lodash methods that return one element return that element, not a wraped object. This is not true for methods that return a collection.

PR #1015, opened 2017-03-19
Proposed new/safer paramters

Dasku:
Just to start the conversation about parameters and updating to better defaults. 

Reasoning:
Room trading - why wouldn't you?
SELL_RATIO - due to really low prices recently, I can't see any reason why you wouldn't sell excess if its possible
MAX_REPAIR - Don't think infinity works
CONTROLLER_SIGN - Non-members who use our code and don't check parameters 
DRIVE_BY_REPAIR_RANGE - CPU save


parameter.js
```diff
@@ -58,20 +58,20 @@ let mod = {
...
     TERMINAL_ENERGY: 100000,
     MIN_SELL_RATIO: {
-        'H': 0.2,
-        'O': 0.2,
-        'U': 0.3,
-        'L': 0.3,
-        'K': 0.3,
-        'Z': 0.3,
-        'X': 0.4
+        'H': 0.02,
```
karlthepagan: giving away minerals for super cheap?

parameter.js
```diff
@@ -117,7 +117,7 @@ let mod = {
     COMBAT_CREEPS_RESPECT_RAMPARTS: false, // causes own creeps not to leave through ramparts
     COST_MATRIX_VALIDITY: 1000,
     CONSTRUCTION_PRIORITY: [STRUCTURE_SPAWN,STRUCTURE_EXTENSION,STRUCTURE_LINK,STRUCTURE_TERMINAL,STRUCTURE_STORAGE,STRUCTURE_TOWER,STRUCTURE_POWER_SPAWN,STRUCTURE_NUKER,STRUCTURE_OBSERVER,STRUCTURE_ROAD,STRUCTURE_CONTAINER,STRUCTURE_EXTRACTOR,STRUCTURE_LAB,STRUCTURE_WALL,STRUCTURE_RAMPART],
-    CONTROLLER_SIGN: false,
+    CONTROLLER_SIGN: true,
```
karlthepagan: Should users sign by default before they understand how everything works?

parameter.js
Dasku: Well minerals sell for credits and super cheap still means credits right? The issue I had was at say 0.4, nothing ever really sold so my storage's just filled up. Had to drop the ratio to get them to sell. I am right on the edge of the map tho so prob less people to trade with. 

I've never really understood "ratio" if I'm honest. But your terminal will always take the best price possible, so having a very low floor just means you can still sell when there isn't any good prices around.

parameter.js
Dasku: Good point, reading the following line shows what the sign is, so this was a "if they don't even bother reading the parameters file, lets get some free advertisement"

parameter.js
OverlordQ: It's the profit per unit of resource after subtracting out the energy cost.

karlthepagan: Energy could probably go much lower. I generally agree with being more aggressive about new players selling minerals, but we don't want to make a run on noob markets.

parameter.js
karlthepagan: I can accept this.

parameter.js
chobobobo: I usually have to set energy lower and all of these below 0.2, with 0 at 0.05 generally to get it to sell successfully.

parameter.js
```diff
@@ -58,20 +58,20 @@ let mod = {
         8: 50000
     },
     MAX_STORAGE_MINERAL:200000, // keep a max of each type of minerals in store
-    ROOM_TRADING: false, // set this true to enable haulers within your colony to request resources from other rooms in your colony
+    ROOM_TRADING: true, // set this true to enable haulers within your colony to request resources from other rooms in your colony
     FILL_POWERSPAWN: true,
-    MIN_MINERAL_SELL_AMOUNT:20000,
+    MIN_MINERAL_SELL_AMOUNT: 20000,
     ENERGY_VALUE_CREDITS: 0.15, // assumed energy exchange rate (in credits) to determine best mineral sell offer
```
chobobobo: I think this needs to be much, much lower as energy typically trades at 0.05 or below in my experience.

parameter.js
Dasku: Updated as suggested
    ENERGY_VALUE_CREDITS: 0.05,
    MIN_SELL_RATIO: {
        'H': 0.2,
        'O': 0.2,
        'U': 1,
        'L': 1,
        'K': 1,
        'Z': 1,
        'X': 1

parameter.js
Dasku: changed to 0.05

chobobobo:
My last remaining concern is that setting default prices at 1 is basically the same as turning off room trading as prices currently fluctuate between 0.05 - 0.25 in my experience.

Dasku:
0.2 seems pretty safe

chobobobo:
I think @karlthepagan's suggestion of lower values for O and H would hold up in my experience, and it just sets a lower bound, you will still sell at higher values if they are available and close to you.  After that are these changes ready to be merged?

Dasku:
Think so, just updated to 0.02 for H and O

chobobobo:
@karlthepagan I believe your requested changes have now been merged

PR #1018, opened 2017-03-20
(behaveYourselves) Refactored creep.action selection and priorities

chobobobo:
Long-term I would like to see resource based actions divided into actions that need something, and actions that use something (inflow vs outflow).  I think more generally we should define what sources a creep is able to get energy from, and prioritize where it should use that energy and this will simplify/deobfuscate some of what we see sometimes with reallocating.

I think inflow and outflow actions are a part that could be modified using behaviours if needed.

This is untested, and not ready for use yet, but putting here to get feedback.  Will do workers next.


chobobobo:
I would like to fix debounce in this branch as well so that we can call isValidAction and isAddableAction using the anticipated resources after we withdraw..

creep.Behaviour.js
```diff
@@ -0,0 +1,43 @@
+// base class for behaviours
+let Behaviour = function(name) {
```
Spedwards: \*cough\* _`const`_ \*cough\*

chobobobo: `const`antly improving.

chobobobo:
Ok, lets pretend those tabs never happened, I am running this code now and I believe it's ready for canary but merge conflicts are fairly likely.

creep.action.withdrawing.js
```diff
@@ -22,18 +22,22 @@ action.onAssignment = function(creep, target) {
...
-        shouldCall = _.some(outflowActions, a => a.newTarget(creep));
+        //FIXME: How do I spoof creep.sum only for this call so we can check validAction & addableAction
+        const dummyCreep = {};
+        for (let key in creep) {
+            dummyCreep[key] = creep[key];
+        }
+        for (let key in dummyCreep.carry) {
+            dummyCreep.carry[key] = 0;
+        }
+        dummyCreep.carry[RESOURCE_ENERGY] = dummyCreep.carryCapacity; // assume we get a full load of energy
```
chobobobo: @karlthepagan is this an evil way to do this?

creep.Behaviour.js
```diff
@@ -0,0 +1,43 @@
...
+	};
+	this.selectInflowAction = function(creep) {
+		for (const action of this.inflowActions(creep)) {
+			if (!action.debounce || action.debounce(creep, this.outflowActions(creep))) {
+				if (this.assignAction(creep, action)) return;
+			}
+		}
+		return Creep.action.idle.assign(creep);
+	};
+	this.selectAction = function(creep, actions) {
```
chobobobo: Should this be written to take either a list of actions, or a function that returns a list of actions and then determines which it is inside here?  it would clean up `mod.selectAction(creep, getList)` vs `mod.selectAction(creep, getList(creep))`

creep.action.withdrawing.js
karlthepagan: this is why I didn't originally use isValidAction in the debounce implementation! :)

chobobobo:
@Spedwards I believe the changes you suggested have been incorporated now

creep.Behaviour.js
```diff
@@ -0,0 +1,43 @@
+// base class for behaviours
+const Behaviour = function(name) {
+    this.name = name;
+    this.actions = (creep) => []; // priority list of non resource based actions
+    this.inflowActions = (creep) => []; // priority list of actions for getting resources
+    this.outflowActions = (creep) => []; // priority list of actions for using resources
```
karlthepagan: when these are functions it allows us to call baseOf.viral.method and allows us to alternate the implementation and delegate to strategies

when these are not functions it is less likely to be implemented in error

creep.behaviour.pioneer.js
```diff
@@ -17,6 +16,41 @@ mod.run = function(creep) {
         logError('Creep without action/activity!\nCreep: ' + creep.name + '\ndata: ' + JSON.stringify(creep.data));
     }
 };
+mod.outflowActions = (creep) => {
+    let priority;
+    if (creep.room.controller && creep.room.controller.level < 2) {
```
karlthepagan: Because this changes the order of actions it's difficult to use a strategy handler.

chobobobo: I will look into whether I can convert this to use strategies for inflow + outflow actions so that they are easily overridable.

chobobobo:
workers don't dismantle, and it looks like pioneers don't harvest

chobobobo:
@karlthepagan after some thought, I can't think of a way to determine priority listing using strategies that would make it both simple for someone debugging to understand the priorities, and extensible using strategies.  If you have an idea in mind, I'd be happy to incorporate it and push this to 0.8.0, however the way it is now I believe this could be in 0.7.0 with a bit more testing.

chobobobo:
I need to investigate whether this change has reintroduced the issue of creeps "idling full" instead of empty

chobobobo:
0.8.0 is quickly becoming the 'refactor release'

chobobobo:
@karlthepagan in light of #1185, I need some guidance on how I should proceed with this refactor.  Holding it alongside /dev for more than a few weeks will be unsustainable because of merge conflicts.  If the behaviour files simply returned an instance eg `module.exports = new Behaviour()` current virals should work as-is, but I will need to rewrite Pioneer, CollapseWorker and a couple of others because they won't be able to extend base-classes as I have them doing now.

Spedwards:
Suggestion: Store the behaviour instance (not the name) on `creep.behaviour`. Similar to how actions is done.

karlthepagan:
This should proceed independently of class refactor if possible.

chobobobo:
Reverted back to pre-class changes, will build using non-prototypical classes for now.

karlthepagan:
how does this relate to #1041 (wip refactor of reallocating)?

chobobobo:
This refactor will leave reallocating alone for now as it is too complex to include.  The goal of this refactor is to simplify action selection debugging, and gather common functionality so that we can optimize in one place rather than several.  Once I have stabilized 0.7.0 I will get back to work on this.

chobobobo:
Currently getting haulers and workers idling instead of feeding, not sure why yet.

chobobobo:
This seems to work now, but will need some tuning for efficiency, 150 tick snapshot of current usage, looks like reallocating is a big drain:
screen shot 2017-04-24 at 10 22 56 pm

chobobobo:
More in depth look at efficiency with this branch:
screen shot 2017-04-25 at 6 32 29 pm

chobobobo:
I have also included a refactor of the miner behaviours and added action.mining to this

chobobobo:
@Spedwards @karlthepagan I would like to merge this into dev now, it's stable and may have one or two minor bugs but I've not detected anything major.  Efficiency is better than previous in most cases as well.

chobobobo:
one notable issue is with privateers, and the invade.exploit flag, that needs some work and may actually be broken in dev as well..

PR #1019, opened 2017-03-20
some minor changes to reallocating

chobobobo:
remove restrictions on maxPerTarget and maxPerAction, allow creeps carrying energy to reallocate rather than sit idle when they have no other action to use the energy.


chobobobo:
This was not a good idea, refactoring reallocating instead.

PR #1020, opened 2017-03-20
improve miner build/repair

chobobobo:
Allow them to resume working if they get interrupted without the 50tick debounce, also only repair/fortify roads and containers (doesn't really matter if they build anything in range)


chobobobo:
I sometimes see idle miners sitting on low containers with energy, so I'm not convinced this is stable yet

chobobobo:
Ok, remaining issues resolved.

PR #1023, opened 2017-03-20
Util Methods

Spedwards:
Created a Util module. This will house all methods that don't belong somewhere specific. I'll move some of the methods from global here as well.

If/When this gets merged, I will go through and convert old methods to use the new Util methods.

### DIFF
--- a/util.js
+++ b/util.js
@@ -1,3 +1,5 @@
+// All methods require a JSDoc comment describing it.
+// http://usejsdoc.org/
 module.exports = {
     /**
      * formats an integer into a readable value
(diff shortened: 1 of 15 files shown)


chobobobo:
I'll move in the things I have added as utils as well, it makes me wonder if I should also split out profiling code

util.js
```diff
@@ -0,0 +1,231 @@
...
+                Memory.statistics.reports = [];
+            } else {
+                mails = Memory.statistics.reports.splice(0, REPORTS_PER_LOOP);
+            }
+            const send = Game.notify;
+            _.forEach(mails, send);
+        }
+    },
+    
+    routeRange(fromRoom, toRoom) {
```
Spedwards: This refactor may require testing, but I'm fairly confident in it.

Spedwards:
Deprecated the methods in global and forwarded them to the Util versions. Couldn't outright remove them and replace everything as it would break virals and internal.

PR #1028, opened 2017-03-21
Proposed refactor of room to split it into smaller parts.

chobobobo:
The motivation for this is that it gets hard to find functionality, and hard to know where to put changes.  Also merge conflicts are awful when someone adds something unrelated to a huge file, so this will ensure that conflicts only tend to happen when two people are changing the same functionality.

Long-term, I am hoping this will enable us to refactor action.reallocating's getNeeds into the different structure types so that we can define a Terminal's input + output limits & needs in one, easy to understand place.


chobobobo:
I have merged and tested this on a private server to debug and test and it worked.  It will likely have conflicts with several other branches when it is merged due to the fact that room.js is currently so monolithic so I will resolve those conflicts when they arise.

main.js
```diff
@@ -204,6 +204,23 @@ global.install = () => {
...
+        Observer: load("room.observer"),
+        Orders: load("room.orders"),
+        Pathfinding: load("room.pathfinding"),
+        PowerBank: load("room.powerBank"),
+        PowerSpawns: load("room.powerSpawns"),
+        Resources: load("room.resources"),
+        Spawns: load("room.spawns"),
+        Structures: load("room.structures"),
+        Terminal: load("room.terminal"),
+        Towers: load("room.towers"),
```
karlthepagan: Does the upper case indicate that these are components of the injection framework?

If that's the case I'd rather they have an underscore prefix.

I'd also be interested in hearing your thoughts about not adding this code to the API base class. Could all of the mod code live separate from the API class?

chobobobo: I'm honestly not certain what it would look like if it wasn't attached to the API Room class.  If they had their own classes we would need to be passing in a room object all the time which to me would be more cumbersome and less OO than the current method.

karlthepagan: Uppercase conventionally indicates a class.

Underscore conventionally indicates a private property that the end user shouldn't mess with.

Spedwards: I discussed the following with Cho on this matter:

 - Since these files are purely for extending the base `Room` object, they should be private. The JavaScript standard for private methods and fields is prefixing with 2 underscores, though how many doesn't matter.

 - Any moved static methods will need to be declared as `Room.*` instead of `mod.*`, and calls to any moved static methods will also need to be changed in the same manner. Since extending this way, it will need to be done within a `mod.extend` method to play safe.

chobobobo:
Until we have branched 0.7.0 there's not much point in pursuing this because the merge conflicts will be unsustainable.  This will likely need to be rewritten properly then.

karlthepagan:
discussed putting this in 0.7.1 - with the caveat that it won't be much of a logical refactor just a structural one

chobobobo:
Decided to split up the refactor into more manageable parts, this will be closed when that process is complete.

PR #1030, opened 2017-03-21
Miners choose spots with containers by default now

chobobobo:


Perrytheplatypus43:
should we do this for upgraders to?

chobobobo:
The fix for upgraders will need to be a bit more complex because if there is an upgrade link you want to make sure they are adjacent to it, it has priority over being close to a container once it exists.

For upgraders the real problem is that the code for finding a spot is only considering locations adjacent to the containers as valid but it should also be considering the actual location of the container as well.

chobobobo:
I think the change would need to be in the function Room.fieldsInRange to somehow check if the target position is walkable or not.

PR #1031, opened 2017-03-21
Remove extend method (Example)

Spedwards:
This is just an example, @karlthepagan wanted to know what I was talking about.

When the compiler runs down the file, it will run everything it sees in order from top to bottom. If the assignment is in a function, it will have to wait for that function to call (and any preceding conditions) before the assignment to take place, otherwise it doesn't matter. This example removes the `mod.extend` method from `room.js` and brings everything that was in it to the top-level.

### DIFF
--- a/room.js
+++ b/room.js
@@ -4,2175 +4,2143 @@ let find = Room.prototype.find;
 let mod = {};
 module.exports = mod;
 
-mod.extend = function(){
-    let Container = function(room){
-        this.room = room;
 
-        Object.defineProperties(this, {
-            'all': {
-                configurable: true,
-                get: function() {
-                    if( _.isUndefined(this.room.memory.container)) {
-                        this.room.saveContainers();
-                    }
-                    if( _.isUndefined(this._container) ){
-                        this._container = [];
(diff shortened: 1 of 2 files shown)


room.js
```diff
@@ -4,2175 +4,2143 @@ let find = Room.prototype.find;
...
-            get: function() {
-                if( _.isUndefined(this._collapsed) ) {
-                    // only if owned
-                    if( !this.my ) {
-                        this._collapsed = false;
-                        return;
+    });
+};
+
+Object.defineProperties(Room.prototype, {
```
karlthepagan: This will be executed when code is loaded the first time. We will need to test it thoroughly.

chobobobo: A simple comparison would be to put a log message on load of extend vs this and compare output between before/after

chobobobo:
It actually seems pretty clear to me that removing extend and moving all of that into the main scope would have the same effect as extend, the only difference I see is when the objects will get extended (on load instead of on call to .extend())

However, for virals, won't this affect the way that internalViral.room.js extends?

PR #1033, opened 2017-03-22
Classify Actions

Spedwards:
As per #1031, `extend` is not required, but fits with how it currently works.

### DIFF
--- a/creep.Action.js
+++ b/creep.Action.js
@@ -1,13 +1,13 @@
 // base class for every action
 const Action = class {
-    constructor(actionName) {
-        // action name
-        this.name = actionName;
+    constructor(...args) {
+        Action.prototype.constructor.apply(this, args);
     }
 };
 module.exports = Action;
 
 Action.extend = function() {
+    // INSTANCE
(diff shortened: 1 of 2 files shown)


chobobobo:
I love OOP and classes, and I think inheritance would make things so much clearer in our code.  That said, the last time I looked into ES6 classes for Screeps there were some concerns about performance.. namely the use of closures in classes being possibly much less efficient.  I've learned a lot since then, so it's possible I was misunderstanding the issues before, any thoughts on that?

Spedwards:
The changes I've made in this PR are purely a syntax change. We were still creating a class, but using the old ES5 syntax.

The only difference here is that I've moved the instance methods and fields to a function, which as stated isn't even necessary.

karlthepagan:
Well I did check with Matt about the performance question. Turns out that benchmarking is hard and different sources come up with wildly different results.

If you want a single authority on ES6 feature performance I highly recommend this one:
http://kpdecker.github.io/six-speed/

chobobobo:
This should get merged into a running branch and tested, but I support it.

PR #1036, opened 2017-03-22
Visuals Refactor

Spedwards:
Makes it a little easier to understand what you're looking at and to extend it with virals. I've set a few styles to start off with.

### DIFF
--- a/main.js
+++ b/main.js
@@ -216,6 +216,8 @@ global.install = () => {
     Spawn.extend();
     FlagDir.extend();
     Task.populate();
+    
+    if (ROOM_VISUALS && !Memory.CPU_CRITICAL && Visuals) Visuals.extend();
     // custom extend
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -216,6 +216,8 @@ global.install = () => {
     Spawn.extend();
     FlagDir.extend();
     Task.populate();
+    
+    if (ROOM_VISUALS && !Memory.CPU_CRITICAL && Visuals) Visuals.extend();
```
chobobobo: This will need to be 
`if (ROOM_VISUALS && Visuals) Visuals.extend()`
otherwise the code never gets loaded, there's actually a bug further up where Visuals is loaded that needs to be changed to:
`Visuals: ROOM_VISUALS ? load('visuals') : undefined,`

PR #1038, opened 2017-03-22
creep.explain() tries to explain what the creep is doing

karlthepagan:
@OverlordQ this one is for you - it's only partially done, but maybe you can help with the innards.

Right now you can do this:
```
> Game.getObjectById('58d2eb8ca45ad69f4dbdf5f2').explain()
< hauler-1250-2: assigned (storing,hauler,)
```

If any of the creep's strategy clients (e.g. Action, Behaviour, Task) implement the explain(creep) function then that method's return value will be appended.

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -349,6 +349,11 @@ mod.extend = function(){
             selector: taskName => Task[taskName] && Task[taskName],
         });
 
+    // Explain API extension
+    Creep.prototype.explainAgent = function() {
+        return `ttl:${this.ticksToLive} pos:${this.pos}`;
+    };
+
     // API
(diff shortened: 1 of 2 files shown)


karlthepagan:
I realized that that this makes more sense as a feature added directly in the strategy API.

karlthepagan:
Final version looks like:
```
> Game.getObjectById('58d2eb8ca45ad69f4dbdf5f2').explain()
< [creep hauler-1250-2]: ttl:213 pos:[room W24N19 pos 8,28] assigned:[uncharging,hauler,]
```

PR #1040, opened 2017-03-23
Change/parts refactor

Spedwards:


chobobobo:
There also may be a bad interaction here between this change and internalViral.creep.js' compileBody, I was getting max sized guards with one extra MOVE part with a guard flag set in an SK room
```
 Offspring failed: ERR_INVALID_ARGS
 - body: {"attack":9,"heal":8,"move":18,"ranged_attack":8,"tough":8}
 - name: guard-E85S26-guard-1
 - behaviour: warrior
 - destiny: [object Object]
```

chobobobo:
Typo was what caused the inverted bodies, but I'm not sure it explains the guard problems, will deploy the fix and let you know about guards

Issue #1032, opened 2017-03-22
Refactor Creep.action.reallocating


chobobobo:
My initial thoughts are this:
Reallocating needs to be split into two actions
Action.loading - putting the required resources in the structure
 - debounced by having somewhere to get the resources from
Action.unloading - removing unwanted resources from a structure
 - debounces by having somewhere to put the unwanted resources

All other resources are handled just by being made 'available' and creep types define their strategies for making use of those resources.  eg. haulers should look for the task that makes the best use of their capacity, whereas workers should just get the needed resources from the nearest available source.

Defining what is 'available' for a structure will be (might already be) defined on the structure itself eg. Room.terminal.getNeeds(RESOURCE_ENERGY) - if positive action.loading, if negative action.unloading as well as resources being defined as 'available' for use.

chobobobo:
This is going to need to wait until #1028 is in because it will require adding functions to labs rather than accessing room.memory.labs directly

PR #1041, opened 2017-03-23
[WIP] Refactor action.reallocating to be an outflow only action

chobobobo:
The idea is that reallocating should only handle deliveries of resources, ~a new action for forced unloading (creep.action.unloading) will handle structures that need to be unloaded (like labs that have changed resources).~ unloading will be handled by withdraw & uncharge

This will require modifying withdraw & uncharging to handle any kind of resource and moving unloadStructure functionality into them, and to debounce their use based on reallocating.findNeeding().  I also plan to add terminals and labs as valid structures for the withdrawing action, returning the nearest valid target for now, strategies can further specialize which structures a creep should get resources from.

Should containers be a part of reallocating, why are they currently?


karlthepagan:
I started observing a cycle in dev, this might be a useful observation or not.

The current stuck cycle is withdrawing -> storing -> (debounce) -> reallocating (nothing to reallocate) -> withdrawing

chobobobo:
The bug to look for there is why reallocating was assigned if there was nothing to reallocate​.

karlthepagan:
How does this change relate to #1032 (WIP refactor reallocating)?

chobobobo:
This is very much on-hold and not up to date with my current understanding of actions, which are expressed in #1032 I will rewrite this once 0.7.0 as branched as it's a big job, and there will be conflicts if I do it now.

PR #1042, opened 2017-03-23
Easily compare flags

Spedwards:
Hate typing out the full comparison check every time

### DIFF
--- a/flagDir.js
+++ b/flagDir.js
@@ -142,7 +142,7 @@ mod.extend = function(){
         }
     });
     
-    Object.defineProperty(Flag.prototype, 'is', {
+    Object.defineProperty(Flag.prototype, 'compareTo', {
         configurable: true,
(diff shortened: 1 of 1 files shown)


flagDir.js
```diff
@@ -133,7 +133,22 @@ mod.extend = function(){
...
+    });
+    
+    Object.defineProperty(Flag, 'compare', {
+        configurable: true,
+        value: function(flagA, flagB) {
+            return flagA.color === flagB.color && flagA.secondaryColor === flagB.secondaryColor;
+        }
+    });
+    
+    Object.defineProperty(Flag.prototype, 'is', {
```
chobobobo: flagA.is(flagB) isn't clear about exactly what is being compared.. this is checking whether it has the same colours as the other flag, essentially it is the same type.  I would suggest naming this something like sameType or typeEqualTo

PR #1045, opened 2017-03-24
Task/lab tech part deux

Dasku:
3rd time's a charm

### DIFF
--- a/creep.behaviour.labTech.js
+++ b/creep.behaviour.labTech.js
@@ -1,0 +1,68 @@
+let mod = {};
+module.exports = mod;
+mod.name = 'labTech';
+mod.run = function(creep) {
+    // Assign next Action
+    let oldTargetId = creep.data.targetId;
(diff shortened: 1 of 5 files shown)


Dasku:
Wiki updated here:
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/wiki/Flags

chobobobo:
screen shot 2017-03-26 at 7 31 01 am

labTech's seem to spend a lot of time idling, not sure exactly why.. they also aren't properly filling power

Dasku:
Yep, but thats down to reallocate isnt it? Do you have lab orders? Also nuker fill isnt working atm. So once everything is working correctly and you have lab orders going they will have more to do.

chobobobo:
Yeah, I don't doubt it's an issue with reallocate, but I kind of figured they should be at least charging empty extensions.  I also haven't started using labs so once labs are going they will be busier.

PR #1047, opened 2017-03-24
Worker body change

SirLovi:
There should be equal number of move parts to other parts when hauling energy without roads (RCL1-3)


Perrytheplatypus43:
I agree to this because:
before: walk time (empty,full)	plain=1,2   road=1,1   swamp=5,10  
after: walk time empty,full	plain=1,1   road=1,1   swamp=3,5

Dasku:
Wouldn't this reduce the number of multi body's available at each level and lose efficiency that way instead? Walking faster doesn't benefit you if you have to make twice as many trips right?

chobobobo:
This also makes the assumption that you are hauling without roads at RCL1-3, however some players (myself included) build roads first, I think this change would be more suitable as a viral, and I'm happy to explain how to do that if you would like @SirLovi

creep.setup.worker.js
```diff
@@ -39,26 +39,26 @@ setup.byPopulation = function(type, start, perBody, limit) {
...
         maxMulti: 8,
         maxCount: room => setup.maxWorker(room),
         maxWeight: 14400
     },
     3: {
         fixedBody: [],
-        multiBody: [CARRY, WORK, MOVE],
-        minAbsEnergyAvailable: 200,
+        multiBody: [CARRY, WORK, MOVE, MOVE],
+        minAbsEnergyAvailable: 250,
```
karlthepagan: please revert the RCL3 settings, but IMO the RCL1 and RCL2 strategy should be no-roads!

karlthepagan:
after having respawned and examined the fast-start strategy for myself I believe that roads at RCL1-2 

I have updated the guide to reflect this recommendation: https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/wiki/How-to-Use-This-Code-(Guide)

@chobobobo @Spedwards opinions?

chobobobo:
for RCL1-2 it's a minor enough change that I'm willing to try, I can always use a viral myself if I want to change it :)

karlthepagan:
Looking for changes @SirLovi - I can get to these if you like after I triage the other issues.

This whole project is getting off the rails lately.

karlthepagan:
I only meant the RCL1 change to be reverted.

PR #1049, opened 2017-03-25
mineralMiner was not properly determining existingBranding

chobobobo:
This is required to fix https://github.com/cyberblast/screeps.ocs.internal/pull/267


chobobobo:
I was wrong, this is working.

PR #1052, opened 2017-03-25
Implement picking strategy for haulers and remoteHaulers

chobobobo:
Defines the default strategy for picking to only pick energy, currently this strategy is only overridden for haulers and remoteHaulers to allow them to pick resources as well as energy.

### DIFF
--- a/creep.Action.js
+++ b/creep.Action.js
@@ -121,8 +121,9 @@ let Action = function(actionName){
     this.selectStrategies = function() {
         return [this.defaultStrategy];
     };
-    this.getStrategy = function(strategyName, creep, args = []) {
-        return creep.getStrategyHandler([this.name], strategyName, args);
+    this.getStrategy = function(strategyName, creep, args) {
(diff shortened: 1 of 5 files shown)


chobobobo:
I think I still saw a remoteWorker with minerals so this might not be working correctly yet

chobobobo:
This should wait for https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/685/files to merge because there will be conflicts, and that PR is much more comprehensive.

creep.Action.js
```diff
@@ -121,5 +121,9 @@ let Action = function(actionName){
     this.selectStrategies = function() {
         return [this.defaultStrategy];
     };
+    this.getStrategy = function(strategyName, creep, args) {
+        if (_.isUndefined(args)) return creep.getStrategyHandler([this.name], strategyName);
+        else return creep.getStrategyHandler([this.name], strategyName, args);
+    };
```
karlthepagan: I see, slightly more clear.

PR #1069, opened 2017-03-26
Update to use new Flag methods

Spedwards:


creep.behaviour.privateer.js
```diff
@@ -66,8 +66,7 @@ mod.nextAction = function(creep){
         // at target room
         if( creep.flag && creep.flag.pos.roomName == creep.pos.roomName ){
             // check invader/cloaking state
-            if( creep.room.situation.invasion &&
-                (creep.flag.color != FLAG_COLOR.invade.robbing.color || creep.flag.secondaryColor != FLAG_COLOR.invade.robbing.secondaryColor )) {
```
Spedwards: I'm surprised the `secondaryColor !== robbing.secondaryColor` never caused any issues, since this is an under an OR.

creep.behaviour.privateer.js
```diff
@@ -66,8 +66,7 @@ mod.nextAction = function(creep){
...
+            if( creep.room.situation.invasion && !Flag.compare(creep.flag, FLAG_COLOR.invade.robbing)) {
```
chobobobo: I think this one could be creep.flag.compareTo(FLAG_COLOR.invade.robbing) right?

Spedwards: Is `creep.flag instanceof Flag`?

chobobobo: I have no idea!  If it isn't then the name is misleading.

Spedwards: This is why I chose to use `Flag.compare` instead. Without knowing for sure, it's a safer choice.

karlthepagan: Yes it's assigned in population analyze.

Such assignments are somewhat problematic because of stringify.

PR #1077, opened 2017-03-27
Allows the option to use functions in certain parameters

### LINKED ISSUE #928: Swap Spawn and Extensions in Construction Priority
Currently the default first two structures in the list of priorities are Spawn, followed by Extensions. In my opinion, these two should be swapped.

For your first spawn, it's unlikely that you'll have extensions ready before it's done, and for all following spawns, it's probably better to build extensions first, so you spawn larger creeps, decreasing CPU usage and decreasing the time it would take to build said spawn.

Let me know what you think. If there's an agreement, we could change this.

Spedwards:
Also resolves #928


parameter.js
```diff
@@ -127,7 +129,14 @@ let mod = {
     NOTIFICATE_HOSTILES: true, // Log any hostiles - Ignores NOTIFICATE_INTRUDER and NOTIFICATE_INVADER
     COMBAT_CREEPS_RESPECT_RAMPARTS: false, // causes own creeps not to leave through ramparts
     COST_MATRIX_VALIDITY: 1000,
-    CONSTRUCTION_PRIORITY: [STRUCTURE_SPAWN,STRUCTURE_EXTENSION,STRUCTURE_LINK,STRUCTURE_TERMINAL,STRUCTURE_STORAGE,STRUCTURE_TOWER,STRUCTURE_POWER_SPAWN,STRUCTURE_NUKER,STRUCTURE_OBSERVER,STRUCTURE_ROAD,STRUCTURE_CONTAINER,STRUCTURE_EXTRACTOR,STRUCTURE_LAB,STRUCTURE_WALL,STRUCTURE_RAMPART],
+    // function parameters: room. expected result: array
+    CONSTRUCTION_PRIORITY: (room) => {
```
karlthepagan: this is not easy for new users to get their hands around

I would rather see a viral.room.js example.

karlthepagan:
I think it's fine to make specialized logic for the first spawn only to be prioritized first.

karlthepagan:
>  this is a fairly ignorant thing to say that could be applied to any change

I think you're trying to say that I've made a hasty generalization, so I will not take offense.

Hoever, in my opinion the contents of parameters.js should be the most safe, comprehensible, and hallowed of all the code we write. I would go as far as to say my use of backticks in the sign message is pushing the limit of what should be done.

My opinion remains that parameters.js is a special case and should be written in a way which is completely comprehensible to anyone who is on their first day of programming.

I would like these values to be represented by a single statement. I don't have a problem with something more complex showing up in virals of course.

Would you accept this being implemented in internalViral.parameters.js? Or in https://github.com/ScreepsOCS/screeps.ocs.overrides-template ?

Issue #1048, opened 2017-03-24
Remote hauler will be assigned to a room with no spawn and get stuck


karlthepagan:
https://screeps.com/a/#!/history/E81N29?t=18243366

The hauler at the right is spamming "no action" messages.

The solution was to put a white/yellow "drop" flag next to the spawn.

A better hierarchy for dropping: drop at flag, or spawn, or spawn construction site, or controller.

Need to figure out if this is a problem for the release candidate.

chobobobo:
resolved by #1078

PR #1078, opened 2017-03-27
RC 0.6.3 Fix dropping with remoteHaulers

chobobobo:
set spawn construction site as a target for dropping, set default action for remoteHaulers to drop everything and idle.  This addresses https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1048


chobobobo:
I will also cherrypick this for dev, as it's still relevant there.

PR #1079, opened 2017-03-27
Fix Flag Finding

Spedwards:


creep.action.dropping.js
```diff
@@ -12,7 +12,7 @@ action.newTarget = function(creep) {
         drop = creep.pos.findClosestByRange(creep.room.structures.spawns);
     }
     if( !drop ) {
-        drop = creep.pos.findClosestByRange(creep.room.find(FIND_FLAGS, {filter: FLAG_COLOR.claim.spawn.filter}));
+        drop = creep.pos.findClosestByRange(creep.room.find(FIND_FLAGS, Flag.flagFilter(FLAG_COLOR.commands.drop)));
```
chobobobo: Yeah, this needs to filter for FLAG_COLOR.claim.spawn, I will remove the change from the other PR.

karlthepagan:
@chobobobo is this ok now?

room.js
```diff
@@ -404,8 +404,9 @@ mod.extend = function(){
                 get: function() {
                     if( _.isUndefined(this._piles) ){
                         const room = this.room;
-                        this._piles = room.find(FIND_FLAGS, {filter: FLAG_COLOR.command.drop.filter})
+                        this._piles = FlagDir.filter(FLAG_COLOR.command.drop, this.room.getPositionAt(25,25), true)
                             .map(function(flag) {
+                                flag = Game.flags[flag.name];
```
chobobobo: I don't think this line is necessary as it's a parameter to the function.

room.js
```diff
@@ -404,8 +404,9 @@ mod.extend = function(){
                 get: function() {
                     if( _.isUndefined(this._piles) ){
                         const room = this.room;
-                        this._piles = room.find(FIND_FLAGS, {filter: FLAG_COLOR.command.drop.filter})
+                        this._piles = FlagDir.filter(FLAG_COLOR.command.drop, this.room.getPositionAt(25,25), true)
```
chobobobo: nitpicking at this point, but `this.room` could be changed to `room` as it's a const above.

room.js
chobobobo: Ah, nevermind, I remember you pointed out that FlagDir doesn't return flags, it returns some other data object.. maybe rename the parameter to something other than flag so other people don't get confused like I do when skimming the code

Spedwards: As I mentioned in Slack, `FlagDir.filter` does not return flags, rather an object containing flag information.

PR #1080, opened 2017-03-27
Fix/dropping

chobobobo:
Fix dropping for remoteHaulers and cases where there is no spawn.


creep.action.dropping.js
```diff
@@ -12,14 +12,21 @@ action.newTarget = function(creep) {
         drop = creep.pos.findClosestByRange(creep.room.structures.spawns);
     }
     if( !drop ) {
-        drop = creep.pos.findClosestByRange(creep.room.find(FIND_FLAGS, {filter: FLAG_COLOR.claim.spawn.filter}));
+        drop = creep.pos.findClosestByRange(FlagDir.filter(FLAG_COLOR.claim.spawn, creep.pos, true));
```
Spedwards: Firstly, `Flag.filter`, it also doesn't return an actual Flag. See #1079

chobobobo: I will remove this line from this PR in favour of #1079, this becomes dependent on that PR.

chobobobo:
This is dependent on https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/1079/files merging first.

Issue #1083, opened 2017-03-27
Semi-auto construction: limited construction sites


karlthepagan:
I'm at 94 construction sites. If I play over 6 auto-build flags the CS creation fails but the flag is still removed.

IMO the flag should wait until there is a construction site underneath it which matches the flag's purpose before it gets removed. I don't trust the Screeps API 100% so I hesitate to trust the return value of the create command.

Spedwards:
This shouldn't be happening...
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/blob/dev/room.js#L1442

Spedwards:
I do have an idea for a hack-attempt to fix this, if the checks do not work.

Spedwards:
Resolved in #1086

PR #1086, opened 2017-03-28
Semi-Auto Construction Flag Fix

Spedwards:
Potential Resolution to #1083


karlthepagan:
I don't have enough cs's or rooms turning over fast enough to repro this.

However I took a much closer look and I understand it. Thanks!

PR #1087, opened 2017-03-28
Add costMatrixInvalid event

chobobobo:
Adds Room.costMatrixInvalid() event which can be called with either a room or a roomName, also Room.rebuildCostMatrix() which can also be called with either room or roomName for the user in emergencies.

Automatically checks for owned structure and construction site changes and rebuilds the costMatrix when changes are detected, also rebuilds when action.dismantling completes.


room.js
```diff
@@ -2742,6 +2767,14 @@ mod.flush = function(){
     }
 };
 mod.analyze = function(){
+    const numSites = _.size(Game.constructionSites);
+    const sitesChanged = !_.isUndefined(Memory.rooms.myTotalSites) && numSites !== Memory.rooms.myTotalSites;
+    Memory.rooms.myTotalSites = numSites;
+
+    const numStructures = _.size(Game.structures);
+    const structuresChanged = !_.isUndefined(Memory.rooms.myTotalStructures) && numStructures !== Memory.rooms.myTotalStructures;
+    Memory.rooms.myTotalStructures = numStructures;
```
Spedwards: Lines 2770 to 2776 should be in it's own function(s).

chobobobo:
@Spedwards have the changes satisfied your requests?

chobobobo:
I believe that we only need to rebuild when a new construction site has been added, or an existing structure we own has been lost, the most recent patch should address that.

PR #1089, opened 2017-03-28
add creeps at their determinatedSpots to the structureMatrix

chobobobo:


chobobobo:
Tested and worked on private server.

karlthepagan:
Seems reasonable. Some day we'll have to re-examine all the things we cache, but not any time soon.

chobobobo:
When creeps die, they need to invalidate the costMatrix

PR #1095, opened 2017-03-29
test for valid flag on dropping assignment

karlthepagan:
might need some help on this one. I've got a quick test that halts broken haulers and sets them in the right direction but I haven't tested the outcome or the normal operation of this fix

this prevents creeps with the dropping action from being assigned to any old flag... still not sure how they get assigned to these flags

### DIFF
--- a/creep.action.dropping.js
+++ b/creep.action.dropping.js
@@ -8,8 +8,7 @@ action.isValidAction = function(creep){
 action.isValidTarget = function(target, creep){
     if (!target) return false;
     if (target instanceof Flag) {
-        return _.eq(target, FlagDir.flagFilter(FLAG_COLOR.claim.spawn)) ||
-                _.eq(target, FlagDir.flagFilter(FLAG_COLOR.command.drop));
+        return target.compareTo(FLAG_COLOR.claim.spawn) || target.compareTo(FLAG_COLOR.command.drop);
     }
     return true;
 };
@@ -22,7 +21,7 @@ action.newTarget = function(creep) {
     if( !drop ) {
         drop = creep.pos.findClosestByRange(creep.room.find(FIND_FLAGS, FlagDir.flagFilter(FLAG_COLOR.claim.spawn)));
     }
-    return drop || false;
+    return drop;
 };
 action.work = function(creep) {
     let ret = OK;
--- a/creep.action.dropping.js
+++ b/creep.action.dropping.js
@@ -5,6 +5,14 @@ action.reachedRange = 0;
 action.isValidAction = function(creep){
     return creep.sum > 0;
 };
+action.isValidTarget = function(target, creep){
+    if (!target) return false;
+    if (target instanceof Flag) {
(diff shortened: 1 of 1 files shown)


creep.action.dropping.js
```diff
@@ -14,7 +22,7 @@ action.newTarget = function(creep) {
     if( !drop ) {
         drop = creep.pos.findClosestByRange(creep.room.find(FIND_FLAGS, FlagDir.flagFilter(FLAG_COLOR.claim.spawn)));
     }
-    return drop;
+    return drop || false;
```
karlthepagan: oops, removing this it was a wild guess that isn't really defensible

chobobobo:
Two things: you can use FlagDir.flagType(flag) to get the name of a flag, perhaps compare each time a target that is a flag gets set on a creep to see exactly when the _OCS flag is set

Also, had a look at FlagDir.find() to see if it could ever be returning the _OCS flag in error.  Looked at the flagColor as a function and didn't see anything wrong there.. if flagColor is undefined it looks like it will return an empty filter, but that should be caught by the `flagColor == null` check.

flagFilter returns undefined if !flagColor is there any other way that flagColor could be falsey but not == null?

What remains is to look at the bottom ```    // some flags found - find nearest
    if( pos && pos.roomName ){
        let range = flag => {
            let r = 0;
            let roomDist = routeRange(pos.roomName, flag.roomName);
            if( roomDist == 0 )
                r = _.max([Math.abs(flag.x-pos.x), Math.abs(flag.y-pos.y)]);
            else r = roomDist * 50;
            if( mod ){
                r = mod(r, flag, modArgs);
            }
            flag.valid = r < Infinity;
            return r;
        };
        let flag = _.min(flags, range); //_.sortBy(flags, range)[0];
        return flag.valid ? flag.name : null;
    } else return flags[0].name;``` and see if it could return a bad flag

creep.action.dropping.js
```diff
@@ -5,6 +5,14 @@ action.reachedRange = 0;
 action.isValidAction = function(creep){
     return creep.sum > 0;
 };
+action.isValidTarget = function(target, creep){
+    if (!target) return false;
+    if (target instanceof Flag) {
+        return _.eq(target, FlagDir.flagFilter(FLAG_COLOR.claim.spawn)) ||
+                _.eq(target, FlagDir.flagFilter(FLAG_COLOR.command.drop));
```
Spedwards: `target.compareTo(FLAG_COLOR.claim.spawn) || target.compareTo(FLAG_COLOR.command.drop)`

karlthepagan: resolved

karlthepagan:
Probably still needs work.

karlthepagan:
I'm rethinking how strict to be about assignment targets

karlthepagan:
Special note about this change: an action's "isValidAction" method should only require that a target is a valid candidate for action. Whether or not we prefer to use that target is a decision that could be left to the target selection / user.

For example, users might manually implement a command to drop energy somewhere specific and this just broke that functionality.

PR #960, opened 2017-03-12
0.6.3 (RC1)

karlthepagan:
DOWNLOAD HERE: https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/archive/2017-03-rc1.zip

Release Candidate hotfixes:
#963 
#965 
#966 
#976 
#1078
#1082
#1117

**12 contributors! awesome community participation!**

bugfixes:
* recycler
* remote mining had issues
* pioneers could not travel without vision
* strategy handlers were inconsistent and somewhat confusing
* upgraders were not limited to 15 EPT at RCL8
* uncharging managed containers had issues
* do not pick up minerals without a storage location
* terminal broker will not run if there is no storage
* fix nextRoom selection for robbing

known issues
* creeps will (still) frequently get stuck on borders and say "NOPATH" - block off the exit or build roads as a pathfinding hint, and look for a solution in OCS 0.7.0
* "Corrupt creep without population entry" has not yet been eliminated, but should occur less frequently.

features
* automatic observer activation
* room visuals
* guards may target a border exit to guard both rooms
* military avoiding action
* energy dropping action (onto white/yellow flag)
* unified mining logic
* unified creep predicted renewal/retirement
* extension feeding and move in the same tick
* resource management API (reacllocating overhaul)
* some low bucket CPU throttling
* room.storage.charge - indicates float fraction between configured min and max energy
* travelling action attempts to move to travelPos / travelRoom
* creep.behaviour.ranger.heal is now the standard healing implementation, all creeps try to heal when fleeing and 
* optimized Creep.prototype.hasActiveBodyparts / Creep.prototype.hasBodyparts
* creeps will not try to repair neutral controller rooms
* universal object cloaking (unstable)
* buildNearby and repairNearby will work in reserved, allied and unowned rooms, but not hostile rooms
* pioneers now count against the max number of workers in a room
* upgraders no longer repair their containers, they focus on upgrading
* combat creeps priority targeting can be customized using `priotiryTargetFilter` strategy handler
* many actions now target by closest range, saving CPU
* harvesting prefers closer sources
* major update to reallocating / getNeeds

very important API changes
* Task.installTask, Task.populate
* Room.prototype.placeRoomOrder, Room.prototype.placeReactionOrder, Room.prototype.placeOrder, Room.prototype.setStore, Room.prototype.cancelReactionOrder
* Room.prototype.exits
* FlagDir.flagType

important params
* ROOM_VISUALS: true - enable room visuals
[...]

Issue #1060, opened 2017-03-26
haulers losing link target


chobobobo:
I noticed that sometimes when my remoteHaulers target a link for charging they repeatedly select the target each tick which makes me suspect that somehow they are valid to add the target, but the target becomes invalid right afterwards, perhaps there is some inconsistency there, this is likely consuming CPU cycles.

![screen shot 2017-03-26 at 8 24 51 am](https://cloud.githubusercontent.com/assets/25397533/24327212/d94dc278-11fd-11e7-80a4-523d981a2abe.png)

PR #1097, opened 2017-03-29
isValidTarget was broken for charging links

chobobobo:
Also converted to use defined constants rather than values.

This should also be an issue in https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/960 and may need to be cherrypicked.


karlthepagan:
I don't see the fix among the refactor. What part was it?

chobobobo:
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1060

It was subtle, comparing against target.storeCapacity instead of target.energyCapacity

PR #1101, opened 2017-03-29
fix flagType error print

karlthepagan:
ternary is tricky


flagDir.js
```diff
@@ -227,7 +227,7 @@ mod.flagType = function(flag) {
             }
         }
     }
-    logError('Unknown flag type for flag ' + flag ? flag.name : 'undefined flag');
+    logError('Unknown flag type for flag: ' + (flag ? flag.name : 'undefined flag'));
```
Spedwards: `logError(`Unknown flag type for flag: ${flag ? flag.name : 'undefined flag'}`);` would be another good way to write this.

PR #1103, opened 2017-03-29
privateer reserves will always use the low priority queue

karlthepagan:
This addresses @TafThorne's problem with reservers spawninig before privateers.

supercedes #621.

### DIFF
--- a/task.reserve.js
+++ b/task.reserve.js
@@ -19,9 +19,9 @@ mod.register = () => {};
 // for each flag
 mod.handleFlagFound = flag => {
     // if it is a reserve, exploit or remote mine flag
-    if( flag.color == FLAG_COLOR.claim.reserve.color && flag.secondaryColor == FLAG_COLOR.claim.reserve.secondaryColor ||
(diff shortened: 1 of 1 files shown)


task.reserve.js
```diff
@@ -19,18 +19,21 @@ mod.register = () => {};
...
+        flag.compareTo(FLAG_COLOR.claim.mining) ) {
         // check if a new creep has to be spawned
         Task.reserve.checkForRequiredCreeps(flag);
     }
 };
 // check if a new creep has to be spawned
 mod.checkForRequiredCreeps = (flag) => {
     let spawnParams;
-    if( flag.color == FLAG_COLOR.claim.mining.color && flag.secondaryColor == FLAG_COLOR.claim.mining.secondaryColor ) {
+    if( flag.compareTo(FLAG_COLOR.claim.mining) ) {
```
Spedwards: This line, and the changed above, have been changed in #1069. I'd suggest reverting these changes and stick with the ones below.

Spedwards: .

karlthepagan: resolved

PR #1109, opened 2017-03-30
Classify Tasks

Spedwards:
Turns all the tasks into an instance of the Task class. Similar to how actions are.

I'm a little worried installing tasks may be broken (specifically custom/internal ones).

In any of the `task.*` files, if you want to extend (not override) one of the instance methods defined in Task.js, I believe `super.method()` should work.


Spedwards:
Concerns:

 - Base tasks are getting installed twice.

Ideas:

 - Remove `mod.name` declaration in each tasks, and add `this.name` assignment to Task constructor.

PR #1110, opened 2017-03-30
allow robbing under owned or public ramparts

karlthepagan:
I need to do some friendly robbing today


creep.action.robbing.js
```diff
@@ -6,7 +6,15 @@ action.isValidAction = function(creep){
     return ( creep.sum < ( creep.carryCapacity * 0.95 ) && !creep.room.my);
 };
 action.isValidTarget = function(target){
-    return !_.some(target.pos.lookFor(LOOK_STRUCTURES), {'structureType':STRUCTURE_RAMPART }) && ( ( target.store && _.sum(target.store) > 20 ) || ( target.energy && target.energy > 20 ) || ( target.mineralAmount && target.mineralAmount > 20 ) );
+    if (_.some(target.pos.lookFor(LOOK_STRUCTURES), {structureType: STRUCTURE_RAMPART, isPublic: false, my: false})) {
```
Spedwards: Shouldn't it really be `!isPublic || !my`? Otherwise to rob under your own ramparts you'd need them to be public.

karlthepagan: Yes. Will have to make a bigger chain.

karlthepagan: Wait, no. I tested it. Logic is hard, here's a table to help:

structure type | isPublic | my | isValidTarget
-- | - | - | -
road | ? | ? | true
rampart | true | false | true
rampart | false | true | true
rampart | true | true | true
rampart | false | false | false

karlthepagan: The only case you can't rob something is under the condition where it is a rampart and it's *not* public and it's *not* mine. If any of those 3 checks fails for all structures under the square then you can rob.

chobobobo: So the options given in that filter are separated by OR not AND?

karlthepagan: No, it's and but they are encased by an outer "not" ... but this is the same thing

after all `!(a && b && c) === !a || !b || !c`

chobobobo: Ah yes, the outer ! Clears that all up.

PR #1111, opened 2017-03-30
Visuals profiling

Spedwards:


visuals.js
```diff
@@ -513,70 +514,94 @@ Visuals.run = function() {
         if (VISUALS.HEATMAP) {
             if (Game.time % VISUALS.HEATMAP_INTERVAL === 0) {
                 Visuals.setHeatMapData(room);
+                p.checkCPU('Heatmap.set');
```
chobobobo: for each of these, you need to set a threshold, I would also add PROFILING.VISUALS and PROFILING.VISUALS_LIMIT to parameters, if you want to go further VISUALS_LIMITS could be an object with keys for each of the limits in here.  Then this could be changed to
`p.checkCPU('Heatmap.set', PROFILING.VISUALS_LIMIT)`

chobobobo: I think I'm going to add a parameter to checkCPU so that we can use parameters to disable certain checks and fine-tune our profiling, so I would just make the change to set PROFILING.VISUALS_LIMIT for now

Spedwards: Looking at the `checkCPU` function, limit looked optional (`if (!limit || used > limit)`)

I was going to eventually set these but just wanted a baseline first.

chobobobo: Yeah, what you can do is you can set the parameter to 0 to establish a baseline, but as long as the parameters make it in we're good.

Issue #1067, opened 2017-03-26
profiler end-of-tick message is always displayed


karlthepagan:
This is very noisy when I'm trying to debug something else.

I would like to make this shown only every 5 or 10 ticks unless PROFILE: true is set.

chobobobo:
I think what I will do is add a parameter for basic profiling because I'm sure you're not the only one who feels that way and if you use visuals it's likely not valuable

PR #1119, opened 2017-03-31
Change/improve profiling

chobobobo:


chobobobo:
this will resolve #1067

population.js
```diff
@@ -249,28 +253,38 @@ mod.analyze = function(){
...
+    _.forEach(Memory.population, c => {
+        validateAssignment(c);
+        p.checkCPU('Validate: ' + c.creepName, PROFILING.ANALYZE_LIMIT / 2);
+    });
 };
 mod.execute = function(){
+    const p = startProfiling('Population.execute', {enabled: PROFILING.CREEPS});
     let triggerCompleted = name => Creep.spawningCompleted.trigger(Game.creeps[name]);
     this.spawned.forEach(triggerCompleted);
+    p.checkCPU('triggerCompleted', PROFILING.EXECUTE_LIMIT / 4);
```
karlthepagan: I don't want to require it but would precomputing all these divides / multiplies help? This might be my own superstition.

chobobobo: I think math is fairly well optimized, it just has to grab the constant and divide it right?  Fails the DRY test, so I will change it.

chobobobo:
@Spedwards I have fixed Util and made use of it globally, as well as making some significant improvements.  I think this is ready to merge now.

util.js
```diff
@@ -449,13 +459,9 @@ module.exports = {
                         Util.logSystem(name + ':' + localName, used);
                     }
                     if (type) {
-                        Util.set(global.profiler.types, type, {
-                            totalCPU: 0,
-                            count: 0,
-                            totalCount: 0,
-                        });
+                        if (_.isUndefined(global.profiler.types[type])) global.profiler.types[type] = {totalCPU: 0, count: 0, totalCount: 0};
```
Spedwards: Did ```Util.set(global.profiler, `types[${type}]`...``` not work?

chobobobo: It did not, I gave up in the end, the only issue is the . In the key, maybe I should change the flag types function to return type: subtype instead so we can use it

PR #1120, opened 2017-03-31
feature/travelerCaching - cache paths to specific destinations for reuse

chobobobo:
This feature is currently working, but needs tuning to improve efficiency.


chobobobo:
This will need path invalidation/rebuilding as well as a move to memory segments once they are available in dev.

chobobobo:
This now makes use of memory segments, still need to do invalidation.

chobobobo:
Also need to add stuck detection and resolution as well as resolve some bugs in a  few of my rooms where I notice creeps getting stuck.

chobobobo:
Ok, the current state of this PR is:
- remote haulers, miners, workers, reservers are caching paths
- cached paths are stored in segments 94 - 90
- cached paths are valid and working well now
- you can visualize paths using Room.showCachedPath(<roomName>, <destination>)
- you can invalidate paths using Room.invalidateCachedPaths(<roomName>, <destination>)

To Do:
- invalidate paths automatically
- when generating paths, get the shortest path to join the existing path
- implement simple avoidance
- reduce the need for loading the entire cache from the segment so often, we lose efficiency when this happens.

chobobobo:
feedback from @bonzai ```
-the path is cached somewhere outside of creep memory and passed in as an argument when calling travelByPath, this will let you calculate it once and reuse it with multiple creeps
-the creep keeps track of which direction it is heading and uses the same data when it needs to travel in reverse
-it will need some different kind of stuck-rerouting, maybe a temporary path that it generates to move around an obstacle to some position further up in the path (I hadn't figured this part out)
```

**Comment by chobobobo** [comment]
Currently, there appear to be cases where the path cache gets deleted by accident.

**Comment by chobobobo** [comment]
Ok, this is working, and I am running it in production, however I think I will refactor it to become an extension of traveler if I can before it gets merged in.  The recommended way to use it is by creating your own virals, I will attach examples that I am using here, and put them in documentation once the feature is in dev.
viral.creep.behaviour.remoteHauler.js *cache only paths that are rooms as targets*
```
const mod = {};
module.exports = mod;
mod.strategies = {
    defaultStrategy: {
        name: `default-${mod.name}`,
        moveOptions: function(options) {
            // enable this to cache ALL routes for this creep
            options.cacheRoutes = true;
            // use this callback to only cache specific destinations
            options.cacheThisRoute = (dest) => dest && dest.x === 25 && dest.y === 25;
            return options;
        },
    }
};
```
viral.creep.Action.js *cache all paths that end at storage*
```
const mod = {};
module.exports = mod;
mod.getDefaultStrategy = function(action) {
    const ret = action.defaultStrategy;
    ret.moveOptions = function(options) {
        // cache all routes that end at the storage
        options.cacheThisRoute = (dest) => {
            const room = Game.rooms[dest.roomName];
            return room && room.storage && room.storage.pos.isEqualTo(dest);
        };
        return options;
    };
    return ret;
};
```
viral.creep.behaviour.claimer.js *cache all paths*
```
const mod = {};
module.exports = mod;
mod.strategies = {
    defaultStrategy: {
        name: `default-${mod.name}`,
        moveOptions: function(options) {
            options.cacheRoutes = true;
            return options;
        },
    }
};
```

PR #1121, opened 2017-03-31
this should allow guards to spawn in the same room they are wanted in.

### LINKED ISSUE #1112: Yellow/Yellow flags placed in own room do not cause guard to spawn.
on dev and master branch
in a newbie area
I only control one room

when I place a yellow/yellow flag in my own room, no guard spawns.

I added console.log messages and found that task.js returns on line 81 because room is null. roomParams.explicit is undefined, and Room.findSpawnRoom(roomParams) is returning null.

If I place a yellow/yellow flag in an adjacent room, a guard spawns and moves to that flag.

chobobobo:
fixes #1112

### DIFF
--- a/task.guard.js
+++ b/task.guard.js
@@ -45,7 +45,7 @@ mod.checkForRequiredCreeps = (flag) => {
                 targetRoom: flag.pos.roomName, 
                 minEnergyCapacity: 200, 
                 rangeRclRatio: 1.8, // stronger preference of higher RCL rooms
-                allowTargetRoom: true
+                allowTargetRoom: true,
             },
             creepSetup => { // callback onQueued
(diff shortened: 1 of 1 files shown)


task.guard.js
```diff
@@ -44,7 +44,8 @@ mod.checkForRequiredCreeps = (flag) => {
             { // spawn room selection params
                 targetRoom: flag.pos.roomName, 
                 minEnergyCapacity: 200, 
-                rangeRclRatio: 1.8 // stronger preference of higher RCL rooms
+                rangeRclRatio: 1.8, // stronger preference of higher RCL rooms
+                allowTargetRoom: true
```
Spedwards: Trailing comma.

PR #1122, opened 2017-03-31
Group common parameters

chobobobo:
Also added parameter for remoteHaulers so they can have a separate repair range.


chobobobo:
Merged into my branch and tested, but will conflict with other branches, ok to merge?

PR #1124, opened 2017-03-31
Allow friendlies to pass

Spedwards:
If an allied creep says `String.fromCodePoint(0x1F6AA)` while next to one of your ramparts, the rampart will become public, allowing them to pass through.

This is only one side of the implementation. The other side has to be done in creep pathing.


room.js
```diff
@@ -2694,6 +2694,28 @@ mod.extend = function(){
...
+        // Close Ramparts
+        _(this.structures.all)                              // Iterate over all room structures
+            .filter(s => s instanceof StructureRampart)     // Filter out structures not a rampart
+            .filter(rampart => rampart.isPublic)            // Filter out any rampart already closed
+            .forEach(rampart => rampart.setPublic(false));  // Close any public rampart
+    
+        // Open Ramparts
+        _(this.allCreeps)                                                                                       // Iterate over all creeps
+            .filter(creep => Task.reputation.allyOwner(creep))                                                  // Filter out any non-friendly creeps
+            .filter(creep => creep.saying === String.fromCodePoint(0x1F6AA))                                    // Filter out any creeps not requesting access
```
karlthepagan: I think the saying check will be less expensive than the allyOwner check

Spedwards:
@DasBrain Intents will override any set prior to them, which is why I have opening after closing.

chobobobo:
Some intents will still deduct their cost, while some (move(direction) comes to mind) will allow you to call again without the additional cost.  Does isPublic fall under the latter?  If not, we should just do a check on all ramparts in this function that defaults to closed and use an if structure to filter between the two so that we only set an intent once.

chobobobo:
The other half of this PR is to consider allied ramparts as public:
```
chobobobo [7:33 AM] 
So new interesting problem, now that we have a PR to open ramparts to allies, how do I determine whether an ally will open their ramparts for me or not?

[7:33]  
More broadly.. do we have a method for asking an ally (programatically) what rights we have

[7:34]  
for pathfinding, I need to know if they are going to open their ramparts, otherwise my creeps will walk into a closed door

spedwards [7:34 AM] 
The reason Cho didn't mention is, we don't want to assume a rampart will open when the other player is running an older version without this functionality.

[7:35]  
What I suggested is that we don't worry about it and let individual players bring it up with their nearby allies.

[7:35]  
Since it's more than just that, we ask you lot :slightly_smiling_face:

chobobobo [7:36 AM] 
I suppose a quick fix would be to assume they are open, but when you walk headfirst into a wall, mark that ally as obstructive
```

creep.action.travelling.js
```diff
@@ -34,6 +34,14 @@ action.step = function(creep){
                 return action.unregister(creep);
             }
         }
+        // TODO: Only check if moving towards the rampart
+        if (_.find(creep.pos.adjacent, pos => {
```
chobobobo: I wonder if there's a way we can avoid this check in most cases, perhaps by caching whether there are any allied ramparts in room memory and only making the check if the creep is in a room with allied ramparts?  Currently, this will result in a lot of extra processing because all moving creeps will be calling this every tick.

Spedwards: Could probably do with a `Task.reputation.allyOwner(creep.room)` before the find.

creep.action.travelling.js
```diff
@@ -38,7 +38,7 @@ action.step = function(creep){
             }
         }
         // TODO: Only check if moving towards the rampart
-        if (creep.room.owner && Task.reputation.allyOwner(creep.room) && _.find(creep.pos.adjacent, pos => {
+        if (creep.room.owner && creep.room.owner.username !== ME && Task.reputation.allyOwner(creep.room) && _.find(creep.pos.adjacent, pos => {
```
chobobobo: creep.room.owner is the username

chobobobo:
Ok, I've updated this to pathfind through allied ramparts, however there's currently no method to resolve cases where the ally doesn't have the latest version of the code.

chobobobo:
K, was finally able to test this, the code in action.travelling needs to be moved to traveler.js so that any traveling creep can trigger it, and needs to be optimized as such so that it only gets triggered when absolutely necessary (it may not be possible to optimize it beyond it's current form)

My creep said the magic words, but the ramparts did not open, not sure why the message wasn't received on the ally's end.

chobobobo:
@Spedwards you might understand what I don't, why did lodash wrapping of the results not work?

Spedwards:
It probably needed a `.commit()` on the end of it. @chobobobo

PR #1125, opened 2017-03-31
Get all positions in radius

Spedwards:
Didn't end up needing this, but it still might be useful.

### DIFF
--- a/extensions.js
+++ b/extensions.js
@@ -54,7 +54,7 @@ mod.extend = function(){
         configurable: true,
         value: function(radius = 1) {
             if (radius === 1) return this.adjacent;
-            if (radius < 1) return [];
+            if (radius < 1) return [this];
             const positions = [];
             for (let x = this.x - radius; x < this.x + radius; x++) {
                 for (let y = this.y - radius; y < this.y + radius; y++) {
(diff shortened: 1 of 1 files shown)


extensions.js
```diff
@@ -50,6 +50,22 @@ mod.extend = function(){
             return this._adjacent;
         }
     });
+    Object.defineProperty(RoomPosition.prototype, 'radius', {
+        configurable: true,
+        value: function(radius = 1) {
+            if (radius === 1) return this.adjacent;
+            if (radius < 1) return [];
```
chobobobo: This is a filled radius correct?  The analogy doesn't exactly hold up, but if radius=0, wouldn't it make sense to return the roomPosition itself?

extensions.js
```diff
@@ -54,7 +54,7 @@ mod.extend = function(){
         configurable: true,
         value: function(radius = 1) {
             if (radius === 1) return this.adjacent;
-            if (radius < 1) return [];
+            if (radius < 1) return [this];
             const positions = [];
             for (let x = this.x - radius; x < this.x + radius; x++) {
```
chobobobo: for radius===1 and pos (1,1) this will loop from 0 - 1, I think you want <= right?

Spedwards: No. `< 1` is exactly right. There is a check above for `=== 1` which refers to the property above this one.

chobobobo: Ok, bad example, check radius===2, pos 2,2 that will loop from x=0-3, isn't that lopsided?  x=4 should be a part of the radius as well

Spedwards: Oh I see what you mean, yeah, it should be `<=` in the for-loops.

PR #1127, opened 2017-03-31
Disable spawning of non-critical roles

### LINKED ISSUE #818: Stop Spawning Non-Critical Roles when Bucket <= CRITICAL_BUCKET_LEVEL * 2/3
Currently well below `CRITICAL_BUCKET_LEVEL`, working on up, but my spawns are working on non-critical roles instead of critical ones that are needed to stay afloat.

While `Game.cpu.bucket <= CRITICAL_BUCKET_LEVEL * 2/3` spawns should skip over non-critical roles.

Spedwards:
Resolves #818

### DIFF
--- a/spawn.js
+++ b/spawn.js
@@ -39,11 +39,16 @@ mod.extend = function(){
     };
     Spawn.prototype.createCreepByQueue = function(queue){
         if (!queue) return null;
+        const original = _.cloneDeep(queue);
         let params;
         do {
-            if (queue.length === 0) return null;
+            if (queue.length === 0) break;
             params = queue.shift();
         } while (Memory.CPU_CRITICAL && !CRITICAL_ROLES.includes(params.behaviour));
+        if (queue.length === 0 && !params) {
+            queue = original;
+            return null;
+        }
         let cost = 0;
         params.parts.forEach(function(part){
             cost += BODYPART_COST[part];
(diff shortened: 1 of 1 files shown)


chobobobo:
Currently there is a bug with tasks that they don't track their queued creeps to see if they actually spawn or not, unfortunately this would hit on that bug big time.

Also, anyone that queues a creep will need to check to see if that creep is in the CRITICAL_ROLES list before queueing it to prevent thrashing adding/removing.

Spedwards:
@chobobobo Not a perfect change just yet, but latest commit will keep the creeps in queue, but still skip over them.

chobobobo:
Both @Spedwards and I have contributed to this one, so we're not in a position to review

spawn.js
```diff
@@ -39,6 +39,7 @@ mod.extend = function(){
     };
     Spawn.prototype.createCreepByQueue = function(queue, level){
         if (!queue) return null;
+        else if (_.isUndefined(this.memory.spawnDelay)) this.memory.spawnDelay = {};
```
Spedwards: This whole line could be `Util.set(this.memory, 'spawnDelay', {});` and then the next line down can be changed to an `if`.

karlthepagan:
Interesting to consider a few cases where critical roles are different from critical spawning:

* upgrader should only be a critical spawn if room.storage.sum > 90% capacity
* mineralMiner should only be a critical spawn if the mineral's renewal is < 10000 ticks (and storage.sum < 90% capacity)
* claimer should never be critical to spawn
* renewer should only be a critical spawn if their assigned room is < 300 reservation

karlthepagan:
ran this for 12h and nothing died on me, bucket is 10k

will try to push my bucket by adding another SK room and file new PR's on any critical issues

PR #1131, opened 2017-04-01
robbers take valuables first

karlthepagan:
This could be considered a work in progress since it deserves the strategy treatment, but I want to refactor storeCall / targetCall into the base OCS API.

I recommend split diff.

### DIFF
--- a/creep.action.robbing.js
+++ b/creep.action.robbing.js
@@ -16,36 +16,20 @@ action.isValidTarget = function(target){
 
     return false;
 };
-action.invasionAdjacentScore = 42;
-action.invasionScoreMultiplier = 1 / Math.log(1.1);
-action.normalAdjacentScore = 50;
-action.normalScoreMultiplier = 1 / Math.log(1.2);
-/*
- container - pile or structure
- */
-action.resourceRobValue = function(type) {
-    if (type === RESOURCE_ENERGY) return 0.2;
-    if (type === RESOURCE_POWER) return 500;
-    return type.length;
-};
-const resourcesDescending = _.memoize(function(action) {
-    return _.chain(global).pick(function(v,k) {
-        return k.startsWith('RESOURCE_');
-    }).sortBy(action.resourceRobValue).values().value();
-});
 action.newTarget = function(creep){
-    // TODO RESOURCE_ priority
-    // LATER after scoring iterate thru top candidates and do pathfinding to find an accessible target!
     const targetPool = creep.room.structures.all;
 
     if (targetPool.length) {
         const targets = _.chain(targetPool)
-            .map(this.targetScore(creep))
+            .map(this.targetScore(creep)) // resource fill priority
             .filter('score')
             .sortBy('score').reverse()
             .value();
 
         const target = _.get(targets, [0, 'target'], null);
+
+        // after scoring iterate thru top candidates and do pathfinding to find an accessible target!
+
 
         // console.log(creep.name, targets.length, target);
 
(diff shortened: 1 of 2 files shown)


karlthepagan:
Working pretty well on Twill's retirement: https://screeps.com/a/#!/room/E2N86

Minerals are now being emptied first.

creep.action.robbing.js
```diff
@@ -16,32 +16,109 @@ action.isValidTarget = function(target){
 
     return false;
 };
+action.invasionAdjacentScore = 42;
+action.invasionScoreMultiplier = 1 / Math.log(1.1);
+action.normalAdjacentScore = 50;
+action.normalScoreMultiplier = 1 / Math.log(1.2);
```
karlthepagan: these numbers create some greatly different curves depending on whether or not there's an invasion

Without invasion containers on the opposite end of the room are 30% less interesting than pretty close targets.

In an invasion scenario something on the other side is hundreds of times less interesting.

image

creep.action.robbing.js
```diff
@@ -16,32 +16,109 @@ action.isValidTarget = function(target){
...
+    const targetPool = creep.room.structures.all;
+
+    if (targetPool.length) {
+        const targets = _.chain(targetPool)
+            .map(this.targetScore(creep))
+            .filter('score')
+            .sortBy('score').reverse()
+            .value();
+
+        const target = _.get(targets, [0, 'target'], null);
```
karlthepagan: if this looks like a standard target priority scoring chain... that's what it is :)

creep.action.robbing.js
```diff
@@ -16,32 +16,109 @@ action.isValidTarget = function(target){
...
+            if (score) {
+                return {amount: capacity, score};
+            }
+            return {amount, score};
+        };
+    })(creep.target);
+};
+const identitySafe = function(count) {
+    return count || 0;
+};
```
karlthepagan: should go in Util

creep.action.robbing.js
```diff
@@ -16,32 +16,109 @@ action.isValidTarget = function(target){
...
+            score = this.storeCall(creep, target,
+                {[RESOURCE_ENERGY]: target.energy, [target.mineralType]: target.mineralAmount}, identitySafe, valueCallback);
+        } else {
+            score = this.storeCall(creep, target, {[RESOURCE_ENERGY]: target.energy}, identitySafe, valueCallback);
+            // TODO dropped resources are a combination of all drops under that point, their decay function relates to the number of separate resources
+            /* instead of identitySafe:
+             const valueTransform = function(count) {
+             return decayFunction(count || 0, range);
+             };
+             */
```
karlthepagan: This targetCall function should turn into a Util function which becomes a standard contents reducing function. It will be reused in many different scoring situation.

creep.action.robbing.js
```diff
@@ -16,32 +16,109 @@ action.isValidTarget = function(target){
...
+action.storeCall = function(creep, target, store, valueTransform, valueCallback, keys) {
+    let capacity = creep.carryCapacity - creep.sum;
+    let value = 0;
+    if (!keys) keys = _.keys(store);
+    for (let i = keys.length - 1; i >= 0; i--) {
+        if (capacity === 0) break;
+
+        const type = keys[i];
+        const count = Math.min(valueTransform(store[type]), capacity);
+        const {amount, score} = valueCallback(type, count < 1 ? 0 : count, capacity);
```
karlthepagan: valueCallback either just computes the value of something taken against the budget... or it actually takes it

karlthepagan:
Add this to your viral.mainInjection to restore the invasion case. I'll move it to internal.

```
const invasionScoreMultiplier = 1 / Math.log(1.1);
const normalScoreMultiplier = 1 / Math.log(1.2);

mod.extend = function() {
      // ...
      Creep.action.robbing.defaultStrategy.resourceScore = function(creep, target, resourceValue) {
        const range = creep.pos.getRangeTo(target);
        const logBase = creep.room.situation.invasion ? invasionScoreMultiplier : normalScoreMultiplier;
        const adjacentValue = creep.room.situation.invasion ? 42 : 50;
        return function(type, amount, capacity) {
            if (amount === 0) return {amount: 0, score: 0};

            const multiplier = resourceValue(type);

            // console.log(creep.name, 'resourceScore', type, amount, multiplier, range, logBase, adjacentValue);

            return {amount, score: multiplier * amount * (adjacentValue - Math.log1p(range) * logBase)};
        }
    };
};
```

Issue #1043, opened 2017-03-24
Memory out of control


Dasku:
Sometimes my memory gets out of control and not sure exactly why. It would appear the vast majority of the memory is saved in Rooms tho. 
![image](https://cloud.githubusercontent.com/assets/22677773/24289154/f02a6816-1077-11e7-9e6a-2b7d84afdacb.png)

I've noticed one thing that is definitely a waste - we save all info for every room. Surely we shouldnt save normal room info in corridor rooms?
![image](https://cloud.githubusercontent.com/assets/22677773/24289183/120f72d2-1078-11e7-8460-d98b0d008db0.png)

When you have 140+ rooms in memory this really adds up I think.

Perrytheplatypus43:
I had the same problem I had my memory to almost max so my cpu was really high. My cpu bucket was empty at gcl 3.

Dasku:
This cant be the issue, it must be something being saved in the rooms somewhere as I'm back up to same rooms but with half as much memory used.

chobobobo:
@Dasku we now have the memoryUsage(key) helper function to support you, separate your keys with . eg, memoryUsage('Task.mining')  I would suspect Memory.rooms.statistics.invaders as one likely culprit, another possibility it roadConstructionTrace if you have auto roads turned on, then there's Task.defense as well.

Dasku:
![image](https://cloud.githubusercontent.com/assets/22677773/24333532/544a8dc8-1251-11e7-8a32-a09903f3db41.png)

I think this is a big contributor over time

chobobobo:
If you have ever turned on visuals and heatmap they are about the worst offenders I've seen yet

chobobobo:
@Dasku so you're aware, I'm working on a patch for this now

chobobobo:
That patch is for rooms and hopefully invader statistics, but Memory.population also needs a look at as it scales so massively by the number of creeps.  Preliminary investigation points to idleMove storing full roomPositions as a potential area for improvement..

PR #1132, opened 2017-04-01
Change/memory reduction

chobobobo:
Attempt to reduce memory footprint (and I couldn't help cleaning up a bit of code along the way)

Primarily focused on not setting memory in rooms that don't have things like towers, observers.
Also looking to fix statistics.invaders balooning


chobobobo:
this is to address https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1043

room.js
```diff
@@ -417,11 +417,11 @@ mod.extend = function(){
...
                 configurable: true,
                 get: function() {
-	                if (_.isUndefined(this.room.memory.observer)) {
+	                if (_.isUndefined(this.room.memory.observers)) {
 		                this.room.saveObserver();
 	                }
-                    if (_.isUndefined(this._observer)) {
-	                    this._observer = Game.getObjectById(this.room.memory.observer.id);
+                    if (_.isUndefined(this._observer) && this.room.memory.observers) {
+	                    this._observer = Game.getObjectById(this.room.memory.observers[0]);
```
karlthepagan: You could set _observers = false in the case that there's nothing in the memory structure after a save.

room.js
```diff
@@ -122,7 +122,7 @@ mod.extend = function(){
                 get: function() {
                     if( _.isUndefined(this._controller) ){
                         let byType = c => c.controller === true;
-                        this._controller = this.all.filter(byType);
+                        this._controller = _.filter(this.all, byType);
```
Spedwards: Not a fan of the changes from native JS to Lodash. A benchmark test would be required before I can approve these changes.

room.js
```diff
@@ -1190,216 +1190,208 @@ mod.extend = function(){
...
     };
     Room.prototype.saveObserver = function() {
-        this.memory.observer = {};
-        [this.memory.observer.id] = this.find(FIND_MY_STRUCTURES, {
+        const observers = this.find(FIND_MY_STRUCTURES, {
             filter: s => s instanceof StructureObserver
-        }).map(s => s.id);
+        });
+        if (observers.length > 0) {
+            this.memory.observers = observers.map(o => o.id);
```
Spedwards: This breaks observer functionality **HEAVILY**. If you're making any changes to observers, this needs to be an object, not an array.

room.js
```diff
@@ -2627,7 +2619,7 @@ mod.extend = function(){
     Room.prototype.controlObserver = function() {
         const OBSERVER = this.structures.observer;
         if (!OBSERVER) return;
-        if (!this.memory.observer.rooms) this.initObserverRooms();
+        if (!this.memory.observers.rooms) this.initObserverRooms();
```
Spedwards: With your changes above, this will no longer work. You've changed `this.memory.observers` into an array, thus `rooms` will never exist.

Either fix every instance below, or revert all observer changes above. There will only ever be one observer per room, so I don't understand these changes either.

room.js
```diff
@@ -2655,7 +2647,7 @@ mod.extend = function(){
                     break;
                 }
             } while (Memory.observerSchedule.includes(nextRoom) || nextRoom in Game.rooms);
-            this.memory.observer.lastLookedIndex = lastLookedIndex;
+            this.memory.observers.lastLookedIndex = lastLookedIndex;
```
Spedwards: Same as `this.memory.observers.rooms`

chobobobo:
Looking again at this in the morning, I think I need to redo it and focus on one change at a time, this PR lost focus.

chobobobo:
@karlthepagan the only remaining concern I have for this, is that it now tries to save* on each tick that the structure is accessed.  Eg. room.structures.containers.all triggers a saveContainers where previously it would notice the empty entry in memory and avoid that.  I can simply remove those checks.. but then I need to add a check for new rooms to make sure they saveStructures properly on first vision/memory rebuilding

torin555:
error with SEND_STATISTIC_REPORTS set to true:
```TypeError: Cannot read property 'controllerProgress' of undefined
    at processRoom (statistics:40:87)
    at Object.mod.process (statistics:96:7)
    at Object.module.exports.loop (main:328:32)
    at __mainLoop:1:52```

PR #1134, opened 2017-04-02
collapseWorker doesn’t protect from a true collapse!

chobobobo:


karlthepagan:
I can convince of a future where it's not.

PR #1136, opened 2017-04-02
infectiousMemory - Change parameters in memory without reloading all code

chobobobo:
Set values in Memory.parameters which will override values in parameter.js.  Ideally, Memory.parameters should be deleted whenever new code is pushed, but currently we can't detect the difference between a code push and an instance switch.

### DIFF
--- a/main.js
+++ b/main.js
@@ -262,8 +262,11 @@ module.exports.loop = function () {
     if (Memory.cloaked === undefined) {
         Memory.cloaked = {};
     }
+    
+    Util.set(Memory, 'parameters', {});
+    _.assign(global, {parameters: Memory.parameters}); // allow for shorthand access in console
     // ensure up to date parameters, override in memory
-    _.assign(global, load("parameter", Memory.parameters));
+    _.assign(global, load("parameter", parameters));
    
     // process loaded memory segments
     OCSMemory.processSegments();
--- a/main.js
+++ b/main.js
@@ -260,7 +260,8 @@ module.exports.loop = function () {
     }
     // ensure up to date parameters
     _.assign(global, load("parameter"));
-    
+    _.assign(global, Memory.parameters);
+   
     // process loaded memory segments
     OCSMemory.processSegments();
     p.checkCPU('processSegments', PROFILING.ANALYZE_LIMIT);
--- a/main.js
+++ b/main.js
@@ -93,7 +93,8 @@ global.infect = (mod, namespace, modName) => {
 };
 // loads (require) a module. use this function anywhere you want to load a module.
 // respects custom and viral overrides
-global.load = (modName) => {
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -258,9 +262,9 @@ module.exports.loop = function () {
     if (Memory.cloaked === undefined) {
         Memory.cloaked = {};
     }
-    // ensure up to date parameters
-    _.assign(global, load("parameter"));
-
```
Spedwards: That space needs to stay. Now there's a 3 space indentation.

chobobobo:
This didn't seem to work for base level parameters eg. PROFILE

Spedwards:
See #1158 for an improved method.

main.js
```diff
@@ -266,7 +262,8 @@ module.exports.loop = function () {
     Util.set(Memory, 'parameters', {});
     _.assign(global, {parameters: Memory.parameters}); // allow for shorthand access in console
     // ensure up to date parameters, override in memory
-    _.assign(global, load("parameter", parameters));
+    _.assign(global, load("parameter"));
+    _.merge(global, parameters);
```
chobobobo: Isn't this currently assigning Memory.parameters to global.parameters, and the assigning the contents of parameters to Global.parameters, and then merging?  Logically it seems like it should be assigning Memory.parameters and then merging

Spedwards: It's merging the contents of `parameters` into `global`. Not `global.parameters`.

chobobobo: Ok, so can this then be condensed to
```
_.assign(global, load("parameter"))
_.merge(global, Memory.parameters)
```?

Spedwards: Yes. But I set the `parameters` global variable as a shorthand for console.

chobobobo: It gets overwritten each tick with​ the contents of parameters and you can current type the parameter name into the console to see the value,  what usage are you considering?

Spedwards: In console, `parameters.SAY_PUBLIC = false;`

It should be in `global.install` though.

chobobobo: Currently that would get immediately overwritten when the parameters are reloaded, `Memory.parameters.SAY_PUBLIC = false` would work however, global parameters is tricky because they are inconsistent between server instances

Spedwards: Actually, it won't matter (cost-wise) as it is right now.

`parameters` is only assigned a reference to `Memory.parameters`. Not the value. Cheap assignment, to make for a QOL accessor in console.

chobobobo: K, I hadn't realized it was a reference, that makes sense now.

karlthepagan:
let's keep iterating on this

chobobobo:
This works as advertised now, profiling may take an extra tick to take effect because of where in the code parameters get set, but that's acceptable.

PR #1137, opened 2017-04-02
Feature/room layout

Spedwards:
Behold!

Layout Image

The layout can be seen in the image above. The construction sites are all roads, and the rest can be figured out from the construction flags. This layout only includes extensions, spawns, and towers.

TODO:

 - ~Move layout out of `Room.roomLayout` so it can be viral'd separately.~


Spedwards:
It will be incredibly difficult to support any non-square layout, so if you change your layout to a design where `height !== width`, you will also have to change `Room.roomLayout`.

Spedwards:
[Simulation Test Code](https://gist.github.com/Spedwards/c69f768befe9cba04dc2902b5be4ee1a)

Open up simulation, paste the entire code in.

Tests:

 - Walls obstruct. Flag is replaced with White/Red
 - No obstruction. Flag is removed, layout is placed.

Run:

1. Place a White/Purple flag in the room.
2. In console, run `roomLayout({name: 'Flag1',color: COLOR_WHITE, secondaryColor: COLOR_PURPLE});`

karlthepagan:
_OCS is white/purple. How about white/cyan?

karlthepagan:
Please double check me. Also I guess we need to set a few color reservations:

White/white - user coded (executable?) Flags
White/purple? - API coded (special marker) flags.

chobobobo:
You know what would be really badass is if there were visuals once you put down the white/cyan so you could easily see whether your plan will work.

chobobobo:
Did some tests, placed the flags and roads very well, doesn't seem to handle failures currently, and if my current construction sites is nearing 100, not all of the roads get placed, perhaps it should keep the white/cyan in place until all roads have been placed?
screen shot 2017-04-09 at 6 21 09 pm
screen shot 2017-04-09 at 6 26 53 pm

PR #1138, opened 2017-04-02
quick fix

karlthepagan:


chobobobo:
Yeah it was in reduceMemory, wrong place for the fix!

PR #1141, opened 2017-04-02
Action comments

Spedwards:
Makes it a little easier to identify what action the creep is performing based on the statement.


karlthepagan:
Unicode comments :joy:

PR #1145, opened 2017-04-03
Feature/SpiralIterator

karlthepagan:
The TS version of this is pretty well tested. 

The dirTransform step possibly could be replaced using the method that Traveler.js does (arrays)

### DIFF
--- a/util.xyIterator.js
+++ b/util.xyIterator.js
@@ -1,0 +1,109 @@
+/**
+ * An endless iterator which advances in a square box beginning at the top-left.
+ *
+ *     67 68 69 70 71
+ *  19 P
+ *  18 O  9  A  B  C
+ *  17 N  8  1  2  D
+ *  16 M  7  0  3  E
+ *  15 L  6  5  4  F
+ *  14 K  J  I  H  G
+ *
+ *  Easy starting-points are 0, 1, A, etc. Use the ring number to indicate these.
+ */
+class XYIterator {
+    // public y;
+    // public x;
+    // private _ring;
+    // private _dir = TOP; // dir of next step
+    // private _step;
+
+    constructor(xy, ring) { // TODO borders
+        this._dir = TOP;
+        this.x = xy.x;
+        if (ring === undefined) {
+            this._ring = 0;
+            this._step = 2;
+            this.y = xy.y + 1;
+        } else {
+            this._ring = ring - 1;
+            this._step = 1;
+            this.y = xy.y;
+        }
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


karlthepagan:
@chobobobo  and I were discussing find vs lookForAt and the engine code definitely makes it seem that lookForAt will save us a lot for driveby repairs.

This tool will let you iterate in a circular spiral out from the center.

util.xyIterator.js
```diff
@@ -0,0 +1,109 @@
...
+ *  19 P
+ *  18 O  9  A  B  C
+ *  17 N  8  1  2  D
+ *  16 M  7  0  3  E
+ *  15 L  6  5  4  F
+ *  14 K  J  I  H  G
+ *
+ *  Easy starting-points are 0, 1, A, etc. Use the ring number to indicate these.
+ */
+class XYIterator {
```
Spedwards: Load this into `Util`, same as the DiamondIterator

util.xyIterator.js
```diff
@@ -0,0 +1,109 @@
...
+        this._dir = TOP;
+        this.x = xy.x;
+        if (ring === undefined) {
+            this._ring = 0;
+            this._step = 2;
+            this.y = xy.y + 1;
+        } else {
+            this._ring = ring - 1;
+            this._step = 1;
+            this.y = xy.y;
```
Spedwards: For simplicity's sake, move `this.y = xy.y` before the if-statement, then inside `ring === undefined`, increment the value (`this.y++`)

util.xyIterator.js
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        return {
+            done: false,
+            value: {x: this.x, y: this.y},
+        };
+    }
+
+    static dirTransform(origin, dir)
+    {
```
Spedwards: Should not be on it's own line.

util.xyIterator.js
karlthepagan: no demonstrable value

many devs have picked up [SSA habits](https://en.wikipedia.org/wiki/Static_single_assignment_form) and will argue to death the opposite

util.xyIterator.js
karlthepagan: I'm also going to rename this to spiral.iterator

karlthepagan:
The only other thing I could wish for this is to add variations which let you advance thru different directions, but that's more effort than its worth.

chobobobo:
How does this know when to stop?

karlthepagan:
This one never stops. That's probably an important API feature.

PR #1146, opened 2017-04-03
improve feeding performance

chobobobo:


room.js
```diff
@@ -343,6 +343,15 @@ mod.extend = function(){
                     return this._urgentRepairableSites;
                 }
             },
+            'feedable': {
+                configurable: true,
+                get: function() {
+                    if (_.isUndefined(this._feedable)) {
+                        this._feedable = this.extensions.concat(this.spawns);
```
karlthepagan: the savings is probably from extensions and spawns using FIND_MY_STRUCTURES instead of FIND_STRUCTURES

chobobobo: They actually both load from memory caches which is the big savings.

chobobobo:
Ah, this has a significant risk that new extensions and spawns will not be detected until the saveExtensions and saveSpawns is run again.. it requires more granular detection of structure changes.

karlthepagan:
Let's save this for 0.8.0

Caching optimization makes me slightly reserved

karlthepagan:
Changed my mind. We can put this in a minor revision but it should be tested separately.

PR #1148, opened 2017-04-03
Pause all code

Spedwards:


karlthepagan:
talked about this in chat. I've found myself doing this and it helps with triage

PR #1151, opened 2017-04-03
allows withdrawing to select a terminal any time it is overcharged

karlthepagan:
This could interfere with reallocating. Going to look at structure.getNeeds as the basis for charge in a terminal.


karlthepagan:
when the haulers were removing energy it reduced the order count so this won't interfere with getNeeds for the overflow case

The case where we are asked to put energy into the terminal is different. I'm considering forcing terminals to negative charge scale to resolve this.

karlthepagan:
Merged for security reasons

PR #1152, opened 2017-04-04
Heavier Remote Workers

chobobobo:


karlthepagan:
We will need a method for users to viral and specify to suppress miners based on the presence of a worker. Maybe also just start with a worker.

Spedwards:
Upon attempt to add to Canary:
```
Error in LiteEvent.trigger: TypeError: Task.mining.strategies.miner.shouldSpawn is not a function
    at Object.mod.checkForRequiredCreeps (task.mining:175:37)
    at Object.mod.handleFlagFound.flag [as handleFlagFound] (task.mining:33:21)
    at Flag.found.on.flag (task:40:62)
    at handlers.slice.forEach.h (global:35:49)
    at Array.forEach (native)
    at trigger (global:35:36)
    at triggerFound (flagDir:234:24)
    at Array.forEach (native)
    at Object.mod.execute (flagDir:238:15)
    at Object.module.exports.loop (main:313:13)  
```

chobobobo:
Should be resolved now.

PR #1155, opened 2017-04-04
Fix reservers spawning when we lack visibility

### LINKED ISSUE #1147: Large body reservers try to go past 5000 ticks
This might be a general bug with reservation (and would affect privateer red/green and mining green/brown flag reserves) but I believe it is specific to the green/grey flag.

https://screeps.com/a/#!/history/E6N83?t=18436600 this room has a green/grey flag

chobobobo:
This needs testing to verify, resolves  #1147


chobobobo:
Looks promising, been doing spot checks of my reservers, no rooms are losing reservation and I haven't discovered any trying to reserve past 5k

chobobobo:
I consider this resolved, as it has been running well for a few days now.

PR #1157, opened 2017-04-04
Feature/it's classified! (Classify Actions)

Spedwards:
Supersedes #1033.

Created the PR mostly for tracking.

As evident by the `tower.Action` file, this will also move tower code into individual actions.


action.js
```diff
@@ -0,0 +1,79 @@
...
+    
+    getTargetByID(id) {
+        return Game.getObjectById(id) || Game.creeps[id];
+    };
+    
+    isValidAction(object) {
+        return true;
+    };
+    
+    isValidTarget(targt, object) {
```
chobobobo: typo

creep.Action.js
```diff
@@ -1,125 +1,57 @@
...
-            var workResult = this.work(creep);
-            if( workResult != OK ) {
+        if (range <= this.targetRange) {
+            const workResult = this.work(creep);
+            if (workResult !== OK) {
                 const tryAction = creep.action;
                 const tryTarget = creep.target;
                 creep.action = null;
                 creep.target = null;
                 creep.handleError({errorCode: workResult, action: this, target: creep.target, range, creep});
```
chobobobo: creep.target will always be null, this should probably be `target: tryTarget` and `action: tryAction` right?

creep.Action.js
```diff
@@ -1,125 +1,57 @@
...
                 return;
             }
-            range = creep.pos.getRangeTo(creep.target); // target may have changed (eg. hauler feed+move/tick)
+            range = creep.pos.getRangeTo(creep.target);
```
chobobobo: that comment explained why we are getting range twice, it avoids someone thinking they can refactor it out.

creep.Action.js
```diff
@@ -1,125 +1,57 @@
...
-            return options || {};
-        }
+    
+    registerAction(creep, target) {
+        Population.registerAction(creep, this, target);
+        this.onAssignment(creep, target);
     };
-    // strategy accessor
-    this.selectStrategies = function() {
-        return [this.defaultStrategy];
```
chobobobo: Where have the action strategy resolution code moved to if they are removed from here?  They should be set in the constructor I believe.

creep.Action.js
Spedwards: Or moved up a couple lines. Wasn't part of my changes but I'll make it one.

creep.Action.js
Spedwards: Comments only got removed when I was off in another world while typing. Will try and go back and add any missing ones at the end.

action.js
```diff
@@ -0,0 +1,79 @@
...
+        if (target && this.isAddableTarget(object, target)) {
+            if (!object.action || object.action.name !== this.name || !object.target || object.target.id !== target.id || object.target.name !== target.name) {
+                this.registerAction(object, target);
+            }
+            return true;
+        }
+        return false;
+    };
+    
+    registerAction(object, target) {
```
karlthepagan: This creates prototype inheritance, right?

I'll be very interested in how this behaves with virals. See https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/413 where my request to do class like inheritance was previously rejected.

PR #1158, opened 2017-04-04
Merge Objects on Inject

Spedwards:
Does the same thing without recursively calling itself.


karlthepagan:
I would like this in 0.8.0

PR #1165, opened 2017-04-05
Fix debounce to work properly

chobobobo:
This will require corresponding changes to labTech https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/1045 but fixes debounce to correctly calculate anticipated energy in the creep after withdrawing, which is required for some actions, like charging.

### DIFF
--- a/creep.action.withdrawing.js
+++ b/creep.action.withdrawing.js
@@ -20,18 +20,31 @@ action.onAssignment = function(creep, target) {
     //if( SAY_ASSIGNMENT ) creep.say(String.fromCharCode(9738), SAY_PUBLIC);
     if( SAY_ASSIGNMENT ) creep.say(ACTION_SAY.WITHDRAWING, SAY_PUBLIC);
 };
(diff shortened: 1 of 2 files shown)


creep.action.withdrawing.js
```diff
@@ -21,18 +23,34 @@ action.onAssignment = function(creep, target) {
...
+        const dummyCreep = {
+            carry:{},
+            owner: creep.owner,
+            pos: creep.pos,
+            room: creep.room,
+            sum: creep.carryCapacity
+        };
+        const stored = creep.room.storage.store[RESOURCE_ENERGY];
+        const maxWithdraw = stored > creep.carryCapacity ? creep.carryCapacity : stored;
+        dummyCreep.carry[RESOURCE_ENERGY] = maxWithdraw; // assume we get a full load of energy
```
karlthepagan: I think this is fine for 0.7 thanks for the hard work.

PR #1167, opened 2017-04-05
Fix/attack controller

chobobobo:
attackController had a few issues that this should resolve.

It sets a maxRange on the spawn so your creeps don't spawn so far away that they immediately invalidate and trigger a new spawn. It seems that maxRange was completely missing, even though it was listed in the comments?

Correctly allow it to target either reservation or downgrade.

### DIFF
--- a/creep.action.attackController.js
+++ b/creep.action.attackController.js
@@ -1,7 +1,7 @@
 let action = new Creep.Action('attackController');
 module.exports = action;
 action.isValidAction = function(creep){ return true; }; 
-action.isValidTarget = function(target){ return target && (!target.reservation || target.reservation !== ME) && creep.flag; };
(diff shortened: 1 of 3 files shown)


creep.action.attackController.js
```diff
@@ -1,10 +1,10 @@
 let action = new Creep.Action('attackController');
 module.exports = action;
 action.isValidAction = function(creep){ return true; }; 
-action.isValidTarget = function(target){ return target && (!target.reservation ); };
+action.isValidTarget = function(target){ return target && (!target.reservation || target.reservation !== ME) && creep.flag; };
```
Spedwards: `target.reservation.owner` I believe. Either way, this should probably be `Task.reputation.allyOwner(target.reservation)` instead of "Not mine"

chobobobo: If someone puts an attackController flag in an allied room, I think they know what they're doing right?  target.reservation.owner is right though..

PR #1171, opened 2017-04-05
Extractor Flag Fix

### LINKED ISSUE #1105: Extractor Construction Flag
Need to figure out more on this issue.

How it works:

1. A temporary flag is placed on the extractor.
2. The flag is passed through a function used for the position.
3. Construct the extractor, and remove the flag.

---

What's happening:

1. A temporary flag is placed on the extractor.

---

Plan(s) to fix:

1. Allow a position to be passed instead of a flag to the function. Flags appear not to be created until the next tick.

Spedwards:
Resolves #1105


room.js
```diff
@@ -1495,16 +1495,18 @@ mod.extend = function(){
         const CONSTRUCT = (flag, type) => {
             if (sitesSize >= 100) return;
             if (!flag) return;
-            flag = Game.flags[flag.name];
-            const POS = flag && flag.pos;
+            const POS = new RoomPosition(flag.x, flag.y, flag.roomName);
             if (!POS) return;
```
karlthepagan: redundant check

karlthepagan:
I've seen this extractor flag

room.js
Spedwards: Forgot to remove. Doesn't really matter.

PR #1173, opened 2017-04-05
Canary

xyzzy529:
Changes to make Ranger heal either Self or Nearby.

Changed method name to .medic to reduce possible confusion on the Game method .heal()


xyzzy529:
This did not do what I thought it would.

Let me try again.

PR #1182, opened 2017-04-07
Supporting Class Virals

Spedwards:
This change will assign the base module to the action, so viral classes can extend it. This has zero affect to the old/current system.

One important change is that `Creep.action` is undefined at the time of assigning so we have to add `_.assign(Creep, { action: {} });` before to define it.

Examples of class virals can be found on issue #1186.

### DIFF
--- a/main.js
+++ b/main.js
@@ -93,7 +93,7 @@ global.infect = (mod, namespace, modName) => {
 };
 // loads (require) a module. use this function anywhere you want to load a module.
 // respects custom and viral overrides
-global.load = (modName, object) => {
+global.load = (modName, object, name) => {
     // read stored module path
     let path = getPath(modName);
     // try to load module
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -93,23 +93,24 @@ global.infect = (mod, namespace, modName) => {
...
+        // try to load module. Log error to console
         mod = tryRequire(path);
     }
-    if( mod ) {
-        // load viral overrides 
-        mod = infect(mod, 'internalViral', modName);
-        mod = infect(mod, 'viral', modName);
+    if (mod) {
+        // load viral overrides
+        object = mod;
```
Spedwards: When `object` exists, you will assign the base module to it. This will allow you to correctly extend the class you are creating a viral for. Previously it would not exist.

main.js
```diff
@@ -158,34 +159,34 @@ global.install = () => {
...
-            recycling:load("creep.action.recycling"),
-            repairing: load("creep.action.repairing"),
-            reserving: load("creep.action.reserving"),
-            robbing:load("creep.action.robbing"),
-            storing: load("creep.action.storing"),
-            travelling: load("creep.action.travelling"),
-            uncharging: load("creep.action.uncharging"),
-            upgrading: load("creep.action.upgrading"), 
-            withdrawing: load("creep.action.withdrawing"),
+            attackController: load("creep.action.attackController", Creep.action.attackController),
```
Spedwards: Pass through the object to assign the base module.

PR #1183, opened 2017-04-07
Not entirely surprised, but healing never worked at range.

chobobobo:


creep.action.healing.js
```diff
@@ -1,5 +1,6 @@
 let action = new Creep.Action('healing');
 module.exports = action;
+action.targetRange = 3;
```
karlthepagan: not targetRange=1; enoughRange=3?

good progress tho - I'm going to put this in a list of things that could make 0.6.4

chobobobo: there's targetRange and reachedRange, this will cause work to start happening at 3, but the creep will still try to move to range 1 right?

PR #1178, opened 2017-04-06
remote haulers and links fix

Vykook:
this fix prevents hauler from another unload destination, if he alredy
started unloading to link.

PR #1184, opened 2017-04-07
remote haulers and links fix

Vykook:
this fix prevents hauler from another unload destination, if he alredy
started unloading to link.

### DIFF
--- a/creep.behaviour.remoteHauler.js
+++ b/creep.behaviour.remoteHauler.js
@@ -27,9 +27,11 @@ mod.nextAction = function(creep){
             if( creep.room.storage ) deposit.push(creep.room.storage);
             // containers?
             if( creep.room.structures.container ) deposit = deposit.concat( creep.room.structures.container.privateers );
+            if( creep.memory.lastLink ) deposit.push(Game.getObjectById(creep.memory.lastLink));
             // Choose the closest
             if( deposit.length > 0 ){
                 let target = creep.pos.findClosestByRange(deposit);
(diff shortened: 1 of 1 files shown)


Vykook:
fixing #1178 PR.

creep.behaviour.remoteHauler.js
```diff
@@ -27,9 +27,11 @@ mod.nextAction = function(creep){
             if( creep.room.storage ) deposit.push(creep.room.storage);
             // containers?
             if( creep.room.structures.container ) deposit = deposit.concat( creep.room.structures.container.privateers );
+            if( creep.memory.lastLink ) deposit.push(Game.getObjectById(creep.memory.lastLink));
```
Spedwards: Move this below the `deposit = links` line.

Spedwards: By what you've put in the PR description, it sounds like all you really need is the following:

```
if (creep.data.lastLink) {
    let target = Game.getObjectById(creep.data.lastLink);
    if (target && target.pos.inRangeTo(creep.pos, 2) && this.assign(creep, Creep.action.charging, target)) {
        return;
    } else {
        delete creep.data.lastLink;
    }
}
```

Then you could remove this line and line 49.

That will avoid the `findClosestByRange` call and should also serve your purpose.

karlthepagan:
If these changes are addressed I would consider this for 0.7

Link performance at RCL5-6 needs this improvement.

chobobobo:
@Vykook please address the requested changes so that this can be tested and merged

PR #1192, opened 2017-04-08
Creep.prototype.assignAction

Spedwards:


extensions.js
```diff
@@ -392,6 +392,70 @@ mod.extend = function(){
...
+            },
+        },
+        flagColour: {
+            get: function() {
+                return _.get(FLAG_COLOR, this);
+            },
+        },
+        flagColor: {
+            get: function() {
+                return this.flagColour;
```
Spedwards: I'm not completely inconsiderate.

creep.js
```diff
@@ -3,6 +3,10 @@ const strategy = load("strategy");
 let mod = {};
 module.exports = mod;
 mod.extend = function(){
+    Creep.prototype.assignAction = function(action, target) {
```
karlthepagan: If this is for the CLI, why not allow action to be a string?

Spedwards: Wasn't necessarily for the console, but it's a simple addition.

PR #1194, opened 2017-04-08
Cleanup pickuphotfix

jschavey:
This picks up a memory hotfix published by choco but never integrated


jschavey:
Misclick - here: #1174

Issue #1176, opened 2017-04-06
Task queued validation


chobobobo:
Currently our tasks don't ever validate their 'queued' list of creeps, so if for some reason they get removed from the spawn queue, the task will get stuck in a bad state.  We need to have some kind of validation from time to time to ensure we can recover from this.

PR #1196, opened 2017-04-08
taskValidation - move spawn queue validation back into task.js

chobobobo:
Every task had this done nearly identically.  This is intended as a step towards unifying creep management into task.js, but a needed fix for stale queued entries in some tasks.
this should help to resolve https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1176


chobobobo:
Getting infinite reservers spawning on a mining flag at the moment, need to triage.

karlthepagan:
I worked pretty hard to get adding/removing creeps from the game to be O(1) in the delivery task.

Validation is absolutely needed to safeguard the system's integrity but I hope that we can reduce it to once every 50 ticks or something soon.

chobobobo:
The end goal of this is to unify the task spawning in one place and eliminate the need for such validation checks entirely, O(1) should be our goal.  In the short term this will likely not be as efficient, but it will have a stronger guarantee of accurate spawning.

task.js
```diff
@@ -106,4 +106,48 @@ mod.spawn = (creepDefinition, destiny, roomParams, onQueued) => {
     if( onQueued ) onQueued(creepSetup);
     return creepSetup;
 };
+mod.validateQueued = function(memory, queues = ['Low']) {
+    let validated = [];
```
Spedwards: `const validated = [];`

chobobobo:
Currently this only validates queued on a 50 tick cooldown because that's where we see the problems.  I have avoided converting task.mining checkSpawning because it looks a bit more complicated than the others and I want to keep the scope of this PR as small as possible.  A larger refactor will follow in 0.8.0

karlthepagan:
good direction. I think the next steps should be held for 0.7.1

we can make some more daring changes to reduce the iterations down to a single run thru the list of creeps and structures

chobobobo:
This is stable in canary now, and I consider the changes complete.

karlthepagan:
I got a flag removal error #1288 while testing this

PR #1197, opened 2017-04-08
wrap main in try/catch

jschavey:
without try/catch, thrown exception halt script execution, collapsing
colonies

### DIFF
--- a/flagDir.js
+++ b/flagDir.js
@@ -201,44 +201,51 @@ mod.flush = function(){
     delete this._hasInvasionFlag;
 };
 mod.analyze = function(){
-    let register = flag => {
-        flag.creeps = {};
-        if( flag.cloaking && flag.cloaking > 0 ) flag.cloaking--;
-        this.list.push({
-            name: flag.name,
-            color: flag.color,
-            secondaryColor: flag.secondaryColor,
-            roomName: flag.pos.roomName,
-            x: flag.pos.x,
-            y: flag.pos.y,
-            cloaking: flag.cloaking
-        });
-    };
-    _.forEach(Game.flags, register);
-    
-    let findStaleFlags = (entry, flagName) => {
-        if(!Game.flags[flagName]) {
-            this.stale.push(flagName);
-        }
-    };
-    _.forEach(Memory.flags, findStaleFlags);
-    const specialFlag = mod.specialFlag(true);
-    return !!specialFlag;
+    try {
+        let register = flag => {
+            flag.creeps = {};
+            if( flag.cloaking && flag.cloaking > 0 ) flag.cloaking--;
+            this.list.push({
+                name: flag.name,
+                color: flag.color,
+                secondaryColor: flag.secondaryColor,
+                roomName: flag.pos.roomName,
+                x: flag.pos.x,
+                y: flag.pos.y,
+                cloaking: flag.cloaking
+            });
+        };
+        _.forEach(Game.flags, register);
+        
+        let findStaleFlags = (entry, flagName) => {
+            if(!Game.flags[flagName]) {
+                this.stale.push(flagName);
+            }
+        };
+        _.forEach(Memory.flags, findStaleFlags);
+        const specialFlag = mod.specialFlag(true);
+        return !!specialFlag;
+    } catch (e) {
+        Util.logError(e.stack || e.message);
+    }
 };
 mod.execute = function() {
+    try {
+        let triggerFound = entry => {
+            if( !entry.cloaking || entry.cloaking == 0) {
+                const p = Util.startProfiling('Flag.execute', {enabled:PROFILING.FLAGS});
+                const flag = Game.flags[entry.name];
+                Flag.found.trigger(flag);
+                p.checkCPU(entry.name, PROFILING.EXECUTE_LIMIT, mod.flagType(flag));
+            }
+        };
+        this.list.forEach(triggerFound);
 
-    let triggerFound = entry => {
-        if( !entry.cloaking || entry.cloaking == 0) {
-            const p = Util.startProfiling('Flag.execute', {enabled:PROFILING.FLAGS});
-            const flag = Game.flags[entry.name];
-            Flag.found.trigger(flag);
-            p.checkCPU(entry.name, PROFILING.EXECUTE_LIMIT, mod.flagType(flag));
-        }
-    };
-    this.list.forEach(triggerFound);
-
-    let triggerRemoved = flagName => Flag.FlagRemoved.trigger(flagName);
-    this.stale.forEach(triggerRemoved);
+        let triggerRemoved = flagName => Flag.FlagRemoved.trigger(flagName);
+        this.stale.forEach(triggerRemoved);
+    } catch (e) {
+        Util.logError(e.stack || e.message);
+    }
 };
 mod.cleanup = function(){
     let clearMemory = flagName => delete Memory.flags[flagName];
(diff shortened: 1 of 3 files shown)


chobobobo:
To be clear, this will still cause colonies to collapse as the main loop execution will halt as soon as an exception is thrown, but it will allow you to modify Memory from the console as it will still serialize.

main.js
```diff
@@ -242,101 +242,108 @@ let cpuAtFirstLoop;
...
+        if ( GRAFANA && Game.time % GRAFANA_INTERVAL === 0 ) Grafana.run();
+        p.checkCPU('grafana', PROFILING.EXECUTE_LIMIT);
+    
+        Game.cacheTime = Game.time;
+    
+        if( DEBUG && TRACE ) trace('main', {cpuAtLoad, cpuAtFirstLoop, cpuAtLoop, cpuTick: Game.cpu.getUsed(), isNewServer: global.isNewServer, lastServerSwitch: Game.lastServerSwitch, main:'cpu'});
+        totalUsage.totalCPU();
+    }
+    catch (e) {
+        console.log(e.message);
```
chobobobo: I think this should be replaced with `console.log('<span style="color:FireBrick">Main' + (e.stack || e.toString()) + '</span>');`

Spedwards: Rather, `Util.logError(e.stack || e.message)`

jschavey:
updated with requested changes, resolved conflicts, tested - requested changes produces a more meaningful error than the previous attempt

karlthepagan:
This needs to be rebased onto dev.

I tried and got lost in the changes tho. Maybe it would be easier if you reimplemented it on the current dev. The concept is simple enough.

karlthepagan:
Execute the following commands to resolve the mess:

```
git checkout dev
git checkout -b temp
git merge --squash change/trycatch
git commit
git checkout change/trycatch
git reset --hard temp
git push -f
```

jschavey:
added try/catch wrapping for flagdir and room as suggested - is room.analyze() wrapping redundant?

chobobobo:
Yes, the reason it is redundant is that you want the try/catch to activate for a specific room/flag, and halt execution of that room/flag.  With the current fix, it will halt execution of all rooms/all flags which is not desirable.  You need to move the scope of the try/catch in FlagDir inside each of the functions that get called on all flags so that errors with one flag won't halt execution of other flags.

flagDir.js
```diff
@@ -201,44 +201,51 @@ mod.flush = function(){
...
-        if(!Game.flags[flagName]) {
-            this.stale.push(flagName);
-        }
-    };
-    _.forEach(Memory.flags, findStaleFlags);
-    const specialFlag = mod.specialFlag(true);
-    return !!specialFlag;
+    try {
+        let register = flag => {
+            flag.creeps = {};
```
chobobobo: try would go inside register

flagDir.js
```diff
@@ -201,44 +201,51 @@ mod.flush = function(){
...
+                roomName: flag.pos.roomName,
+                x: flag.pos.x,
+                y: flag.pos.y,
+                cloaking: flag.cloaking
+            });
+        };
+        _.forEach(Game.flags, register);
+        
+        let findStaleFlags = (entry, flagName) => {
+            if(!Game.flags[flagName]) {
```
chobobobo: inside here too

flagDir.js
```diff
@@ -201,44 +201,51 @@ mod.flush = function(){
...
+        const specialFlag = mod.specialFlag(true);
+        return !!specialFlag;
+    } catch (e) {
+        Util.logError(e.stack || e.message);
+    }
 };
 mod.execute = function() {
+    try {
+        let triggerFound = entry => {
+            if( !entry.cloaking || entry.cloaking == 0) {
```
chobobobo: inside here

room.js
```diff
@@ -2760,92 +2760,100 @@ mod.totalStructuresChanged = function() {
...
-    _.forEach(Game.rooms, r => {
-        getEnvironment(r);
-        p.checkCPU(r.name, PROFILING.ANALYZE_LIMIT / 5);
-    });
+    try {
+        const p = Util.startProfiling('Room.analyze', {enabled:PROFILING.ROOMS});
+        const totalSitesChanged = Room.totalSitesChanged();
+        const totalStructuresChanged = Room.totalStructuresChanged();
+        const getEnvironment = room => {
+            try {
```
chobobobo: this causes errors to be caught on a per-room basis, rather than for all rooms

room.js
```diff
@@ -2760,92 +2760,100 @@ mod.totalStructuresChanged = function() {
...
+            if( NOTIFICATE_INVADER || (NOTIFICATE_INTRUDER && creep.owner.username !== 'Invader' && creep.room.my) || (NOTIFICATE_HOSTILES && creep.owner.username !== 'Invader') ){
+                Game.notify(`Hostile intruder ${creep.id} (${bodyCount}) from "${creep.owner.username}" in room ${creep.pos.roomName} at ${toDateTimeString(toLocalDate(new Date()))}`);
+            }
+            // trigger subscribers
+            Room.newInvader.trigger(creep);
+        }
+        let triggerKnownInvaders = id =>  Room.knownInvader.trigger(id);
+        let triggerGoneInvaders = id =>  Room.goneInvader.trigger(id);
+        let run = (memory, roomName) => {
+            const p2 = Util.startProfiling(roomName, {enabled:PROFILING.ROOMS});
```
chobobobo: try/catch goes inside run, which calls the other functions anyways

jschavey:
try/catch wrapper scope shifted to throw per flag/room

karlthepagan:
I'm interested in CPU feedback. @xephael maybe?

karlthepagan:
To give you context on that CPU concern, proximo had shared with me his analysis that a try/catch slows down everything it calls, but I'm skeptical.

jschavey:
As I understand it the concern is valid.  Here is an interesting test demonstration: https://jsperf.com/try-catch-performance-overhead

chobobobo:
It would appear it is optimized in V8 versions 5.3.1 and later https://github.com/v8/v8/commit/9aac80f
based on http://screeps.wikia.com/wiki/Internals Screeps is currently running 5.1.281.75 how do we determine the current node.js/V8 version live on Screeps I wonder?

jschavey:
Second link doesn't mention but first link states V8 5.1.281.75.  I suppose this means anything we wrap in try/catch won't be optimized.  I wonder if this is part of my creep.run() has always profiled as highest usage for me (it was already in a try/catch before this change see creep.js line 382)

chobobobo:
Yes, this information has made me rethink the value there, I'm going to profile the difference between enabled/disabled.

chobobobo:
Ok, did some profiling by moving the try/catch inside of Creep.prototype.run to only cover the code for running creeps (don't profile NON_CRITICAL/spawning) and the results are:
public server:
with try/catch
public - with try catch
without
public - without try catch

private server:
with
with try catch - private server
without
without try catch - private server

I noticed about 10% higher usage on the private server, but equal usage over a few tests on the public server.. this leads me to suspect that they have updated their version of node.js since they last told us

jschavey:
My profiling tests yielded the same results.  On a private server, with all try/catches in tact over 1000 ticks the average CPU usage roughly 10% more than with all try/catches removed (same tick period).  On the official server, the delta is 0.02 CPU over 1000 ticks.  I believe we can use try/catches with little penalty on official servers.

chobobobo:
@karlthepagan tests have come out showing that on production there's no appreciable difference, I'll leave final approval with you for this change though.

karlthepagan:
Excellent work. That's sufficient I think.

PR #1201, opened 2017-04-08
Draw Line Upon Action Assignment

### LINKED ISSUE #1189: Draw Line Upon Action Assignment
When a creep is assigned an action, I want to draw a line from the creep to it's target, just for the tick it was assigned.

Spedwards:
Resolves #1189.


creep.Action.js
```diff
@@ -108,8 +108,12 @@ let Action = function(actionName){
         return false;
     };
     // assignment postprocessing
-    // needs implementation in derived action
-    this.onAssignment = (creep, target) => {};
+    this.onAssignment = function(creep, target) {
+        if (SAY_ASSIGNMENT) creep.say(ACTION_SAY[this.name.toUpperCase()], SAY_PUBLIC);
+        if (target instanceof RoomObject || target instanceof RoomPosition && VISUALS.ACTION_ASSIGNMENT) {
+            Visuals.drawLine(creep, target);
```
karlthepagan: let me help with this: we should have a means to explain target using a callback so that special flag and out of room targets are drawable

Spedwards: How would you draw a line to something out of the room?

karlthepagan:
I want this in 0.7 - it's a nice change that also removes repetition appropriately.

PR #1210, opened 2017-04-10
modulesReload - module inconsistency

chobobobo:
The issue is that when you delete Memory.modules, it forces global.install() to get run, but it only gets run once, so the other server instances do not get updated with the new modules code.  This change sets the timestamp Memory.modules.valid and reloads on each server that Global.modulesValid does not match that timestamp.

Looking at @cyberblast's updated code, we could overhaul our main loop to better track server instances and reloads, but this is just enough to fix the current issue for 0.7.0


karlthepagan:
Looks like this will reload on all blades multiple times. Will take a closer look.

chobobobo:
The first iteration I wrote did, I am currently running this in production and have tested it to confirm it only reloads on each instance once.

PR #1211, opened 2017-04-10
hostileRampartSites - fix costMatrix for hostile rampart construction sites

chobobobo:
The issue is that hostile rampart construction sites were being considered as obstacles, this change makes it so they are not set to 0xFF as they were before.


karlthepagan:
The description needs more info to describe this change I think.

Also the comment says what is going on but not why

Spedwards:
@karlthepagan Hostile rampart construction sites are getting counted the same as friendly ones, so our creeps won't walk over them.

PR #1212, opened 2017-04-10
fix first load where modules.valid is undefined

chobobobo:
In existing builds, modules.valid was undefined causing the new code to constantly run global.install() this ensures that it gets set for a smooth transtition.


karlthepagan:
image

between the red lines is dev without this fix

after the 2nd red line... with the fix... yes pls

PR #1213, opened 2017-04-10
improveProfilingAgain - further improvements to profiling and simplification

chobobobo:
adds `wrap(function(), name, limit, type)` to avoid needing two lines for every profile
eg. `wrap(room.execute(), 'execute', PROFILING.ANALYZE_LIMIT)`
as well as `checkAndRun(object, method, limit, type)`
eg. `checkAndRun(room, 'saveContainers')`
both wrap the calls and pass on the return values

also some formatting fixes, set the default threshold when none is defined, fixed CREEP_TYPE displaying when CREEP was not set.

### DIFF
--- a/main.js
+++ b/main.js
@@ -251,7 +251,7 @@ module.exports.loop = function () {
     if (!cpuAtFirstLoop) cpuAtFirstLoop = cpuAtLoop;
     // ensure required memory namespaces
     if (Memory.modules === undefined)  {
-        global.install();
+        p.wrap(global.install(), 'install');
     }
     if (Memory.debugTrace === undefined) {
         Memory.debugTrace = {error:true, no:{}};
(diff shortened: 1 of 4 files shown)


karlthepagan:
Fewer lines of code, but not easier to understand what's happening.

main.js
```diff
@@ -282,42 +281,33 @@ module.exports.loop = function () {
     p.checkCPU('flush', PROFILING.FLUSH_LIMIT);
 
     // Room event hooks must be registered before analyze for costMatrixInvalid
-    Room.register();
+    p.wrap(Room.register(), 'Room.register', PROFILING.REGISTER_LIMIT);
```
chobobobo: invoke perhaps?

karlthepagan: Definitely not invoke - the function doesn't do the advertised verb in that case.

Decorate is probably ok.

PR #1220, opened 2017-04-10
classic OCS boosting action

karlthepagan:
works with current OCS code. I think this would work for 0.7


chobobobo:
I'll put this in canary as well

SirLovi:
Shouldn't this have the upgrader change as well? The boosting does not work without it..

karlthepagan:
@SirLovi upgrader change will be an internal / tutorial task.

I do intend to show players how to do this precisely by modifying upgrader in the tutorial. This also means an upgrader rewrite so that it's less ugly.

PR #1225, opened 2017-04-11
hostileFindRoute - findRoute was broken for hostile rooms

chobobobo:
Noticed yet another creep trying to traverse my neighbours room, and finally discovered the error.  I shouldn't have put this up with the Room.prototype methods in the first place.

It was using `this.hostile`, but `this` did not refer to the room that was being checked.  Checking `Memory.rooms[roomName].hostile` will return true for positive numbers, and overridden true, and false for undefined and overridden false values


room.js
```diff
@@ -2935,6 +2899,42 @@ mod.findSpawnRoom = function(params){
...
+        if( roomName !== destination && ROUTE_ROOM_COST[roomName]) {
+            return ROUTE_ROOM_COST[roomName];
+        }
+        let isHighway = false;
+        if( options.preferHighway ){
+            const parsed = /^[WE]([0-9]+)[NS]([0-9]+)$/.exec(roomName);
+            isHighway = (parsed[1] % 10 === 0) || (parsed[2] % 10 === 0);
+        }
+        let isMyOrNeutralRoom = false;
+        const hostile = _.get(Memory.rooms[roomName], 'hostile', false);
```
karlthepagan: will always return a safe value no matter the status of memory: `_.get(Memory, ['rooms', roomName, 'hostile], false);`

but I think this is going to be safe unless memory was deleted in the current tick

PR #1228, opened 2017-04-11
change/reserveMaxRange

karlthepagan:
(EDIT: pasted the wrong text here... I'm tired :P)

set max spawn range to something reasonable

40% effectiveness is bad but not universally unacceptable

users can override it in their mainInjection by setting Task.reserve.spawnRoomMaxRange = 4


Spedwards:
In response to your initial description, this could be viral'd in `viral.task.reserve` simply enough as well.

PR #1229, opened 2017-04-11
feature/explainSayAssignment

karlthepagan:
say assignment and draw line when creep.explain() is called

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -369,7 +369,7 @@ mod.extend = function(){
     // Explain API extension
     Creep.prototype.explainAgent = function() {
         if (this.action) {
-            this.action.onAssignment(this, this.target);
+            this.action.showAssignment(this, this.target);
         }
         return `ttl:${this.ticksToLive} pos:${this.pos}`;
     };
--- a/creep.js
+++ b/creep.js
@@ -368,6 +368,9 @@ mod.extend = function(){
 
     // Explain API extension
     Creep.prototype.explainAgent = function() {
+        if (this.action) {
+            this.action.onAssignment(this, this.target);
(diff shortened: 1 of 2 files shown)


karlthepagan:
this is simple enough unless there's an objection I'd like to consider it for 0.7.0

karlthepagan:
This is a bad implementation. I should be using the full explain API. It works tho, so please put it in canary and see how you like it.

creep.js
```diff
@@ -368,6 +368,9 @@ mod.extend = function(){
 
     // Explain API extension
     Creep.prototype.explainAgent = function() {
+        if (this.action) {
+            this.action.onAssignment(this, this.target);
```
Spedwards: I don't like calling `onAssignment` directly. Just in case a future action adds some sort of post-assignment verification to their `onAssignment`, we don't want to be calling it.

That isn't a problem now, but it could be a problem in future.

Perhaps move what you want from `onAssignment` into another method, then call that method from `onAssignment` and here.

karlthepagan: 100% agreed. Feel free to give me a method to refactor onto.

I'm also going to spend some time debugging the Explain API which isn't working exactly as I'd like.

Spedwards: `showAssignment` would be a good method name for Creep.Action. Just call it from `onAssignment` and you're good.

karlthepagan:
Thanks for the explain assignment fix. I still need to change where this gets called from. It can wait.

karlthepagan:
This is on hold because the callback from the Explain API is not working properly and I want to do it right the first time.

PR #1233, opened 2017-04-12
Update Room Layout

Spedwards:
New Layout: https://i.imgur.com/ZsJt4LN.png


chobobobo:
Does this replace the previous room layout, or add a second option?

Spedwards:
It replaces the previous layout. This one is an improvement upon the last.

PR #1234, opened 2017-04-12
Get Room Layout

Spedwards:
Includes fancy RegEx Magic.


Spedwards:
@chobobobo Viral the layout in room with the array output and then test by placing the White/Cyan flag.

Issue #1172, opened 2017-04-05
Profile 


chobobobo:

chobobobo:
best evidence so far I caught on a private server ```[11:11:49 PM]main:processSegments > 12.63  
[11:11:49 PM]main:flush > 11.45  
[11:11:49 PM]main:Room.analyze > 229.6  
[11:11:49 PM]main:Population.analyze > 16.98  
[11:11:49 PM]W5N8 > Off to work remoteMiner-W5N9-mine-5!  
[11:11:49 PM]main:FlagDir.execute > 48.72  
[11:11:49 PM]main:creep.execute > 77.95  
[11:11:49 PM]main >  loop:484.17 other:0.17 avg:53.39 ticks:10082 bucket:10000  
[11:11:49 PM]
[11:11:54 PM]Script execution has been terminated with a hard reset: CPU limit reached
[11:12:05 PM]Your script is temporary blocked due to a hard reset inflicted to the runtime process.
Please try to change your code in order to prevent causing hard timeout resets.
[11:12:07 PM]Your script is temporary blocked due to a hard reset inflicted to the runtime process.
Please try to change your code in order to prevent causing hard timeout resets.
[11:12:12 PM]Your script is temporary blocked due to a hard reset inflicted to the runtime process.
Please try to change your code in order to prevent causing hard timeout resets.
[11:12:17 PM]Your script is temporary blocked due to a hard reset inflicted to the runtime process.
Please try to change your code in order to prevent causing hard timeout resets.
[11:12:22 PM]Your script is temporary blocked due to a hard reset inflicted to the runtime process.
Please try to change your code in order to prevent causing hard timeout resets.```

chobobobo:
For context the 'bad ticks' are ticks that hit the CPU limit with no log messages.. I have traced them down to global.install and the usage looks something like :
```
[21:10:42]install
[21:10:42]mem 4.28
[21:10:42]inject global 9.98
[21:10:42]assign parameters 31.91
[21:10:42]load mainInjection 19.68
[21:10:42]install:assign global > 88.4  
[21:10:42]install:Util.assign > 19.95  
[21:10:42]install:Task.assign > 153.35  
[21:10:42]install:Creep.Action > 0.37  
[21:10:42]install:Creep.Setup > 17.87  
```
a normal tick looks like
```
[21:09:04]install
[21:09:04]mem 21
[21:09:04]inject global 3.17
[...]

PR #1237, opened 2017-04-12
fix/badTicks

chobobobo:
This is an attempt at helping to fix https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1172 it removes the assignment of the return values, as the side effect of running infect is that mod is injected with the updated values.

### DIFF
--- a/main.js
+++ b/main.js
@@ -73,7 +73,7 @@ global.inject = (base, alien, namespace) => {
             typeof alien[key] === 'object' && !Array.isArray(alien[key])) {
             global.inject(base[key], alien[key], namespace);
         } else {
-            base[key] = alien[key]
+            base[key] = alien[key];
         }
     }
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -106,8 +106,8 @@ global.load = (modName) => {
     }
     if( mod ) {
         // load viral overrides 
-        mod = infect(mod, 'internalViral', modName);
-        mod = infect(mod, 'viral', modName);
+        infect(mod, 'internalViral', modName);
+        infect(mod, 'viral', modName);
```
Spedwards: This basically equates to `a = a`. Not `a = {}` then `a = {}`. There's no lost value for GC to clean.

chobobobo:
Yeah, sadly it wasn't this.

PR #1241, opened 2017-04-13
preferTerminalWithdraw

chobobobo:
Set the charge for terminals to increase at a faster rate than storage so they are more preferable as withdraw targets.


chobobobo:
I've thought more about this, and I think this should remain unchanged, I'm experimenting with modifying withdraw to target the terminal whenever its charge is > 1 instead.

PR #1240, opened 2017-04-13
Visuals Refactor

Spedwards:
Basically reverted the last refactor and made everything instances.

PR #1246, opened 2017-04-13
Visual/arrows

Spedwards:
I hope I got my maths right. Haven't done Trig since 9th grade.

Depends on #1240 in every sense of the word.


Spedwards:
When the line is vertical (possibly horizontal as well), the arrow handles are not displayed. It seems as though the closer the line is to being vertical, the closer the arrow handles are to the line.

That being said, small arrows are cute.

chobobobo:
This seems to be working properly and should merge once visualsRefactor is merged.

Issue #1249, opened 2017-04-13
Terminal charge should consider setStore values


chobobobo:
Currently, the terminal charge depends only on TERMINAL_ENERGY, it should consider all orders and needs when determining its ideal charge.

This should likely be done as a part of the reallocating refactor.

PR #1250, opened 2017-04-13
withdrawingTargets - use terminal as valid withdraw when charge > 1 and don't debounce terminal

chobobobo:
This change should not be merged until #1249 is resolved as it will drain all terminals down to TERMINAL_STORAGE and put the energy in storage and start a competition between withdraw and reallocating.  However, it does set what I consider to be the correct priority for withdrawing from terminals.

### DIFF
--- a/creep.action.withdrawing.js
+++ b/creep.action.withdrawing.js
@@ -27,9 +27,9 @@ action.work = function(creep){
     return creep.withdraw(creep.target, RESOURCE_ENERGY);
 };
 action.assignDebounce = function(creep, outflowActions) {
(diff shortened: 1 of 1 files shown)


karlthepagan:
0.7.1 if this operates without a denounce refactor

chobobobo:
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/pull/1251 addresses #1249 and is simple enough for 0.7.1 I believe.  The only change to debounce here is to first check what target is returned, and only debounce if the target is storage.

chobobobo:
0.7.1 is here, and this works

PR #1255, opened 2017-04-14
Feature/skip room

Spedwards:


chobobobo:
What is the additional cost of running FlagDir.find on every creep and every room?  Another possible way to implement this would be to have the flag set room.memory.skip, and remove it when the flag is removed, and then add a skip property to Room.

room.js
```diff
@@ -2799,6 +2799,7 @@ mod.analyze = function() {
         }
     };
     _.forEach(Game.rooms, r => {
+        if (FlagDir.find(FLAG_COLOR.command.skipRoom, r)) return false;
```
Spedwards: This should be `return;`.

Returning `false` will exit the loop entirely whereas returning anything else will act as `continue;`.

chobobobo:
@Spedwards I have concerns about calling FlagDir.find on every creep and every room as you mentioned to me previously that it's performance is O(n) based on the number of flags.  Have you considered using memory instead set from the flag?

Issue #1256, opened 2017-04-14
Improve Remote Mining V2


chobobobo:
Current TODO for 0.7.0:
- if any CS exists outside range(3) of the miner build and send a remote worker
- delay sending haulers until the container has been completed
- haulers decide if they are going to live long enough to make another round trip, and turn themselves into local haulers if not
- proper cleanup when a flag is removed (remove queued creeps, recycle current creeps and spawning creeps, delete room memory)

Future Goals:
- don't send the miner until the worker has completed the container
- don't send haulers until the roads and container have been completed (setup first)
- investigate whether roads are necessary, what benefit they offer and support optional roads if they are a viable option
- split remote mining source destinations to support multiple destinations from a single remote room
- convert all settings to refer to carry parts rather than weight
- define in the wiki how we determine the ideal number of carry parts for a single source room at a given distance from the destination room so they have an idea how to calibrate using Task.mining.carry()
- define in the wiki the cost/distance ratio to help people determine whether that single source room 2 rooms away from spawn is actually worth it

chobobobo:
https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/972 is tracking cleanup

karlthepagan:
default parameters for 0.7.0 IMO:

```
    PIONEER_UNOWNED: true,
    REMOTE_HAULER: {
        ALLOW_OVER_CAPACITY: 2450,
        DRIVE_BY_BUILD_ALL: true,
        DRIVE_BY_BUILD_RANGE: 2,
        DRIVE_BY_BUILDING: true,
        DRIVE_BY_REPAIR_RANGE: 1,
        MIN_LOAD: 0.75,
        MIN_WEIGHT: 1700,
        MULTIPLIER: 12,
        REHOME: true,
    },
    REMOTE_RESERVE_HAUL_CAPACITY: 0.25,
    REMOTE_WORKER_MULTIPLIER: 1,
```

Spedwards:
Idea:
Instead of the miner waiting for the worker to build the container, have the miner build the container. Efficiency may be higher in rooms with 2 sources where one container is up.

karlthepagan:
I think that's what @chobobobo is suggesting in his item that I moved into 0.7.1

chobobobo:
[...]

PR #1258, opened 2017-04-14
taskCleanup - remove queued creeps from task.mining on flag removal

chobobobo:
cleanup can later be converted to be used for all tasks, currently this is only set for task.mining as a part of https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1256


chobobobo:
tested and working

chobobobo:
I don't know what I was thinking, task.cleanup will only work for mining, it's not generalized properly at all currently.

task.js
```diff
@@ -60,6 +60,32 @@ mod.memory = (task, s) => { // task:  (string) name of the task, s: (string) any
     if( !Memory.tasks[task][s] ) Memory.tasks[task][s] = {};
     return Memory.tasks[task][s];
 };
+mod.cleanup = function(subKeys, task, s) {
+    mod.removeQueued(mod.memory(task, s), subKeys);
+    mod.clearMemory(task, s);
+};
+mod.removeQueued = function(memory, subKeys) {
+    const removeEntries = function(mem) {
```
Spedwards: Should probably use an arrow function here instead. Allows for use of `this`. Not used here but it might be in future.

chobobobo: with my understanding, using an arrow function here would result in `this` being undefined, but using a function declaration `this` works properly because of the way we do binding.  When this is converted to be a class, we will have proper use of `this`

Spedwards: As is, any use of `this` will be the instance of `removeEntries`, not `mod`.

chobobobo: oh, you mean removeEntries should be an arrow function, I get it now!

Spedwards: Heh, I would have commented on `removeQueued` otherwise.

PR #1259, opened 2017-04-14
remoteRecycling - recycle remote mining related creeps on flag removal

chobobobo:
This code needs to be tested and run before being merged, there's an off-chance it will cause all remote mining related creeps to recycle unexpectedly which would be disastrous.

This is to resolve the 0.7.0 requirements for https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1256


chobobobo:
Ok, tested and working now.

creep.behaviour.remoteHauler.js
```diff
@@ -2,12 +2,18 @@ let mod = {};
 module.exports = mod;
 mod.name = 'remoteHauler';
 mod.run = function(creep) {
+    const flag = creep.data.destiny ? Game.flags[creep.data.destiny.targetName] : null;
```
Spedwards:
```
const flag = creep.data.destiny && Game.flags[creep.data.destiny.targetName];`
```

Same applies for the other behaviours.

PR #1261, opened 2017-04-14
Feature/avoid sk

chobobobo:
A potential starting point for getting creeps to avoid SK creeps whenever possible.


traveler.js
```diff
@@ -365,7 +365,7 @@ module.exports = function(globalOpts = {}){
             if (_.isUndefined(options.useFindRoute)) options.useFindRoute = global.ROUTE_PRECALCULATION;
             if (_.isUndefined(options.routeCallback)) options.routeCallback = Room.routeCallback(this.pos.roomName, destination.roomName, options);
             if (_.isUndefined(options.getCreepMatrix)) options.getCreepMatrix = room => room.creepMatrix;
-            if (_.isUndefined(options.getStructureMatrix)) options.getStructureMatrix = room => room.structureMatrix;
+            if (_.isUndefined(options.getStructureMatrix)) options.getStructureMatrix = room => Room.isSKRoom(room.name) ? room.avoidSKMatrix : room.structureMatrix;
```
chobobobo: Do we just build SK avoidance into the room based on options.avoidSK?  Some creeps, like SKK should be getting within range so we certainly don't want them using this matrix.

room.js
```diff
@@ -935,6 +935,22 @@ mod.extend = function(){
                 return this._creepMatrix;
             }
         },
+        'avoidSKMatrix': {
+            configurable: true,
+            get: function () {
+                if (_.isUndefined(this._SKMatrix) ) {
+                    this._SKMatrix = {};
+                    const SKCreeps = this.hostiles.filter(c => c.owner.username === 'Source Killer');
+                    const lairs = this.find(FIND_STRUCTURES, {filter: s => s.structureType == STRUCTURE_KEEPER_LAIR});
```
Spedwards: `this.structures.all`?

room.js
```diff
@@ -2690,6 +2706,21 @@ mod.extend = function(){
...
         }
     };
+    // toAvoid - a list of creeps to avoid sorted by owner
+    Room.prototype.getAvoidMatrix = function(toAvoid) {
+        for (const owner in toAvoid) {
+            const creeps = toAvoid[owner];
+            const avoidMatrix = this.structureMatrix.clone();
+            for (const creep in creeps) {
+                for (let x = Math.min(0, creep.pos.x - 3); x < Math.max(49, creep.pos.x + 3); x++) {
+                    for (let y = Math.min(0, creep.pox.y - 3); y < Math.max(49, creep.pos.y + 3)) {
```
Spedwards: Missing a `y++`

chobobobo:
Currently if the SK creeps are moving, this will be invalid.. we could just cache the positions of the SK creeps and if they have changed, rebuild.

chobobobo:
@karlthepagan this is a basic implementation and may be good enough for 0.7.0 public, however I know you have plans for extending this further.   Will this implementation work with your ideas, would you like the call signature to Room.getAvoidMatrix to be modified to support your own work in progress?

karlthepagan:
I think this will work. Use your judgement and we can always refactor more later

chobobobo:
Ok, this needs a few days in canary for testing before merge as it could result in creeps getting lost/dying in SK rooms, we need to decide what creeps should be given the privilege and what should not.. also some (many) SK rooms will not have good traversal/exits so some small creeps should be told somehow to avoid those rooms

karlthepagan:
I've run this all day. On the plus side my rooms from the other side of the sector can supply my SK rooms pretty well.

On the minus side because of my using findRoute the haulers from my center room were going thru SK rooms (without roads) until I changed their route cost.

So with some tuning this is working perfectly.

PR #1267, opened 2017-04-15
Accept false when using Util.get

Spedwards:


util.js
```diff
@@ -40,7 +40,7 @@ module.exports = {
      */
     get(object, path, defaultValue, setDefault = true) {
         const r = _.get(object, path);
-        if (!r && !_.isUndefined(defaultValue) && setDefault) {
+        if (!_.isUndefined(r) && !_.isUndefined(defaultValue) && setDefault) {
```
chobobobo: in this context I think you want _.isUndefined(r) not !_.isUndefined(r)

PR #1271, opened 2017-04-15
Profiler was spammy

jschavey:
Enabling PROFILE: true in parameter.js outputs a profiling message every tick.  The line I removed was adding a second blank line to output.  This just cleans up the console log a bit and prevents messages from leaving the conosle buffer as soon.


chobobobo:
The extra line was put there so you could separate profiling data between ticks, and was by design.  If you can find a way to make it programmatic, perhaps a parameter, so that people have the option of disabling it then I would approve that change.

PR #1280, opened 2017-04-16
hotfix/remoteHaulersAvoid - creeps don't idle near hostile enemies

chobobobo:
~adds the avoid action to remote haulers.~

### DIFF
--- a/creep.js
+++ b/creep.js
@@ -199,10 +199,10 @@ mod.extend = function(){
     Creep.prototype.idleMove = function( ) {
         if( this.fatigue > 0 ) return;
         // check if on road/structure
-        const needToMove = _.find(this.room.structures.piles, {'pos': this.pos}) ||
-                   this.room.lookForAt(LOOK_STRUCTURES, this.pos) ||
-                   this.room.lookForAt(LOOK_CONSTRUCTION_SITES, this.pos, {filter: s => s.my}) ||
-                   this.pos.findInRange(this.room.hostiles, 3);
+        const needToMove = _.find(this.room.structures.piles, {'pos': this.pos})
+            || this.room.lookForAt(LOOK_STRUCTURES, this.pos)
+            || this.room.lookForAt(LOOK_CONSTRUCTION_SITES, this.pos, {filter: s => s.my})
+            || this.pos.findInRange(this.room.hostiles, 3);
         if (needToMove) {
             let path;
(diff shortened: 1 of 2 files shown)


creep.behaviour.remoteHauler.js
```diff
@@ -7,7 +7,9 @@ mod.run = function(creep) {
         //TODO: in the future look for a nearby room we can support
         return Creep.action.recycling.assign(creep);
     }
-
+    if (Creep.action.avoiding.run(creep)) {
+        return;
```
karlthepagan: I'd like to consider an additional check or possibly hold this for 7.1

Most often haulers can run thru dangerous situations and retrieve energy via picking.

chobobobo: This can definitely wait for 0.7.1, and there's probably a more nuanced approach to take, the use case that's an issue for me is SK rooms when they idle next to the SK creep.. maybe that's the approach to take.

chobobobo:
Perhaps a better approach to take would be to avoid hostiles during idleMove?

creep.js
```diff
@@ -199,11 +199,11 @@ mod.extend = function(){
...
         // check if on road/structure
-        let here = _.chain(this.room.structures.piles).filter('pos', this.pos)
-            .concat(this.room.lookForAt(LOOK_STRUCTURES, this.pos))
-            .concat(this.room.lookForAt(LOOK_CONSTRUCTION_SITES, this.pos, {filter: s => s.my}))
-            .value();
-        if( here && here.length > 0 ) {
+        const needToMove = _.find(this.room.structures.piles, {'pos': this.pos}) ||
+                   this.room.lookForAt(LOOK_STRUCTURES, this.pos) ||
+                   this.room.lookForAt(LOOK_CONSTRUCTION_SITES, this.pos, {filter: s => s.my}) ||
+                   this.pos.findInRange(this.room.hostiles, 3);
```
Spedwards: The extra indentation is unnecessary.

chobobobo:
I'm a bit concerned about the CPU cost of the idle action as it is used so often.

chobobobo:
After testing, this causes significantly higher CPU usage for all idling creeps, perhaps these changes should be integrated into a larger refactor of idling instead.

karlthepagan:
Possibly adding avoiding action to haulers with a strategy is one way to approach this.

I think that idle should be kept relatively simple. Adding the logic for "keep away from danger" will make idle less cohesive. Unfortunately our current architecture doesn't let us consider multiple actions simultaneously.

PR #1286, opened 2017-04-16
travelerFindRoute

chobobobo:
This sets useFindRoute as the default behaviour, but allows for pathfinding without findRoute in the event it fails (eg, blocked pathfinding within a room requires finding another way around).


chobobobo:
See https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1287 for more details on why we're doing this.  I am open to other fixes that don't involve forcing findRoute for linear route distances <= 2, but I would need an explanation as to why it's better.

PR #1292, opened 2017-04-16
jump off border when avoiding

### LINKED ISSUE #1289: avoiding creeps get border stuck
https://screeps.com/a/#!/history/E4N86?t=18681399

the miners are avoiding the SK nearby, when I remove the flag all creeps went to recycle, but the miners are caught on the border until I move them manually

karlthepagan:
This should resolve #1289 (avoiding creeps get border stuck)


karlthepagan:
Usually the first on-border move is in a different room. After leaving the border it will have a chance to select a new action.

It's a marginal improvement for now.

Issue #1130, opened 2017-04-01
[RC 0.7.0] Resolve 'incomplete path' messages


chobobobo:
We should either implement fixes that account for these incomplete paths being an inconsequential and normal part of execution, or investigate why they are happening and resolve them.  We can't have a release that's spewing warnings like this.

karlthepagan:
With the current settings you won't get range undefined in this message. Perhaps output a new message when range = 0 saying that this creep's path is probably blocked.

PR #1294, opened 2017-04-17
quietTraveler - make debug messages optional, and more informative

chobobobo:
options.debug is set to global.DEBUG by default, but if false it will turn off all log messages from traveler.  Also, when options.range === 0, the message now informs you the destination may be blocked.

This resolves https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1130


traveler.js
```diff
@@ -227,24 +227,34 @@ module.exports = function(globalOpts = {}){
...
+                            `origin: ${creep.pos}, dest: ${destPos}`); 
+                    }
                 }
                 if (ret.incomplete) {
                     const route = ret.route && ret.route.length;
-                    console.log(`TRAVELER: incomplete path for ${creep.name} from ${creep.pos} to ${destPos}, range ${options.range}. Route length ${route}.`);
+                    if (options.debug && options.range === 0) {
+                        console.log(`TRAVELER: incomplete path for ${creep.name} from ${creep.pos} to ${destPos}, destination may be blocked.`);
+                    } else {
+                        console.log(`TRAVELER: incomplete path for ${creep.name} from ${creep.pos} to ${destPos}, range ${options.range}. Route length ${route}.`);
```
karlthepagan: else if (options.debug)?

PR #1298, opened 2017-04-17
enable creep path visualization in SK rooms

jschavey:
I'm not sure if this was by design, but creep path visualization was not working in SK rooms as they donot have controllers.  I rely on these when manually placing roads, so I wanted them turned back on.


karlthepagan:
I apologize for not keeping up with PR's. Sped approved the implementation in #1312

PR #1305, opened 2017-04-18
change/highwayWalls - Is Target Accessible

Spedwards:
Checks for highway rooms. Checks to see if the target is accessible based on highway walls.


Spedwards:
@karlthepagan I would like to eventually add a pathfinding check to `isAccessibleTarget` for non-highway rooms.

chobobobo:
This has been tested and works, but will fail in the (rare)
 edge case that a room's walls are non-uniform
eg.
screen shot 2017-04-23 at 2 21 36 pm

Also, corner rooms appear to be non-uniform as well, so checks there may need to be improved.

room.js
```diff
@@ -2746,22 +2746,43 @@ mod.extend = function(){
...
+                const bottom = !!_.find(this.getPositionAt(25, 26).lookFor(LOOK_STRUCTURES, s => s instanceof StructureWall));
+                const right = !!_.find(this.getPositionAt(26, 25).lookFor(LOOK_STRUCTURES, s => s instanceof StructureWall));
                 
                 // both in same quadrant
                 if (Math.floor(object.x / 25) === Math.floor(target.x / 25) && Math.floor(object.y / 25) === Math.floor(target.y / 25)) return true;
                 
                 if (top) {
+                    if (bottom) {
+                        // cross section
+                        if (left || right) {
```
chobobobo: Think this was meant to be `&&`

Spedwards: No. If it's a vertical wall with either a left or right offshoot, if they're not in the same vertical half, the target won't be accessible. `&&` would mean all are true, which is handled above.

room.js
```diff
@@ -2724,6 +2724,104 @@ mod.extend = function(){
...
+                
+                if (top && left && bottom && right) {
+                    // https://i.imgur.com/8lmqtbi.png
+                    return getQuadrant(object) === getQuadrant(target);
+                }
+                
+                if (top) {
+                    if (bottom) {
+                        // cross section
+                        if (left || right) {
```
chobobobo: I see how it's changed now, this if is redundant, if top && bottom and they aren't in the same vertical half we want to return false regardless of other walls.

room.js
```diff
@@ -2724,6 +2724,104 @@ mod.extend = function(){
...
+                            if (getVerHalf(object) !== getVerHalf(target)) return false;
+                        }
+                        if (left) {
+                            return Util.areEqual(RIGHT, getVerHalf(object), getHorHalf(target));
+                        } else {
+                            return Util.areEqual(LEFT, getVerHalf(object), getHorHalf(target));
```
chobobobo: I think we want `Util.areEqual(RIGHT, getVerHalf(object), getVerHalf(target)) || getQuadrant(object) === getQuadrant(target)`
and the same with `LEFT`

room.js
```diff
@@ -2724,6 +2724,104 @@ mod.extend = function(){
...
+                        }
+                    }
+                    if (left && right) {
+                        // cross section
+                        if (getHorHalf(object) !== getHorHalf(target)) return false;
+                        return Util.areEqual(BOTTOM, getHorHalf(object), getHorHalf(target));
+                    }
+                    if (Util.areEqual(BOTTOM, getHorHalf(object), getHorHalf(target))) return true;
+                    if (left) {
+                        if (Util.areEqual(RIGHT, getVerHalf(object), getHorHalf(target))) return true;
```
chobobobo: second should be getVerHalf

room.js
```diff
@@ -2724,6 +2724,104 @@ mod.extend = function(){
...
+                        if (getQuadrant(object) === TOP_LEFT && (getVerHalf(target) === RIGHT || getHorHalf(target) === BOTTOM)) return false;
+                    } else {
+                        if (Util.areEqual(LEFT, getVerHalf(object), getHorHalf(target))) return true;
```
chobobobo: verHalf

room.js
```diff
@@ -2724,6 +2724,104 @@ mod.extend = function(){
...
+                        if (getQuadrant(object) === TOP_RIGHT && (getVerHalf(target) === LEFT || getHorHalf(target) === BOTTOM)) return false;
+                    }
+                } else {
+                    if (left && right) {
+                        // cross section
+                        if (getHorHalf(object) !== getHorHalf(target)) return false;
+                        return Util.areEqual(TOP, getHorHalf(object), getHorHalf(target));
```
chobobobo: They could also be in the same bottom quadrant and still be valid.

room.js
```diff
@@ -2724,6 +2724,104 @@ mod.extend = function(){
...
+                    }
+                    if (Util.areEqual(TOP, getHorHalf(object), getHorHalf(target))) return true;
+                    if (left) {
+                        if (Util.areEqual(RIGHT, getVerHalf(object), getHorHalf(target))) return true;
+                        if (getQuadrant(object) === BOTTOM_LEFT && (getVerHalf(target) === RIGHT || getHorHalf(target) === TOP)) return false;
+                    } else {
+                        if (Util.areEqual(LEFT, getVerHalf(object), getHorHalf(target))) return true;
+                        if (getQuadrant(object) === BOTTOM_RIGHT && (getVerHalf(target) === LEFT || getHorHalf(target) === TOP)) return false;
+                    }
```
chobobobo: The last bit can be simplified as above by checking quadrants, and the second get should be `getVerHalf`

PR #1308, opened 2017-04-18
refactor/roomObservers

chobobobo:
Another tested structural refactor, this time into room.observers.

This also includes a slight improvement to room.labs because I moved the related analyze functionality out of room.js


Spedwards:
I wouldn't be opposed to change `room.structures.observer` to `room.observer` if it meant moving it to the new file.

chobobobo:
Yes, I think I'll leave room.structures until the other major parts have been moved over, but I agree, maybe any structure that's unique to a room should just have a room.name like terminals and storage do now (observer, nuker, powerSpawn)

PR #1312, opened 2017-04-19
show room visuals in non-controller room when using the room visibility browser plugin

karlthepagan:
I wanted creep paths in SK rooms too


jschavey:
Related? #1298

karlthepagan:
@jschavey oh man really sorry I didn't mean to steamroll that PR :(

I don't see either one as significantly better than the other, but I'll defer to @Spedwards because he's contributed so much to the visualization.

Issue #1314, opened 2017-04-20
placeReactionOrder does not scale up to "burst mode"


karlthepagan:

karlthepagan:
![image](https://cloud.githubusercontent.com/assets/308745/25244917/6ea81266-25b7-11e7-97ea-fb444071ad9a.png)

Canceling a reaction order will place the slave labs in an idle reaction state when they should still have 2 master labs. This means that they will be candidates for other reactions and mineral unloading.

karlthepagan:
Canceling one order unraveled the whole flower:
![image](https://cloud.githubusercontent.com/assets/308745/25246749/74ccc4b0-25bd-11e7-8fa8-e6a9fba84ad0.png)

karlthepagan:
overwriting lab orders might be a good way to force reactions into a burst mode

there just needs to be some checks when canceling an order to only change the lab's mode when the link is bi-directional

chobobobo:
Since this hasn't ever worked, I don't know that it urgently needs to be in 0.7.0, 0.7.1 would be fine with me.

karlthepagan:
There's a workaround for this in 0.7.0-rc2, but it may be worth tracking against our objective for any getNeeds / reactions work.

PR #1318, opened 2017-04-20
don't cancel child lab order if master id doesn't match

karlthepagan:
This resolves a bug I found with manually creating a burst mode: #1314


karlthepagan:
The comments in#1314 cover it. Basically things get messy because of the bad accounting. This prevents that mess from putting other reactions in limbo.

Issue #1315, opened 2017-04-20
healing action prevents border guard from defending both rooms


karlthepagan:
center room invaders are ruining my day

https://screeps.com/a/#!/history/E5N85?t=18740200

I'm going to look into a quick fix for 0.7.0 - this might be an internalViral

chobobobo:
Was this resolved by #1319 ?

karlthepagan:
Yes. I need to add the filter code into tutorial / internal.

karlthepagan:
Will be resolved with internal changes. I highly recommend others writing your own viral (as I do) if you want to get finer control like aggressively avoiding moving away from the guard position in different scenarios.

karlthepagan:
Internal was merged.

PR #1319, opened 2017-04-20
healing strategy

karlthepagan:
adds strategy capability to healing so that users / internal can viral the target selection


karlthepagan:
This resolves #1315 - either this needs to go into the release or healing action should be removed until 0.7.1

karlthepagan:
Yup. That's my standard for 0.7 right now: try not to make any new features.

PR #1322, opened 2017-04-20
Energy Available Fix

Spedwards:


visuals.js
```diff
@@ -400,7 +400,7 @@ const Visuals = class {
             }
     
             // Display Energy Available
-            if (!room.controller.reservation && !room.controller.owner) {
+            if (!room.controller.reservation && room.controller.owner) {
```
chobobobo: Should this check to see whether you are the owner, or is it meant to run on any room with an owner?

Spedwards: Any room.

Issue #1309, opened 2017-04-18
CostMatrix: timing and invalidation fixes


karlthepagan:
When the CostMatrix goes invalid from not having vision to a room mark it as dirty (but don't delete it).

When we path into that room queue it for scouting / observer.

When we gain vision of a dirty CostMatrix room we should update it.

When all rooms in a creep's route have had vision in < ticks than it has been on the route, make them re-evaluate their route.

When a creep on a route moves into a room which we did not have vision of the CostMatrix should be immediately updated and a local route segment should be recalculated?

karlthepagan:
For 0.7.0 we really need to stop invalidating costmatrixes when we don't have vision.

chobobobo:
I think the issue may be that we don't return valid cost matrices when we don't have vision, currently cost matrices are not rebuilt until there is vision.
```
mod.getCostMatrix = function(roomName) {
    var room = Game.rooms[roomName];
    if(!room) return;
    return room.costMatrix;
};
```

karlthepagan:
That seems like an easy fix.

chobobobo:
I'll put together a PR and then you can decide, it's not as simple as just removing the if check.

karlthepagan:
Ok let's keep it as simple and direct as possible. I will be very happy to have this in 0.7.0

karlthepagan:
I think that this is sufficiently worked around for 0.7.0-rc2

The step where we invalidate paths on gaining room visibility will have to wait.

PR #1323, opened 2017-04-23
fix/cachedMatrices

chobobobo:
Use our cached matrices for rooms without visibility.

Still need to invalidate current paths when a creep enters a room and triggers a costMatrix update.


karlthepagan:
This resolves #1309 I'll open a new issue for invalidating paths.

chobobobo:
Looks good with these changes, there is the additional issue that we do not cache SK room matrices at all currently, so this doesn't help with that case yet.

PR #1326, opened 2017-04-23
Check isActive

### LINKED ISSUE #1257: Check isActive on structures
We have a lot of checks to see if a structure exists, but we need to also add a check to see if the structure is active. Any rooms that downgrade, or are claimed after unclaiming will fail.

Spedwards:
Resolves #1257


creep.action.fortifying.js
```diff
@@ -4,10 +4,10 @@ action.maxPerTarget = 1;
 action.maxPerAction = 1;
 action.targetRange = 3;
 action.isValidAction = function(creep){
-    return (creep.carry.energy > 0 && ( !creep.room.storage || creep.room.storage.charge > 0.6 ));
+    return (creep.carry.energy > 0 && ( (!creep.room.storage || !creep.room.storage.isActive()) || creep.room.storage.charge > 0.6 ));
 };
 action.isValidTarget = function(target){
-    return ( target != null && target.hits && target.hits < target.hitsMax);
+    return ( target && target.isActive() && target.hits && target.hits < target.hitsMax);
 };
```
chobobobo: I don't think we need this check for fortifying because it involves a withdraw, not a transfer right?

Spedwards: Better safe than sorry.

Spedwards: @chobobobo You can still withdraw from invalid structures. You can't transfer to them.

You also cannot fortify inactive structures.

PR #1331, opened 2017-04-23
SCU label is now an integer representing number of spawning + queued creeps

jschavey:
This resolves an issue that did not respect Util.formatNumber's limitations and presents more useful and less redundant information to the user.

### DIFF
--- a/visuals.js
+++ b/visuals.js
@@ -341,7 +341,7 @@ const Visuals = class {
             let count = _(Game.spawns).filter('spawning').size();
             count += _(Game.rooms).map(r => r.spawnQueueHigh.concat(r.spawnQueueMedium, r.spawnQueueLow)).flatten().size();
             const SCU_PERCENTAGE = count / spawnCount;
-            this.drawPie(vis, SCU_PERCENTAGE, 1, 'SCU', getColourByPercentage(SCU_PERCENTAGE), {x, y: y++});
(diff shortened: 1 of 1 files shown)


karlthepagan:
It would make sense if this related to the number of body parts.

visuals.js
```diff
@@ -341,7 +341,7 @@ const Visuals = class {
             let count = _(Game.spawns).filter('spawning').size();
             count += _(Game.rooms).map(r => r.spawnQueueHigh.concat(r.spawnQueueMedium, r.spawnQueueLow)).flatten().size();
             const SCU_PERCENTAGE = count / spawnCount;
-            this.drawPie(vis, SCU_PERCENTAGE, 1, 'SCU', getColourByPercentage(SCU_PERCENTAGE), {x, y: y++});
+            this.drawPie(vis, Math.round(count), spawnCount, 'SCU', getColourByPercentage(SCU_PERCENTAGE), {x, y: y++});
```
Spedwards: This fits better with the rest of the pie displays in my opinion. `count` is a whole number though, so `Math.round` is not necessary.

Spedwards: `getColourByPercentage(SCU_PERCENTAGE)` should be `getColourByPercentage(Math.min(1, SCU_PERCENTAGE))`

chobobobo:
I second the comment that it should be based on parts rather than creep count, as that's a more reliable metric for spawn time.

PR #1332, opened 2017-04-24
hotfix/feeding - targets where energy === 0 were considered invalid

chobobobo:
Fixes a major issue with feeding to get it working again.


chobobobo:
Actually, this should be an undefined check I believe, reverting to a null check is not satisfactory.

PR #1336, opened 2017-04-25
Structure.isActive Caching

### LINKED ISSUE #1333: structure.isActive is actually an expensive call
While doing some profiling on action.feeding, I discovered that it nearly doubled the average usage of haulers and labTechs because of the cost of checking isActive on all spawns and extensions.
http://support.screeps.com/hc/en-us/community/posts/206611209-Structure-isActive-Absurdly-CPU-intensive references this issue, and @artem suggests caching.

My first thought is that we could check isActive once, and only update it when RCL changes.

Spedwards:
Resolves #1333


room.js
```diff
@@ -9,6 +9,9 @@ mod.register = function() {
         if (Room._ext[key].register) Room._ext[key].register();
     }
     Room.costMatrixInvalid.on(room => Room.rebuildCostMatrix(room.name || room));
+    Room.RCLChange.on(room => room.structures.all.filter(s => !(s instanceof StructureRoad)).forEach(s => {
```
chobobobo: Also no need to set active=false for Walls & Ramparts as they will not be checked, and could take up a significant amount of memory in an inactive room.

PR #1295, opened 2017-04-17
0.7.0

karlthepagan:
**10 contributors! Only slightly down from last month's release!**

* Integration of the [BonzAI Traveler API](https://github.com/bonzaiferroni/bonzAI/wiki/Traveler-API)
* travelling action supports many Traveler and border travel options
* Resource management improvements
* programmatic room layout & CYAN flags used for preplanned construction site placement.
* building action will work on allied structures
* bulldozing action will drive over enemy construction sites
* most actions target closest by range
* guards will heal injured creeps in the room
* major reallocating changes
* major robbing changes
* major withdrawing & hauler change: "debounce" prevents most withdrawing / storing infinite cycles
* workers spawned to prevent room collapse have their own behaviour
* labTech behaviour to handle reallocating and reaction fulfillment
* major miner & upgrader changes
* pathfinding avoids most fixed-location workers
* delay spawning of non-critical creeps when CPU bucket is low
* major task validation changes
* fortifying is allowed with less energy in storage

API
* Boosting action (tutorial coming soon)
* objects merge with the parent object when they are "viral'd"
* Action.assign now always validates 2nd argument using isAddableTarget
* OcsMemory: memory segment queue API (using segments 99 and counting down to 90 currently)
* Task: installTask, forceSpawn
* Util class
* FlagDir filter and find changes.
* Creep.assignAction
* Creep.explain() describes what a creep is doing
* Flag.compare, flag.compareTo
* storage.charge
* terminal.charge
* creep setup specifications now accept part count maps
* RoomPosition: radius(), newFlag
* Room.newFlag
* Structure.getNeeds
* room: newFlag(), structures.extensions, powerSpawns.all, nukers.all, structures.my, myConstructionSites, constructionSites now includes hostile, immobileCreeps, structureMatrix, creepMatrix, avoidSKMatrix, showCostMatrix()
* events: Room.costMatrixInvalid

Parameters
* ALL parameters (and other objects) now merge with the parent object when they are "viral'd"
* ACTION_SAY - customize action emojis
* DEBUG_STACKS: true - add stacktrace to every logging statement
* PIONEER_UNOWNED: true - lets pioneers build before claim happens
* PROFILE: true - profile different major OCS execution phases
* REMOVE_CONSTRUCTION_FLAG: false - don't remove CYAN flags after site placed
* REMOTE_HAULER - many remote hauler config options
* ROBBER_REHOME: true - robbers will deliver their haul to a specified room
[...]

PR #1337, opened 2017-04-25
0.7.0 - Hotfix/reuseCachedMatrices

chobobobo:
This merges into the 0.7.0 RC to bring in a fix to properly invalidate rather than remove cached matrices so they can get rebuilt when we get vision again later.


chobobobo:
This is for #1295

karlthepagan:
Thanks for this fix. I'll try to test it tonight.

karlthepagan:
i've hit a few rebuilds without issue, will keep moving

Issue #1339, opened 2017-04-26
Extensions aren't being added to memory on their own


torin555:
*As discussed on Slack*

Hmm, interesting little bug. When I build new extensions it's not adding them to memory. I've been having to delete my room memory and have it recreated to get my new extensions to work. Any idea why that might be happening?

chobobobo:
It's entirely possible that we are not updating the in-memory cache when new extensions are constructed, need to investigate a good way to do this.

PR #1340, opened 2017-04-26
hotfix/cachedFeeding - make sure memory cache of extensions and spawns is up to date

chobobobo:
This resolves https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/1339

### DIFF
--- a/room.js
+++ b/room.js
@@ -606,14 +606,18 @@ mod.extend = function(){
 
     Room.prototype.countMyStructures = function() {
         const numStructures = _.size(this.structures.my);
-        // only trigger when a structure has been destroyed, we already avoid unpathable construction sites, and treat road sites like roads
         if (!_.isUndefined(this.memory.myTotalStructures) && numStructures < this.memory.myTotalStructures) {
+            // only trigger when a structure has been destroyed, we already avoid unpathable construction sites, and treat road sites like roads
             Room.costMatrixInvalid.trigger(this);
-            // these are vital for feeding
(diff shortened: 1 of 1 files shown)


room.js
```diff
@@ -606,14 +606,18 @@ mod.extend = function(){
...
         if (!_.isUndefined(this.memory.myTotalStructures) && numStructures < this.memory.myTotalStructures) {
+            // only trigger when a structure has been destroyed, we already avoid unpathable construction sites, and treat road sites like roads
             Room.costMatrixInvalid.trigger(this);
-            // these are vital for feeding
-            this.saveExtensions();
-            this.saveSpawns();
         }
-        if (numStructures > 0) this.memory.myTotalStructures = numStructures;
+        if (numStructures > 0) {
+            if (numStructures !== this.memory.myTotalStructures) {
```
Spedwards: Could you not just change that previous check from `num < total` to `num !== total`?

chobobobo: I'm not sure I fully understand, but if you're talking about `numStructures < this.memory.myTotalStructures` the reason it's `<` is mentioned in the comment, we don't need to rebuild the costMatrix when structures are completed as the costMatrix already makes construction sites unpathable when they are unpathable structures.. although treating road construction sites like roads is an optional parameter.. so perhaps it's best just to be safe.

chobobobo:
This needs to be merged ASAP as feeding will be broken for new bases on the current dev.

PR #1349, opened 2017-05-05
Virtual Flags: Add 4 Flag Colours

Spedwards:
This will allow for 4 more flag colours, which is a total of 40 more flag options. This change will allow us to move forward and add more flag options without having to worry about hitting the limit of 100 combinations in the original colours.

TODO:
 - Hide original flag and display visual showing new colours.
 - Allow for a tertiary colour (instant 900 combinations from the original 10 colours alone)

### DIFF
--- a/flagDir.js
+++ b/flagDir.js
@@ -195,6 +195,62 @@ mod.extend = function(){
             return pos.newFlag(flagColour, name);
         },
     });
+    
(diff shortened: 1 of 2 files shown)


chobobobo:
This hasn't broken anything, so I'm comfortable merging it into dev, but is there a good way to test that it functions as intended?

PR #1351, opened 2017-05-11
Fix ranger melee attack due to changes in the latest patch

chobobobo:
moving to a hostile used to be an implicit attack, it no longer is.


creep.action.defending.js
```diff
@@ -47,7 +47,10 @@ action.run = {
             }
             if( range < 3 ){
                 let direction = creep.target.pos.getDirectionTo(creep);
-                if( direction ) creep.move(direction);
+                if( direction ) {
+                    creep.move(direction);
+                    creep.attack(creep.target)
```
karlthepagan: I forgot to set `creep.attacking = true` with this recommendation

chobobobo: However, creep.attacking was not set previously, so my thinking is that it doesn't need to be backported?

chobobobo:
squashed it so I can cherrypick more easily.

creep.action.defending.js
```diff
@@ -47,7 +47,12 @@ action.run = {
             }
             if( range < 3 ){
                 let direction = creep.target.pos.getDirectionTo(creep);
-                if( direction ) creep.move(direction);
+                if( direction ) {
+                    creep.move(direction);
+                    if (range === 1) {
+                        creep.attacking = creep.attack(creep.target) == OK;
```
Spedwards: (\*screams\*) _THREE EQUAL SIGNS!_

PR #1354, opened 2017-05-14
hotfix/miningRoomUndef - Allow task.mining to operate when rooms are not visible.

chobobobo:
Changes task.mining to focus on the room name rather than the actual room object, since we only ever needed the name.


task.mining.js
```diff
@@ -148,8 +148,8 @@ mod.checkForRequiredCreeps = (flag) => {
             }
 
             // haulers set homeRoom if closer storage exists
-            const storageRoom = REMOTE_HAULER.REHOME && mod.strategies.hauler.homeRoom(roomName) || spawnRoom;
-            let maxWeight = mod.strategies.hauler.maxWeight(roomName, storageRoom, memory); // TODO Task.strategies
+            const storageRoomName = REMOTE_HAULER.REHOME && mod.strategies.hauler.homeRoomName(roomName) || spawnRoom.name;
```
Spedwards: If you want to save 2 characters, turn this into a ternary if-statement.

Issue #789, opened 2017-02-20
Robbers try withdrawing from invalid structures


OverlordQ:
Playing on PTR, notice robbers trying to withdraw energy from a Nuker structure which is not a valid target.

For reference, for valid Energy, Power, Mineral targets: https://github.com/screeps/engine/blob/a983452b8c8e3262f50260171fa7bf33b848682d/src/game/creeps.js#L578

Dasku:
Hmmm as a side note, we need to add nuker to that list to make it a valid target once the nuker/powerspawn/observer room objects are added

OverlordQ:
Maybe I'm misunderstanding, but nukers are deposit only structures

Dasku:
Didn't know that. Thought we could rob an enemy nuker

PR #1357, opened 2017-05-14
robbing will not attempt to work on nuker or powerspawn

karlthepagan:
also I'm terrible and didn't document robbing before I forgot it

also this is in 0.7.0

### DIFF
--- a/creep.action.robbing.js
+++ b/creep.action.robbing.js
@@ -10,6 +10,10 @@ action.isValidTarget = function(target){
         return false;
     }
 
+    if (target.structureType === STRUCTURE_NUKER || target.structureType === STRUCTURE_POWER_SPAWN) {
+        return false;
+    }
+
     return target.store || target.energy || target.mineralAmount;
 };
 action.newTarget = function(creep){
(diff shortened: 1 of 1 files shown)


chobobobo:
This resolves https://github.com/ScreepsOCS/screeps.behaviour-action-pattern/issues/789

karlthepagan:
I had a recent opportunity to rob a power spawn and my robbers sat next to it doing nothing.

It might be nice to generalize this solution by looking for ERR_INVALID_TARGET and banning targets that have that return.

chobobobo:
Well looking at the engine code, it appears powerSpawn is a valid withdraw target for both energy and power, so perhaps there was another reason it was invalid.  A check for ERR_INVALID_TARGET would address this.

karlthepagan:
Ok let's investigate for 7.1

chobobobo:
Ok, so withdraw is valid for a powerSpawn, but I just confirmed that there is a bug in the engine code preventing it from working currently, it always returns ERR_NOT_ENOUGH_RESOURCES
http://support.screeps.com/hc/en-us/community/posts/115002185869-bug-in-src-game-creeps-js-withdrawing-from-power-spawn?page=1#community_comment_115001366669

creep.action.robbing.js
```diff
@@ -10,6 +10,10 @@ action.isValidTarget = function(target){
         return false;
     }
 
+    if (target.structureType === STRUCTURE_NUKER || target.structureType === STRUCTURE_POWER_SPAWN) {
```
chobobobo: Add a comment here saying why PowerSpawns are disabled so we don't forget to re-enable it later, perhaps with a link to the bug issue, and I'm happy to merge this in

karlthepagan: +1

PR #1362, opened 2017-05-16
feature/idleFlags - improve idle performance and add idle flags

chobobobo:
Reduces the pathfinding costs from 1-2CPU/idle to 0.1-0.2 when white/brown idle flags are placed in the room.


chobobobo:
I think ultimately this is a stopgap, what we should do is have rooms determine 4-5 ideal idle spots and use those.

PR #1368, opened 2017-05-20
hotfix/collapseWorkerRecycle - don't recycle collapseWorkers right away

### LINKED ISSUE #1329: recoveryWorker despawns too soon
If your room is totally collapsed, the recoveryWorker despawns as soon as the worker begins spawning, it should wait until spawning is complete.

chobobobo:
Keep collapseWorkers around for 50 ticks after the room is no longer collapsed to give creeps time to spawn and take over.

This resolves #1329


karlthepagan:
This applies to 0.7.0 too

chobobobo:
Util.set is actually set only if it's not set, whereas _.set would set it every tick.

chobobobo:
functionality is not broken in 0.7.0 just not ideal, I would argue that the backport is not needed

PR #1371, opened 2017-05-20
hotfix/populationUndefined - avoid creeps without population entry

chobobobo:
This was fixed for tasks, but not for spawns that use setup.  Prevents creeps from taking names of creeps that have died in the same tick.


chobobobo:
This needs to be backported to 0.7.0

